Issue #126, opened 2014-11-18
add a thermostatic three way valve


damienpicard:

mwetter:
Closed until work on implementation continues.

PR #127, opened 2014-11-18
Issue126 thermostatic three way valve

damienpicard:
This pull request address issue #126


damienpicard:
@mwetter 
Thank you for the feedback. I updated the code accordingly except for the pressure:

You are right that if the resistances of the two legs differ from each other, the implementation is not correct. The model was originally made without pressure drop. Is this a too high limitation or would a hard coded implementation which obliges the resistances to be equal sufficient?

mwetter:
@damienpicard 

The documentation still says that no pressure drops are calculated (and it is not according to the style guide).

I think it is fine that the model only tracks the leaving set point if there is no additional resistance in the flow leg. But this need to be clearly explained in the documentation (otherwise, you have to answer it for various users individually, so better put it once in the documentation). If these resistances are in `dpFixed_nominal`, then you should consider them in your calculation of `k`. This will need to be added to the model (and a corresponding unit test added to verify that the implementation is correct).

`T_max` is still a parameter, but it is not required.

The current implementation also leads to the situation that two way valves and air dampers have the option to set their outlet temperature, which is not desired.

I don't understand what you wanted to verify in Annex60.Fluid.Actuators.Valves.Examples.ThermostaticThreeWayValve
Adding documentation and the .mos file would help here. Looking at the plot below, you never track the set point except during flow reversal.
thermostaticthreewayvalve

damienpicard:
@mwetter I made the required changes and merged the master branch. Let me know if some updates are still necessary.

damienpicard:
There is still a problem in the implementation: if the system uses pressure drops, the hydraulic sub-systems connected to each leg will have a different pressure drop. But the opening of the twoWayValve is calculated assuming the pressure drop is the same in each leg...

This could be solved by using a prescribed mass flow rate in one leg instead of twoWayValves. But a PID controller might then be more efficient as the balancing of the hydraulic circuits does not need to be exactly solved at each time step?

mwetter:
Damien will implement a model that exhibit the problem reported above so that Aachen can have a look at it.

mwetter:
Closed pull request until the implementation has been finished to a point where it can be moved.

Issue #131, opened 2014-11-21
Bug in mFactor implementation


Mathadon:
There is a bug in my implementation of the mFactor parameter in MixingVolumes. When having a variable moisture level the added 'mass' will react as if it also has a variable specific internal energy - which should in my opinion not be the case. Instead it should have a constant specific heat. I will propose a fix for this.

mwetter:
Merged to master.

Issue #65, opened 2014-05-08
small fixes needed to synchronize Annex60 with Buildings


mwetter:
This ticket will address minor fixes to allow synchronizing the `Annex60` with the `Buildings` library.

PR #134, opened 2014-11-21
fixed implementation of mFactor in ConservationEquation

Mathadon:
it now works properly with moist air

This is for #131


Mathadon:
I'll convert the input into a parameter, hopefully later this week.

We could indeed have changed the density. However this value for mFactor was used in an old unit test. Back then we did not use fluids so mFactor was used. Mathematically there is no reason why you cannot use mFactor < 1 but I agree it is not logical. Still, since it works and there may be some use for it I'd suggest to allow it.

mwetter:
@Mathadon : Can't you simply change the old unit tests. I think it is awkward to allow non-physical values just because one can mathematically. I see no use for mFactor < 1.
Also, recently I saw that E+ uses a similar construct also for the latent capacity of a room. Therefore, I suggest to rename `mFactor` to `mSenFactor`, which would allow later to add `mLatFactor`.

Mathadon:
Okay, I’ll implement the proposed changes

Mathadon:
I addressed your suggestions, ran unit tests and updated documentation. I think this is now ready.

Mathadon:
I had to merge the master branch again. Rerunning unit tests now.

Mathadon:
After merging the master branch and fixing mSenFactor naming in some new models, I am suddenly getting different unit test results:
```
*** Warning: Heater_T.mat: THeaOut.T has absolute and relative error = 2.381e-01, 7.978e-04.
             Maximum error is at t = 25920.0

*** Warning: Heater_T.mat: TVol.T has absolute and relative error = 2.384e-01, 8.138e-04.
             Maximum error is at t = 25920.0

*** Warning: Heater_T.mat: hea.TSet has absolute and relative error = 4.589e-01, 1.590e-03.
             Maximum error is at t = 141696.0

*** Warning: Heater_T.mat: hea.Q_flow has absolute and relative error = 1.005e+01, 1.010e-01.
             Maximum error is at t = 27648.0

             Acccept new file and update reference files? (Close plot window to continue.)
             Enter: y(yes), n(no), Y(yes for all), N(no for all): n
*** Warning: Heater_u.mat: THeaOut.T has absolute and relative error = 2.749e-01, 9.222e-04.
             Maximum error is at t = 25920.0

*** Warning: Heater_u.mat: TVol.T has absolute and relative error = 2.254e-01, 7.699e-04.
             Maximum error is at t = 25920.0

*** Warning: Heater_u.mat: hea.Q_flow has absolute and relative error = 3.620e+01, 1.000e+00.
             Maximum error is at t = 27648.0

             Acccept new file and update reference files? (Close plot window to continue.)
             Enter: y(yes), n(no), Y(yes for all), N(no for all): n
*** Warning: MixingVolumeMoistAir.mat: QLat.u has absolute and relative error = 8.114e-05, 8.062e-03.
             Maximum error is at t = 66.0

*** Warning: MixingVolumeMoistAir.mat: QSen.u has absolute and relative error = 6.944e-03, 3.628e-02.
             Maximum error is at t = 150.0

             Acccept new file and update reference files? (Close plot window to continue.)
             Enter: y(yes), n(no), Y(yes for all), N(no for all): n
```
Can these be numerical errors?

Mathadon:
Oh this may be because the mFactor implementation changed compared to the master branch and I have not yet updated the reference results for these new models (which I did already do for the others). I will therefore accept the new results

Mathadon:
Annex60_Fluid_MixingVolumes_Examples_MixingVolumeMoistAir.txt also gives deviating reference results. Although the difference is very small, I cannot explain it. This may however be related to #65 

@mwetter can you check if you also get different results?

mwetter:
Small differences in results can happen in C if different compilers are used.

Mathadon:
Would it then be wise to relax the tolerances in BuildingsPy to avoid confusion?

mwetter:
Just relaxing the tolerance seems to be a bit dangerous as it is already
now 1E-3. We may need to first analyze for what variables and values the
test fails, and then see whether we want to add a test for relative error
if it turns out that only variables with a large value fail.

Currently the test is:

                tolTim = 1E-3 # Tolerance for time

                if (abs(tR[0] - tS[0]) > tolTim) or abs(tR[-1] - tS[-1]) >
tolTim:

mwetter:
This is replaced by pull request https://github.com/iea-annex60/modelica-annex60/pull/151

Mathadon:
Using a relative tolerance does sound better/complementary. Is this something to discuss further in a BuildingsPy issue?

mwetter:
I just added an issue at https://github.com/lbl-srg/BuildingsPy/issues/49

PR #151, opened 2015-01-15
Issue131 m factor bug

mwetter:
This replaces pull request https://github.com/iea-annex60/modelica-annex60/pull/134 which contains the changes of that pull request and many additional revisions.


mwetter:
@Mathadon : Please review the commits I did today. There were many changes necessary. Several models (all movers, three way valves, expansion vessels and delay) still had the parameter `mSenFactor`. It does in my view not make sense to have it there (and I thought you removed it, but this seems not to be the case).

I also renamed `mSenFactor` to `mSenFac` to comply with the naming convention.

I also corrected the info section to use <code>xxx</code> to refer to variables, and <i>xx</i> to refer to expressions. Again this is according to the naming convention. Please review next time before a pull request is issued.

Let me know when you reviewed it so I can merge it to the `master`

Mathadon:
I reviewed and agree with the changes.

I propagated ```mSenFactor``` in models extending ```LumpedVolumeDeclarations``` instead of putting it to ```final 1```. I still think this can be useful in some cases but I won't argue further :)

I missed the difference between ```mSenFactor``` and ```mSenFac``` and forgot to update some of the documentation, my apologies.

PR #168, opened 2015-02-13
Fix typos

tbeu:
See https://github.com/vlajos/misspell_fixer


mwetter:
@tbeu Thanks for the pull request.

PR #172, opened 2015-02-20
also ignore Dymola success file

thorade:
And another file created by Dymola, saw this one for the first time today

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ request
 stat
 status
 stop
+success
 # Ignore files created by the unit tests.
 simulator.log
 unitTests.log


Mathadon:
Why don't you run your simulations in a directory other than the library directory?

thorade:
I haven't even thought about that yet. I just use the Load or Open button in Dymola and run the simulation in the current directoy (Load will cd the working directory to the directory of the package, I believe).
Would you recommend running the simulation in a different directory? 

PS: I know I had a couple of Pull Requests for the gitignore file, sorry for that, it was not on purpose. On the other hand they were all quite harmless & easy to review/merge.

Mathadon:
I use a script that loads all packages and then sets a default working directory where I do my simulations. I chose this directory to be /tmp/rendering (linux) such that it gets cleaned every time I reset my pc (I also have a startup script that creates the directory and copies all data files into it). The script (dymola.mos) contains:
```
openModel("pathToPackage/package.mo")
openModel("pathToPackage/IDEAS/package.mo")
openModel("pathToPackage/Annex60/package.mo")
openModel("pathToPackage/Buildings/package.mo")

OutputFlatModelica:=true;
Advanced.OutputModelicaCode:=true;
Hidden.PrintFailureToDifferentiate = true;
OutputCPUtime:=true;
Evaluate:=true;
Advanced.InvisibleConnections=true;
Advanced.EfficientMinorEvents = true;
Advanced.GenerateBlockTimers = true;
LogDifferentiatedEquations:=true;


cd("/tmp/rendering");
```

I think it's not a bad idea to split the library and working directory, partly because it avoids the problem you're dealing with :)

Issue #199, opened 2015-03-24
Types Azimuth and Tilt 


thorade:
Something went wrong and we now have 
[Annex60.HeatTransfer.Types.Azimuth](https://github.com/iea-annex60/modelica-annex60/tree/master/Annex60/HeatTransfer/Types)
as well as 
[Annex60.Types.Azimuth](https://github.com/iea-annex60/modelica-annex60/tree/master/Annex60/Types).

The first one should be removed.

thorade:
Until now, the second one is also not used -  is that true?

mwetter:
I will add `Types.Azimuth` to
`Annex60.BoundaryConditions.SolarGeometry.Examples.IncidenceAngle`

PR #201, opened 2015-03-24
Additional changes and bug fix for example that uses tilt and azimuth

mwetter:
These are additional modifications for #199.
It also contains a bug fix in 6fc0e4f


thorade:
Looks good to me.
You can go ahead and merge, I don't have the rights.

PR #203, opened 2015-03-25
Issue191 inline

### LINKED ISSUE #191: Remove inline=false for Media functions
I see no reason why they should not be inlined. For example, `Modelica.Media.SimpleAir.specificEnthalpy` has also no inline annotation.

mwetter:
@thorade : Can you please review and let me know if it is OK to merge. This will close #191


thorade:
Looks good to me.

PR #213, opened 2015-04-03
Issue212 ttws

### LINKED ISSUE #212: Apply ttws
Run
```
ttws --strip Annex60
ttws --eol=LF Annex60
ttws --clean Annex60
```
again to clean up code. See also https://github.com/iea-annex60/modelica-annex60/issues/71

mwetter:
This will close #212


damienpicard:
I'm fine with the changes. My only concern is that we will need to run 'ttws' every time we add a model or that we change some graphical interfaces. Dymola seems to add by itself this white spaces...

Issue #216, opened 2015-04-13
flow reversal simplification in conservationEquation


Mathadon:
When the flow never reverses (ie ```allowFlowReversal = false```) equations can be simplified. This is currently not being done for the enthalpy calculations in ```ConservationEquation``` since ```semiLinear()``` is not being used. See section 8.9.5 in Dymola User Manual Volume 2 for more information.

I'll propose a fix that changes the implementation, allowing such non-linear systems to be simplified.

Mathadon:
Using this change I managed to decouple a non-linear system of:
```
Sizes of nonlinear systems of equations: {7, 7, 7, 357}
Sizes after manipulation of the nonlinear systems: {2, 2, 2, 40}
```
into
```
Sizes of nonlinear systems of equations: {7, 7, 7, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18, 18}
Sizes after manipulation of the nonlinear systems: {2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2}
```

mwetter:
@Mathadon :

Did you check with the dynasim support why these simplifications are not also done if the `actualStream` operator is used? This would be more elegant to be handled by the symbolic processor.

Could you please also commit a small test case that illustrates the behavior? This way we can keep it with the regression tests.

Regardless, as the current Dymola version does not seem to handle this case, I suggest to add such a change to the `master`. However, a few changes should be done:
- the same should be applied to all other stream properties, not just to `h_outflow`.
- I suggest some construct like
```
  if allowFlowReversal then // or in view of the comment below "if allowFlowReveral and (massDynamics == ...SteadyState) then "
   .... // use actualStream, as this is the preferred implementation
 else
  ... // use semiLinear for this special case
end if
```
   This way, we can keep the stream concept where it makes sense, and fall back to using semiLinear for the special case.
- Your implementation seems incorrect if `massDynamics != Modelica.Fluid.Types.Dynamics.SteadyState`. If `medium.d` changes, then the mass in the volume changes and you can have flow reversal. Maybe the fluid volume should be such that the volume expands, rather than the mass? Or alternatively, `massDynamics` should by default be set to `energyDynamics` unless `allowFlowReveral=true`, in which case it should be steady state?

Mathadon:
Hi @mwetter 

[...]

PR #217, opened 2015-04-13
ConservationEquation.mo now uses semiLinear() for the calculation of energy conservation

Mathadon:
this allows equations to be simplified by Dymola when allowFlowReversal = false

See issue #216 for further motivation


Mathadon:
Unit tests are ok

Mathadon:
I need this for the #241 branch. Can this be merged soon or shall I merge it into branch ```issue241_speed``` myself?

Mathadon:
Can we have a look at this?

mwetter:
I will review the code today.

mwetter:
@Mathadon : 
Can you revise the numbers in
<h3>
AllowFlowReversal = true
</h3>
<p>
Sizes of nonlinear systems of equations: {7, 21, <b>56</b>}<br/>
Sizes after manipulation of the nonlinear systems: {2, 10, <b>12</b>}
</p>
<h3>
AllowFlowReversal = false
</h3>
<p>
<b>Old implementation</b>
</p>
<p>
Sizes of nonlinear systems of equations: {7, 21, <b>56</b>}<br/>
Sizes after manipulation of the nonlinear systems: {2, 10, <b>12</b>}
</p>
<p>
<b>New implementation</b>
</p>
<p>
Sizes of nonlinear systems of equations: {7, 21, <b>4</b>}<br/>
Sizes after manipulation of the nonlinear systems: {2, 10, <b>1</b>}
</p>

For example, with `allowFlowReversal=false`, I get
```
Sizes after manipulation of the nonlinear systems: {1, 9, 1}
```
with and without your changes to `ConservationEquation.mo`. Hence, for this example that should demonstrate the benefit of using `semiLinear`, I see no difference.


Also, please provide an `.mos` script with `allowFlowReversal=false`, otherwise we will not detect if we revert your changes.

Please first merge your branch with https://github.com/iea-annex60/modelica-annex60/tree/issue216_semiLinConsEq to avoid conflicts.

Mathadon:
Hey Michael,

the problem did indeed disappear after merging https://github.com/iea-annex60/modelica-annex60/tree/issue216_semiLinConsEq  into my branch. This is due to the changed implementation of ```MixingVolume``` that now uses ```Annex60.Fluid.Interfaces.StaticTwoPortConservationEquation``` so it makes sense that changes in ```Annex60.Fluid.Interfaces.ConservationEquation``` no longer have an influence.

I changed the example model by setting ```each vol.allowFlowReversal = true```. This causes ```Annex60.Fluid.Interfaces.ConservationEquation``` to be used again and the issue then reappears.

Note that there seems to be a bug in Dymola when you set ```Evaluate = true```, which I used to do by default. The result then change from
```
Sizes of nonlinear systems of equations: {6, 11, 4}
Sizes after manipulation of the nonlinear systems: {1, 9, 1}
```

into
```
Sizes of nonlinear systems of equations: {7, 21, 4}
Sizes after manipulation of the nonlinear systems: {2, 10, 1}
```
when ```allowFlowReversal=false```. This also causes the documentation to be 'wrong'. I'll fix this.

Mathadon:
Note that it seems a bit dull that I set ```vol.allowFlowReversal = true``` since the whole idea is that Dymola simplifies equations when ```allowFlowReversal=false```. However ```allowFlowReversal``` is still ```false``` in ```res```. Dymola detects that this is the case and therefore the simplification still happens. Note that this 'propagation' of ```m_flow(min=0)``` is not always properly executed and it may depend on how alias variables are chosen. I have submitted a bug report for this at Claytex.

PR #221, opened 2015-04-13
Removed LinePattern.None for #220

### LINKED ISSUE #220: Correct wrong use of LinePattern.None
Various examples in `Fluid.Actuators.Examples` use
```
  connect(y.y, valLin.y) annotation (Line(
      points={{-19,40},{10,40},{10,14}},
      color={0,0,127},
      pattern=LinePattern.None));
```
This is rendered as a solid visible line in Dymola 2015 FD01, but in 2016 beta 3, it is a line that is not visible.
Such annotations should be changed to either remove the `pattern` attribute, or set it to
`pattern=LinePatter.Solid`.
Also, all other classes in the library should be inspected for these wrong declaration.

mwetter:
This will close #220

### DIFF
--- a/Annex60/Fluid/Actuators/Dampers/Examples/Damper.mo
+++ b/Annex60/Fluid/Actuators/Dampers/Examples/Damper.mo
@@ -31,8 +31,7 @@ model Damper
 equation
   connect(yRam.y, res.y) annotation (Line(
       points={{1,50},{10,50},{10,32}},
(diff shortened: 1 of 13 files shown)


thorade:
Nothing changed but the LinePattern, in Dymola 2015 FD01 both versions look the same.
Seems to be OK, so go ahead and merge.

PR #229, opened 2015-05-01
Changed sign of dp_actual for #226

### LINKED ISSUE #226: Fluid.Movers.FlowControlled_dp has wrong sign for dp_actual
This component has connectors `dp_in` and `dp_actual`. It multiplies `dp_in` by a gain of `-1` before sending it to the filter and from there to `dp_actual`, or it sends it directly to `dp_actual`.

The gain should be moved to the input of `preSou` so that `dp_in` and `dp_actual` have the same sign. This also avoids a warning because both of these variables declare `min=0`. It also makes the model more consistent with the other movers which all have `*_in = *_actual` if the filter is removed.

mwetter:
This will close #226


Mathadon:
looks okay!

Issue #239, opened 2015-05-08
add translation statistics to reference results


mwetter:
This ticket will add translation statistics to the reference results.
The statistics can be generated with `BuildingsPy` from
```
git clone git@github.com:lbl-srg/BuildingsPy.git
cd BuildingsPy
git checkout -b issue53_stat
```
The statistics is added to catch model changes that can change the number and size of systems of linear and nonlinear equations and the number of state variables.

See also https://github.com/lbl-srg/BuildingsPy/pull/54

mwetter:
This change adds statistics to the reference results in the form
```
statistics-initialization={'numerical Jacobians': '0', 'nonlinear': '1, 1, 2,  14, , 14, 1', 'linear': '0, 0, 0, 0'}
statistics-simulation={'numerical Jacobians': '0', 'nonlinear': '1, 1, 1, 1, 8, 1, 14, 1, 0', 'number of continuous time states': '199', 'linear': '0, 0, 0'}
```

mwetter:
Closed issue as this is now on the master.

PR #240, opened 2015-05-08
Regenerated reference results for #239

mwetter:
These were generated with https://github.com/lbl-srg/BuildingsPy/commit/0fc5f3300817281465a18306644fbf8de1b1a696 branch issue53_stat

All results are the same, but the statistics for the initialization and simulation problem has been added. Also, trailing zeros are now no longer written to the reference results, e.g., rather than `1.000000000`, the output will be `1.`

I assign it to @Mathadon as he works mostly on efficiency of models, but @thorade , @damienpicard , @marcusfuchs, @mlauster, @nytschgeusen please also comment whether you agree with the revised format. Let's wait for a week before merging to see if there is any comment/objection.

This will flag any model change that leads to smaller or larger coupled systems of equations.

Reference results start now with lines such as
```
head -n 3 Resources/ReferenceResults/Dymola/Annex60_Fluid_Movers_Examples_PumpsSeries.txt
last-generated=2015-05-08
statistics-initialization={'numerical Jacobians': '0', 'nonlinear': '1'}
statistics-simulation={'numerical Jacobians': '0', 'nonlinear': '1, 3', 'number of continuous time states': '4', 'linear': ' '}

```
Any change to these statistics need to be confirmed by the user, similar to a change in the results.
The changes have also been tested with the `Buildings` library.


Mathadon:
@mwetter 
is my interpretation correct that you did not make a distinction between sizes before and after manipulation? I think it would be worth adding this. Linear systems often become reduced to size 0 and this could easily lead to a situation where you would not be able to spot that the linear system size before manipulation has increased.

mwetter:
@Mathadon 
Yes, for now I only recorded the systems after the manipulation as I am not sure if there is value to track the sizes before the manipulation as this code is not simulated. Do you have an example that would justify adding the sizes before manipulation?

Mathadon:
I don't have any right now and I think it'll only matter on relatively rare occasions so let's merge this!

PR #245, opened 2015-05-20
Removed stateSelect for p

### LINKED ISSUE #225: pressure state for incompressible media
When using incompressible media derivatives of pressures are generated when ```massDynamics not = steadyState``` because of the ```stateSelect.prefer``` declaration. This causes additional derivatives to be computed since there is no real state for the pressure when the medium is incompressible.

I will do some tests to check if it's best to remove the stateSelect.prefer or to set it to the mixing volume mass. We should also run unit tests and performance benchmarks to assess the impact.

mwetter:
This will close #225 and also required to export volumes as FMUs for https://github.com/lbl-srg/modelica-buildings/issues/313.

All unit tests run successful for the `Annex60` and the `Buildings` library.


Mathadon:
I checked the changes and they seem ok. I would however not close #225 since the ```stateSelect.prefer``` in ```Annex60.Fluid.Interfaces.ConservationEquation``` is not removed, which causes the 'problem' in ```Annex60.Fluid.MixingVolumes.Examples.MixingVolume``` not to be fixed.
Note that this option is already removed in the Buildings library. :
```

February 6, 2014 by Michael Wetter:
Removed 
p(stateSelect=if not (massDynamics == Modelica.Fluid.Types.Dynamics.SteadyState)
then StateSelect.prefer else StateSelect.default)
because the previous declaration led to the translation error 
The model requires derivatives of some inputs as listed below:
1 inlet.m_flow
1 inlet.p
when translating Buildings.Fluid.FMI.Examples.HeaterCoolerPrescribed. 
```
Can you merge this into Annex60 too?

mwetter:
@Mathadon : What do you mean by "this" in "Can you merge this into Annex60 too?"
`Annex60.Fluid.Interfaces.ConservationEquation` and `Annex60.Fluid.MixingVolumes.Examples.MixingVolume`
are the same on `Annex60`, branch `issue225_stateSelect` and on the master of the `Buildings` library.

I will merge what we have now on `issue225_stateSelect` and keep #225 open

Mathadon:
On ```Annex60```, branch ```issue225_stateSelect``` model ```Annex60.Fluid.Interfaces.ConservationEquation``` contains
```
p(start=p_start,
      stateSelect=if not (massDynamics == Modelica.Fluid.Types.Dynamics.SteadyState)
                     then StateSelect.prefer else StateSelect.default),
```
on ```Buildings``` branch ```issue313_fmi``` (I thought it was on ```master``` too, which is not the case) I find
```
p(start=p_start),
```
.

In the revision history of ```issue313_fmi``` I find
```

February 6, 2014 by Michael Wetter:
Removed 
p(stateSelect=if not (massDynamics == Modelica.Fluid.Types.Dynamics.SteadyState)
then StateSelect.prefer else StateSelect.default)
because the previous declaration led to the translation error 
The model requires derivatives of some inputs as listed below:
1 inlet.m_flow
1 inlet.p
when translating Buildings.Fluid.FMI.Examples.HeaterCoolerPrescribed. 
```

that gives a motivation for this difference. I would like to make the same change (removing stateselect) on the Annex60 master.

Issue #260, opened 2015-06-05
Removing Xi.stateSelect and preferredMediumStates from ConservationEquation


Mathadon:
Quite a few ```stateSelect``` modifiers have already been removed from ```ConservationEquation```. The current code is:
```
Medium.BaseProperties medium(
    preferredMediumStates= not (energyDynamics == Modelica.Fluid.Types.Dynamics.SteadyState),
    p(start=p_start),
    h(start=hStart),
    T(start=T_start),
    Xi(start=X_start[1:Medium.nXi],
       each stateSelect=if (not (substanceDynamics == Modelica.Fluid.Types.Dynamics.SteadyState))
                     then StateSelect.prefer else StateSelect.default),
    X(start=X_start),
    d(start=rho_nominal)) "Medium properties";
```
I propose to also set ```preferredMediumStates=false``` and to remove the ```stateSelect``` modifier for ```Xi```. I _believe_ this will result in better code in most of the cases. For instance ```Annex60.Fluid.MixingVolumes.Examples.MixingVolumeMoistAir``` contains 13 times ```der(``` in dsmodel.mof before and 7 times after the change. This is because ```mXi``` is used as a state instead of ```Xi```, which removes the need of the following equation:
```
der(vol1.dynBal.medium.Xi[1]) :=  -(der(vol1.dynBal.m)*vol1.dynBal.medium.Xi[1]
        -der(vol1.dynBal.mXi[1]))/vol1.dynBal.m;
```
Note that it makes more sense to use ```mXi``` as a state, since this allows better error control by the integrator on the to be conserved quantity: mass. 

Furthermore my building model will not translate with the current version of the code.

I will make a pull request.

mwetter:
@Mathadon :

I updated also the medium. I still get
```bash
grep -c "der(" dsmodel.mof 
7
```
Please review https://github.com/iea-annex60/modelica-annex60/pull/263 which contains your earlier pull request.

Mathadon:
I agree with the proposed changes.

mwetter:
This is now on the master

PR #261, opened 2015-06-05
Issue260 state select

Mathadon:
This is for #260 

I ran unit tests, which give minor differences in results and a relatively large change in result for ```vol1.p``` in ```Annex60.Fluid.MixingVolumes.Examples.MixingVolumeMoistAir```. This can be reduced by increasing the solver tolerance.

I did not yet accept the results since you'd probably want to change the differences and because I get small numerical differences in results compared to the existing results, even without any changes to the code.


mwetter:
Closed as this is integrated in https://github.com/iea-annex60/modelica-annex60/pull/263

Issue #272, opened 2015-06-11
Thermal losses in temperature sensor


Mathadon:
When there is no flow, the output temperature of a temperature sensor remains constant. It would be nice if we could add the option to simulate thermal losses, characterised by a thermal loss time constant and a steady state temperature. 
This is because I'm experiencing problems with sensors not exhibiting this behaviour. Is it desirable to add this functionality (disable by default)?

mwetter:
@Mathadon 

I like the proposed implementation (with a few minor modifications).

I revised the code on the new branch `issue272_lossTsensors` and did mainly the following changes:

I changed `enableThermalLosses` to `transferHeat` and `tauLoss` to `tauHeaTra` because for cooling systems, the sensor has a heat gain, not a loss, and hence heat transfer is a better term.

I added graphics that is enabled when `transferHeat=true` to indicate graphically the configuration of the sensor.

I updated the user guide.

I moved in the base class `m_flow_small` from the `Advanced` group to the `Advanced` tab. This is where this variable is for other models such as for heat exchangers. So for consistency, I did this change (which was not caused by your new sensor).

Please let me know if you agree to the changes or propose further modifications before I process the pull request.

Mathadon:
I like the changes. I propose however to remove the statement about Dymola since I do not think it is relevant and it is incorrect when ```Evaluate=false``` (I verified this in the meantime).

Mathadon:
Nice drawing skills by the way!

Mathadon:
This drew my attention to the fact that we cannot distinguish visually between the sensor with and without dynamics. In general I think it may be interesting to add a symbol when certain components can be configured between steady state and dynamic operation (similar to pump model). The only problem is that Dymola does not evaluate complicated parameter dependencies while editing so using them to determine whether or not the symbol should be shown, may lead to wrong indications while editing..

mwetter:
This is now on the master.

PR #277, opened 2015-06-19
Issue272 loss tsensors

Mathadon:
This is for issue #272 

I think the proposed approach is a good way of dealing thermal losses in a relatively simple way.

I updated the unit test example and results. I verified all unit tests in the fluid package.

Documentation was updated.


Mathadon:
I tink this is ready!

Issue #279, opened 2015-06-25
optimize Fluid.BaseClasses.FlowModels.basicFlowFunction_dp


mwetter:
The function `Fluid.BaseClasses.FlowModels.basicFlowFunction_dp` and `basicFlowFunction_m_flow` lead to many operations outside the turbulent region.
This issue should look at whether a simpler implementation leads to faster computation. One could for example use a linear function for |m_flow| < m_flow_turbulent/2 and use regularization between these two regions. This regularization could be done with a 3rd order polynomial. For fixed resistances, all polynomial coefficients are parameters, so there should be little overhead, whereas for valves, these coefficients need to be computed during the time stepping. Hence, maybe extending the linear function to 
|m_flow| < 2/3*m_flow_turbulent may be more efficient.

Mathadon:
I'm in favour of  this change since the combined functions are indeed very complex and often they require many evaluations because they are used in non-linear systems. Moreover a linear regime would better correspond to the laminar flow regime that occurs in practice. I suggest to look for a solution in this direction. For valves things may be a bit more complicated, although a pragmatic approximation may be sufficiently accurate?

Mathadon:
I created an (arguably dumb) implementation for this `Annex60.Fluid.BaseClasses.FlowModels.basicFlowFunction_dp`:

```
 m_flow :=smooth(1, if dp>0
                    then Annex60.Utilities.Math.Functions.spliceFunction(x=dp-dp_turbulent,pos=if dp>dp_turbulent/4 then sqrt(k*dp) else 0, neg=dp/dp_turbulent*sqrt(k*dp_turbulent)/2, deltax=dp_turbulent/3)
 else 
     Annex60.Utilities.Math.Functions.spliceFunction(x=-dp_turbulent-dp,pos=if dp<-dp_turbulent/4 then -sqrt(k*abs(dp)) else 0, neg=dp/dp_turbulent*sqrt(k*dp_turbulent)/2, deltax=dp_turbulent/3));
```

I then evaluated `Annex60.Fluid.Examples.Performance.Example2(from_dp.k=true)`, which does pretty much nothing else than evaluate that function. Results are

### before
```
Nonlin sys(5)  :    10,       20.357,       33.63 (      24.00 to      567.00),   605328
```
### after
```
Nonlin sys(5)  :    10,        8.987,       14.85 (       5.00 to      354.00),   605328
```

[...]

PR #295, opened 2015-07-16
Issue279 flow functions

Mathadon:
This is for #279


mwetter:
@marcusfuchs 
From my point of view, this is ready to be merged. If you agree with my latest commits, please go ahead and merge it to the master.

Issue #305, opened 2015-08-15
Simplify moisture fraction when allowFlowReversal=false


Mathadon:
`Annex60.Fluid.Interfaces.PartialTwoPortTransport` can be simplified when `allowFlowReversal=false`:

```
  port_a.Xi_outflow =  inStream(port_b.Xi_outflow);
  port_b.Xi_outflow = inStream(port_a.Xi_outflow);
  port_a.C_outflow = inStream(port_b.C_outflow);
  port_b.C_outflow = inStream(port_a.C_outflow);
```
then becomes
```
  port_a.Xi_outflow = if allowFlowReversal then inStream(port_b.Xi_outflow) else Medium.X_default[1:Medium.nXi];
  port_b.Xi_outflow = inStream(port_a.Xi_outflow);
  port_a.C_outflow = if allowFlowReversal then inStream(port_b.C_outflow) else zeros(Medium.nC);
  port_b.C_outflow = inStream(port_a.C_outflow);
```
this is similar to `Annex60.Fluid.Interfaces.StaticTwoPortConservationEquation`.

I'll make a  pull request.

mwetter:
This is now on the master.

PR #306, opened 2015-08-15
for #305 : implemented more efficient code for when allowFlowReversal…

Mathadon:
This is for #305


Mathadon:
All unit tests produced identical results.

PR #307, opened 2015-08-19
Issue303 derivatives

### LINKED ISSUE #303: Verify implementation of derivatives
- [x] The following examples need to be revised so that the derivative of the function argument is different from 1. This is needed to verify the correct implementation of the implementation of the derivative function:
  - [x] Media.Examples.AirDerivativeCheck
  - [x] Media.Examples.WaterDerivativeCheck
  - [x] Media.Specialized.Air.Examples.PerfectGasDerivativeCheck
  - [x] Media.Specialized.Water.Examples.TemperatureDependendentDerivativeCheck

- [x] The functions in the package `Utilities.Math.Functions.BaseClasses` do not multiply the return value with the derivative of the argument. This need to be corrected and the regression tests need to be updated to catch this error.

- [x] The package `Utilities.Psychrometrics.Functions.BaseClasses` also need to be verified.

- [x] The function
`Media.Specialized.Water.TemperatureDependentDensity.density_derT_p`
declares `smooth()` and `smoothOrder` which causes Dymola (and likely other tools too) to assume it is twice differentiable rather than once.

mwetter:
This will close #303.
The regression tests are updated and the changes have also been tested with the `Buildings` library.


marcusfuchs:
looks good!

PR #309, opened 2015-09-02
Issue266 pedantic

mwetter:
These changes are required for the models to translate in the pedantic Modelica mode.

The file `bin/runUnitTests.py` has also been changed to run the unit tests in the pedantic modelica mode. This requires the latest version of BuildingsPy from the master branch at https://github.com/lbl-srg/BuildingsPy
To use older versions of  BuildingsPy, the line
```
ut.pedanticModelica(True)
```
needs to be removed in `bin/runUnitTests.py`.
By default, BuildingsPy does not use pedantic modelica check as I don't expect that AixLib and IDEAS already work for that mode. Buildings does not yet work in pedantic check, but we will fix this.

### DIFF
--- a/Annex60/Experimental/Benchmarks/AirFlow/Components/SimpleZone.mo
+++ b/Annex60/Experimental/Benchmarks/AirFlow/Components/SimpleZone.mo
@@ -4,7 +4,7 @@ model SimpleZone "A room as a thermal zone represented by its air volume"
   replaceable package Medium = Modelica.Media.Interfaces.PartialMedium
     "Medium in the component";
 
-  parameter Modelica.SIunits.Temperature TRoom = 293.15
+  parameter Medium.Temperature TRoom(start=293.15) = 293.15
     "Indoor air temperature of room in K";
   parameter Modelica.SIunits.Height heightRoom = 3 "Height of room in m";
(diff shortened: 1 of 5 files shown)


Mathadon:
> By default, BuildingsPy does not use pedantic modelica check as I don't expect that AixLib and IDEAS already work for that mode.

You assumed correctly!

Mathadon:
Looks good. I do, however, get different reference results for these files:

Annex60_Airflow_Multizone_Examples_ChimneyShaftNoVolume.txt

Annex60_Airflow_Multizone_Examples_ClosedDoors.txt

Annex60_Fluid_HeatExchangers_Examples_Heater_T.txt

 Annex60_Fluid_HeatExchangers_Examples_Heater_u.txt
 
 Annex60_Fluid_Interfaces_Examples_Humidifier_u.txt
 
 Annex60_Fluid_Movers_Validation_Pump_Nrpm_stratos.txt
 
 
The differences are small, but it's cumbersome to look at the results all the time. Do we know what causes this yet?

mwetter:
@Mathadon : Regarding the small difference in results, this is because the numerics is different between different compilers and operating system. 
The ticket for this is https://github.com/lbl-srg/BuildingsPy/issues/49

Issue #281, opened 2015-06-30
Simplifying h_outflow calculation when allowFlowReversal = false


Mathadon:
Currently Dymola is able to simplify some calculations when ```allowFlowReversal = false```. Consider for instance the following equation.
```H_in = semiLinear(m_flow, inStream(port_a.h_outflow), inStream(port_b.h_outflow))``` 
```allowFlowReversal``` typically sets the ```min``` attribute of ```m_flow``` equal to zero. Dymola can then detect that mass flow will never become negative and that therefore ```inStream(port_a.h_outflow)``` should always be used in the equation: 
```H_in = m_flow*inStream(port_a.h_outflow)```
Since the equation now no longer depends on  ```inStream(port_b.h_outflow)``` this can lead to smaller algebraic loops.

However there are some cases where the ```min``` and ```max``` attributes are not propagated properly by Dymola and therefore these simplifications are not made. Moreover OpenModelica does not make these simplifications at all.  This can lead to large algebraic loops, which deteriorate simulation speed and stability.

I propose to make some changes in the code such that this simplification is always made. To this end I would like to change ```Annex60.Fluid.Interfaces.StaticTwoPortConservationEquation```, ```Buildings.Fluid.BaseClasses.PartialResistance```, ```Annex60.Fluid.Sensors.BaseClasses.PartialFlowSensor``` and similar models. 

In the case of ```Annex60.Fluid.Sensors.BaseClasses.PartialFlowSensor``` the change would be to replace

```
  port_a.h_outflow = inStream(port_b.h_outflow);
  port_b.h_outflow = inStream(port_a.h_outflow);
  port_a.Xi_outflow = inStream(port_b.Xi_outflow);
  port_b.Xi_outflow = inStream(port_a.Xi_outflow);
  port_a.C_outflow = inStream(port_b.C_outflow);
  port_b.C_outflow = inStream(port_a.C_outflow);
```

by
```
  port_a.h_outflow = if not allowFlowReversal then Medium.h_default else inStream(port_b.h_outflow);
  port_b.h_outflow =  inStream(port_a.h_outflow);
  port_a.Xi_outflow = if not allowFlowReversal then Medium.Xi_default else  inStream(port_b.Xi_outflow);
  port_b.Xi_outflow = inStream(port_a.Xi_outflow);
  port_a.C_outflow = if not allowFlowReversal then Medium.C_default else  inStream(port_b.C_outflow);
[...]

PR #312, opened 2015-09-14
implemented more efficient code when allowFlowReversal = false for Pa…

Mathadon:
…rtialResistance and PartialFlowSensor

rewrote equations in StaticTwoPortConservationEquation for better readability
this is for #281


Mathadon:
Unit tests in the Fluid package gave the same results

Issue #316, opened 2015-09-19
Control components


Mathadon:
For my PhD I have created various control component models. Some of them are quite specialist and do not belong in Annex 60 in my opinion. However some are probably interesting to add:
- heating curve model
- hysteresis model with some additional options
- a 'minimum on time' block

Then I have some integer/boolean blocks that seem to be missing from MSL, but that may not be relevant to add:
- integer switch
- integer 'contains': i.e. returns true/false if an input integer i is in a parameter set of integers {i1, i2, ...}
- integer greater than comparison
- integer delay block
- integer equality comparison
- RSFlipFlop (not that same as MSL since reset/set dominance differs)

I'll already create a pull request for the heating curve model since I assume this is of interest. Other models I'll process based on feedback to this issue. Is there an interest to add these other models?

mwetter:
Make a branch `issue316_control` and each developer should put their models on that branch for review and eventual integration into the `Annex60` library.

Mathadon:
done

PR #317, opened 2015-09-19
added heating curve model, unit test, documentation, reference result

Mathadon:
for #316


Mathadon:
This is superseded by #366

PR #318, opened 2015-09-19
added hysteresis controller model, unit test, reference result and example

Mathadon:
for #316


Mathadon:
This is superseded by #367

PR #326, opened 2015-10-02
Issue324 default annotations

mwetter:
This will merge the changes from https://github.com/iea-annex60/modelica-annex60/pull/325

I reviewed the changes and run the regression tests.
I created a ticket at https://github.com/lbl-srg/BuildingsPy/issues/68 to avoid writing trailing white spaces in the reference result files.


tbeu:
As only annotations were deleted the regression tests should remain untouched.

Issue #264, opened 2015-06-05
Add pipe with heat loss


nytschgeusen:
The majority of the case studies of Activuty 2.1 used a thermal pipe model, describing the thermal losses and pressure losses within thermal loops. Up to now the Annex 60 library owns only a adiabatic pipe (Annex60.Fluid.FixedResistances-LosslessPipe). E.g. the the pipe model from the Buildings library or a similar one would fill this gap.

mwetter:
Created branch `issue264_pipe` for integration of pipe model.

marcusfuchs:
I pushed a first implementation of an adiabatic pipe model using the spatialDistribution operator to this branch. In a next step, different options for heat transfer can be tested based on this model. I also included an example that demonstrates the different mass flow rates optained with 2 hydraulic resistances from pipes of 50 m length each compared to a single pipe of 100 m length. The files are in the Experimental package.

marcusfuchs:
I just pushed the pipe with heat loss implementation presented at the Leuven meeting. This can serve as the foundation for the working group on pipe modeling.

bramvdh91:
I have just added our previous implementation to my local version of the Annex 60 library. I would like to push it, but I get an error "Unable to access 'https://github.com/iea-annex60/modelica-annex60.git/': The requested URL returned error: 403". Could it be that I first need permission to push to the issue264_pipe branch?

Mathadon:
You need to have write access before you can push to the annex60 repository directly. It's probably best to fork the library onto your own account and push your changes there and post a link here :)

bramvdh91:
Thanks for the advice. I've pushed our files to https://github.com/bramvdh91/modelica-annex60/tree/issue264_pipe. I see that a few improvements are still needed to solve some inheritance issues, but the model structure is there already.

marcusfuchs:
In preparation to the TelCo on the pipe models next week I started a few comparisons between the two implementations (The comparison models are in the pipe examples on the branch. It seems like the KUL implementation, as expected, is ahead on the heat loss implementation while the A60 model seems to behave better at flow reversal. For the simplest case of the lossless pipe, with the same mass flow rate, the flow reversal does not show in the inlet become outlet temperature of the KUL model:

![01_mass_flow_rate](https://cloud.githubusercontent.com/assets/5516900/10226362/3accd51c-6867-11e5-83f6-02864c6823a3.png)
[...]

PR #329, opened 2015-10-28
Issue264

carlesRT:


carlesRT:
Models + Example to test abs() and show spatialDistribution noise #264

PR #335, opened 2015-11-05
Reorder `public`

jmoeckel:
To improve readability all  `public` variables are declared before `protected` ones.   
The top `public` command does not change any functionality, so it has been deleted.


mwetter:
@jmoeckel : Thanks for the changes. I will test them and then merge to the `master`.

Issue #351, opened 2015-11-13
Consider removing port_a_exposesState and port_b_exposesState


mwetter:
Consider removing in `PartialTwoPort` the parameters
`port_a_exposesState` and `port_b_exposesState`.
I don't think any library is using this feature.

mwetter:
I suggest to remove these parameters as they cannot be dynamically changed on the diagram layer. For example, change in `Movers.BaseClasses.PartialFlowMachine` the `extends` statement to be
```  
extends Buildings.Fluid.Interfaces.PartialTwoPortInterface(show_T=false,
    port_a_exposesState=energyDynamics <> Modelica.Fluid.Types.Dynamics.SteadyState,
```
Looking at `Movers.Validation.PumpCurveConstruction` shows that the icon still shows the graphical annotation despite that the pumps are configured to be steady-state.

I therefore suggest to remove these parameters. What is the position of Aachen, UdK and Leuven?

mlauster:
From Aachen side, we are using this, but I'm not sure yet if it works in our cases. We need to check this.

mlauster:
Aachen should be fine with removing this.

Mathadon:
Ok, we never use it!

carlesRT:
I think we also do not use it.

Mathadon:
This seems to be merged.

PR #360, opened 2015-11-19
Addressed annotations for #349 and #351

### LINKED ISSUE #349: Remove showDesignFlowDirection
The flag `showDesignFlowDirection` is in my view not needed as it should be clear that the flow is from port a to port b. Also, this feature is not consistently used now by models with two ports, and it is not used in model with four ports.

I therefore propose to remove this graphical annotation. What are the positions of Aachen, Leuven and UdK?

mwetter:
This will close #349 and #351.
I added `visible=not allowFlowReversal` based on the recommendation of @carlesRT 
All unit tests run successfully.


Mathadon:
Will try to review this tonight or next week!

PR #361, opened 2015-11-20
Issue354 movers

### LINKED ISSUE #354: Merge Fluid.Movers from Buildings
This issue is to merge the revised movers to the Annex 60 library.

The package `Fluid.Movers` has been significantly revised in https://github.com/lbl-srg/modelica-buildings/issues/288 and in https://github.com/lbl-srg/modelica-buildings/issues/457.

They now allow a constant control signal, or a stage signal in addition to a continuous control signal. Also, parameters that were exposed twice, in the `record` and as a regular `parameter`, have been revised.

The use of `power` is no longer possible in `Fluid.Movers.FlowControlled_dp` and `Fluid.Movers.FlowControlled_m_flow` as it would require the speed of the pump which is not known in these two models.

mwetter:
This will close #354.

The revision was done with @Mathadon and input from @bknielsen.
It merges the changes that were required for
https://github.com/lbl-srg/modelica-buildings/issues/457
and for
https://github.com/lbl-srg/modelica-buildings/issues/288
to the Annex 60 library.

All unit tests run successfully, some with fewer nonlinear equations during the initialization. I also merged the files back to the `Buildings` library to verify that everything is merged properly to this branch.

As @Mathadon was involved in the revisions, I will assign the review to @mlauster .


Mathadon:
Wohoo!

mlauster:
I get warnings regarding four models when checking:
SpeeControlled_Nrpm
SpeedControlled_y
BaseClasses.FlowMachineInterface
BaseClasses.SpeedControlled

All seem to originate from the same "problem":

unbenanntes bild

Anyway, the message says that this might be okay. Is this warning acceptable?

Mathadon:
The warning is removed when you use the model in a real example, so for me this is ok!

mwetter:
This warning goes away once the model is used in an example, e.g., when the parameters are assigned. I don't see a practical way to avoid this warning, so I am fine with it.

PR #362, opened 2015-11-20
Replaces `NumberOfIntervals`

jmoeckel:
The flag `NumberOfIntervals` within annotations is not standardized and throws an error in pedantic mode. 

This pull request replaces `NumberOfIntervals` with flag `Interval` (`= (StopTime - StartTime)/NumberOfIntervals`)

### DIFF
--- a/Annex60/Experimental/Benchmarks/AirFlow/Examples/ZoneStepResponse.mo
+++ b/Annex60/Experimental/Benchmarks/AirFlow/Examples/ZoneStepResponse.mo
@@ -66,7 +66,7 @@ equation
   connect(step.y, weaDat.TDryBul_in) annotation (Line(
       points={{-79.1,-33},{-63,-33}},
       color={0,0,127}));
-  annotation (    experiment(StopTime=6e+006, __Dymola_NumberOfIntervals=5000),
+  annotation (    experiment(StopTime=6e+006, Interval=200),
     __Dymola_Commands(file="modelica://Annex60/Resources/Scripts/Dymola/Experimental/Benchmarks/AirFlow/Examples/ZoneStepResponse.mos"
         "Simulate and plot"),
     Documentation(info="<html>
--- a/Annex60/Experimental/Benchmarks/AirFlow/Examples/ZoneStepResponse.mo
+++ b/Annex60/Experimental/Benchmarks/AirFlow/Examples/ZoneStepResponse.mo
@@ -66,7 +66,7 @@ equation
   connect(step.y, weaDat.TDryBul_in) annotation (Line(
       points={{-79.1,-33},{-63,-33}},
       color={0,0,127}));
-  annotation (    experiment(StopTime=6e+006, Interval=200),
+  annotation (    experiment(StopTime=6e+006, __Dymola_NumberOfIntervals=5000),
     __Dymola_Commands(file="modelica://Annex60/Resources/Scripts/Dymola/Experimental/Benchmarks/AirFlow/Examples/ZoneStepResponse.mos"
         "Simulate and plot"),
     Documentation(info="<html>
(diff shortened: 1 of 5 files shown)


Annex60/Experimental/Benchmarks/AirFlow/Examples/ZoneStepResponse.mo
```diff
@@ -66,7 +66,7 @@ equation
   connect(step.y, weaDat.TDryBul_in) annotation (Line(
       points={{-79.1,-33},{-63,-33}},
       color={0,0,127}));
-  annotation (    experiment(StopTime=6e+006, __Dymola_NumberOfIntervals=5000),
+  annotation (    experiment(StopTime=6e+006, Interval=200),
```
mwetter: The change of the dymola specific annotation to `Interval` is good, but can you please change the 200 back to 5000. I think @marcusfuchs requires such high a number for his benchmarks

Annex60/Fluid/Examples/Performance/Example1v1.mo
```diff
@@ -71,10 +71,10 @@ For this script to work, make sure that Dymola stores at least 4 results.
 cpuOld=OutputCPUtime;<br/>
 evaluateOld=Evaluate;<br/>
 OutputCPUtime:=true;<br/>
-simulateModel(\"Annex60.Fluid.Examples.Performance.Example1v1(allowFlowReversal.k=true, from_dp.k=false)\", stopTime=10000, numberOfIntervals=10, method=\"dassl\", resultFile=\"Example1v1\");<br/>
-simulateModel(\"Annex60.Fluid.Examples.Performance.Example1v2(from_dp.k=true, allowFlowReversal.k=true)\", stopTime=10000, numberOfIntervals=10, method=\"dassl\", resultFile=\"Example1v2\");<br/>
-simulateModel(\"Annex60.Fluid.Examples.Performance.Example1v1(allowFlowReversal.k=false, from_dp.k=false)\", stopTime=10000, numberOfIntervals=10, method=\"dassl\", resultFile=\"Example1v1\");<br/>
-simulateModel(\"Annex60.Fluid.Examples.Performance.Example1v1(allowFlowReversal.k=false, from_dp.k=true)\", stopTime=10000, numberOfIntervals=10, method=\"dassl\", resultFile=\"Example1v1\");<br/>
+simulateModel(\"Annex60.Fluid.Examples.Performance.Example1v1(allowFlowReversal.k=true, from_dp.k=false)\", stopTime=10000, Interval=1000, method=\"dassl\", resultFile=\"Example1v1\");<br/>
```
mwetter: These changes should be reverted as these lines are the Dymola-specific scripts. They have a different syntax than the MSL, and Dymola uses `numberOfIntervals` in its scripting language.

The same applies to the other examples.

mwetter:
@jmoeckel : Thanks for the pull request. Please see the above comments.

You wrote "The flag NumberOfIntervals within annotations is not standardized and throws an error in pedantic mode". How did you get this error? I set `Advanced.PedanticModelica=true;` but do not get an error in Dymola 2016. I am interested in reproducing it so we can add it to our regression tests.

Annex60/Fluid/Examples/Performance/Example1v1.mo
jmoeckel: Ok, see your point. Actually I did not know, that Dymola-specific script syntax differs from the MSL. I will revert these changes (as well as for the other examples).

jmoeckel:
I have reverted the changes for the examples. Actually I did not know, that the Dymola script language differs from the MLS - especially for the same flags. 

This is also the reason, why no `error` in pedantic mode had been thrown. I assumed, that these scripts, once they were executed as a standalone application, would fail. 

The main thought of this pull request remains: The combination of the flag 'NumberOfIntervals' with pedantic modelica will throw an error (in the MLS). My fault was, that I did not differentiate between the mentioned usecases. I am sorry, if my mistake costs you unnecessary time. 

Ironically, the one change, that makes sense (replacing `__Dymoly_NumberOfIntervals`) just has been a kind of side-product of this pull request, as this useage is Ok as long as one uses the model within Dymola. 

Should this pull request therefore be closed and the one change proposed again with a different title?

mwetter:
@jmoeckel Thanks for the clarification. Don't worry about the mistake, we are all learning.

I will merge your pull request to the `master` branch. I also verified that this is the only file with this annotation.

PR #366, opened 2015-11-24
added heating curve model, unit test, documentation, reference result

Mathadon:
This is for #316


Mathadon:
This issue is closed for now. It may be picked up again from the reference in #316

PR #367, opened 2015-11-24
added hysteresis controller model, unit test, reference result and ex…

Mathadon:
…ample

for #316


Mathadon:
This issue is closed for now. It may be picked up again from the reference in #316

PR #368, opened 2015-11-27
Fix typos

tbeu:
See also UdK-VPT/BuildingSystems#28.


mwetter:
@tbeu : Thank you for the fixes. I will merge to the master.

PR #371, opened 2015-11-30
Issue370 efficiency

mwetter:
This adds more changes to #369.
It also removes specifying the <code>min=0</code> as this is already done at the type declaration.

@thorade Can you please review and let me know whether I caught all occurrences. All unit tests work.


thorade:
Found one more occurence:
https://github.com/iea-annex60/modelica-annex60/blob/master/Annex60/Fluid/MassExchangers/ConstantEffectiveness.mo
Otherwise it looks good and can be merged.

PR #373, opened 2015-12-02
Issue372 co2

Mathadon:
I created a first implementation for a changed mixing volume implementation. Let's see if we can agree on the basics before continuing with documentation and unit tests!

### DIFF
--- a/Annex60/Fluid/Interfaces/ConservationEquation.mo
+++ b/Annex60/Fluid/Interfaces/ConservationEquation.mo
@@ -60,7 +60,9 @@ model ConservationEquation "Lumped volume with mass and energy balance"
   Modelica.Blocks.Interfaces.RealInput mWat_flow(unit="kg/s")
     "Moisture mass flow rate added to the medium"
     annotation (Placement(transformation(extent={{-140,0},{-100,40}})));
-
+  Modelica.Blocks.Interfaces.RealInput[Medium.nC] C_flow(unit="kg/s")
+    "Trace substance mass flow rate added to the medium"
+    annotation (Placement(transformation(extent={{-140,-40},{-100,0}})));
   // Outputs that are needed in models that extend this model
   Modelica.Blocks.Interfaces.RealOutput hOut(unit="J/kg",
                                              start=hStart)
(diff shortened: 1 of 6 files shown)


Mathadon:
Note that I was able to run unit tests successfully while there was a singularity in the equations of `StaticTwoPortConservationEquation`. So there seem to be no unit tests yet that use trace substances.

Annex60/Fluid/Interfaces/ConservationEquation.mo
```diff
@@ -60,7 +60,9 @@ model ConservationEquation "Lumped volume with mass and energy balance"
   Modelica.Blocks.Interfaces.RealInput mWat_flow(unit="kg/s")
     "Moisture mass flow rate added to the medium"
     annotation (Placement(transformation(extent={{-140,0},{-100,40}})));
-
+  Modelica.Blocks.Interfaces.RealInput[Medium.nC] C_flow(unit="kg/s")
```
mwetter: @Mathadon : I will remove the unit declaration as for PPM, the unit would be wrong.

Annex60/Fluid/Interfaces/ConservationEquation.mo
```diff
@@ -236,7 +239,7 @@ equation
   if traceDynamics == Modelica.Fluid.Types.Dynamics.SteadyState then
     zeros(Medium.nC)  = mbC_flow;
```
mwetter: @Mathadon : I will add C_flow to the steady-state equation because what is added from the connector has to leave from the ports.

Annex60/Fluid/Sensors/TraceSubstancesPpm.mo
```diff
@@ -0,0 +1,69 @@
+within Annex60.Fluid.Sensors;
+model TraceSubstancesPpm
+  "Ideal one port trace substances sensor outputting in parts per million"
+  extends Annex60.Fluid.Sensors.BaseClasses.PartialAbsoluteSensor;
+  extends Modelica.Icons.RotationalSensor;
+  parameter String substanceName = "CO2" "Name of trace substance";
+
+  parameter Real MM_fraction = 44/29
```
mwetter: @Mathadon : This should by 29/44, not 44/29. I will correct it and add a regression test.

Mathadon: I noticed this in my own branch but forgot to push it, my apologies.

PR #384, opened 2016-01-06
Issue383 run unit tests

### LINKED ISSUE #383: expose commonly used option of bin/runUnitTests.py
I often set the option for

* the number of processors,
* single packages to be validated, and
* batch mode
manually, which results in a change to `bin/runUnitTests.py`.

This issue is opened to revise this file so that these options can be specified as command line arguments.

mwetter:
This will close #383.
All changes are backwards compatible, and no change is needed to `BuildingsPy`.

### DIFF
--- a/bin/runUnitTests.py
+++ b/bin/runUnitTests.py
@@ -4,7 +4,7 @@
 # only checks the html syntax.
 #
 # To run the unit tests, this script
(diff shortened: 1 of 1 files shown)


marcusfuchs:
Thanks, this is a really useful improvement and worked perfectly in my tests. I only fixed one small typo.

PR #388, opened 2016-01-08
Issue156 pre ind val

Mathadon:
I did not use the smoothmax function since the overshoot of this function results in a non-increasing pressure drop function:
screen shot 2016-01-08 at 07 49 43
(see slope at tooltip in figure above)

also I added a min() function. Otherwise the proposed function had an asymptotic limit, which causes the function not to be defined/invertible for large mass flow rates.


Mathadon:
I also added a new unit test that checks if the function can be solved when the mass flow rate is prescribed.

Issue #396, opened 2016-01-19
New, staged pump records


Mathadon:
I have some records for the Wilo TOP-S series (http://productfinder.wilo.com/en/COM/productrange/000000090002c94500020023/fc_range_description?lang=en&iwp%5B%5D=powertree-R-0).

These are staged pumps and it seems we have not foreseen a proper way of handling this in `Annex60.Fluid.Movers.Data.SpeedControlled_Nrpm`. In my opinion it should contain:
```
  parameter Real[:] speeds(each final unit="1/min") = {0} 
    "Vector of speed set points when using stages";
```

since it's a property of the pump.

I propose to add this so I can add the records for these pumps correctly.

I'm not sure if we should also add

```
  parameter Real[:] normalized_speeds(each final unit="1") = {0} 
    "Vector of normalized speed set points when using stages"
    annotation(Dialog(enable=inputType == Annex60.Fluid.Types.InputType.Stages));
```
to `Annex60.Fluid.Movers.Data.SpeedControlled_y`.

I'll already make a pull request without this.

mwetter:
I have the following comments in addition to the comments in the pull request (https://github.com/iea-annex60/modelica-annex60/pull/397/files)

- [ ] In `SpeedControlled_Nrpm`, the parameter `N_nominal` and `speeds` (which should be called `N`) should be assigned from the record `per` as they are a property of the pump.
- [ ]  The same should be done for `SpeedControlled_y` with `normalized_speed` and `normalized_speeds`, which need to be added to the record definition and given default values. Should we rather call this `n_nominal` and `n` to indicate that it is dimensionless? `normalized_speed` does not indicate anything about being for the nominal conditions and hence it looks like something different from `N_nominal`. Since `Data.SpeedControlled_Nrpm` extends `Data.SpeedControlled_y`, it needs to assign `n` with the final keyword so that these entries are consistent.

Mathadon:
Regarding your second suggestion: if we add parameter `n_nominal` then I think we should change the implementation of `speedControlled_y` since it currently assumes that the speed/efficiency/power curve is defined for `n_nominal=1`. The user may/should interpret the value of `n_nominal` as the normalised speed for which he/she provides the pump curves and not changing it may lead to wrong model use. Do you agree we should add this too?

mwetter:
I suggest to

[...]

PR #397, opened 2016-01-19
added records for Wilo TOP-S pumps

Mathadon:
added parameter in SpeedControlled_Nrpm to allow this
changed default value in SpeedControlled_Nrpm

this is for #396 

Unit tests in /Movers run ok


Annex60/Fluid/Movers/Data/SpeedControlled_Nrpm.mo
```diff
@@ -5,12 +5,21 @@ record SpeedControlled_Nrpm "Generic data record for FlowMachine_Nrpm"
   parameter Modelica.SIunits.Conversions.NonSIunits.AngularVelocity_rpm N_nominal = 1500
     "Nominal rotational speed for flow characteristic";
 
+  parameter Real[:] speeds(each final unit="1/min") = {0}
```
mwetter: - The units should be `Modelica.SIunits.Conversions.NonSIunits.AngularVelocity_rpm` and not `Real`.
- Because we used `N_nominal`, this should be `N` as it is the physical quantity for which `N_nominal` is its nominal value, no?

Annex60/Fluid/Movers/SpeedControlled_Nrpm.mo
```diff
@@ -16,7 +16,7 @@ model SpeedControlled_Nrpm
   parameter Real speed(final unit="1/min") = 0
     "Speed set point when using constant set point"
     annotation(Dialog(enable=inputType == Annex60.Fluid.Types.InputType.Constant));
-  parameter Real[:] speeds(each final unit="1/min") = {0}
+  parameter Real[:] speeds(each final unit="1/min") = per.speeds
```
mwetter: This should be `N`, or maybe better `Nrpm` to identify that this is not in a SI unit.

Annex60/Fluid/Movers/Data/SpeedControlled_Nrpm.mo
Mathadon: For 1): I copied the implementation from `Annex60.Fluid.Movers.SpeedControlled_Nrpm` so I'll change it there too.
For 2): the variable name `N` does not really reflect that the variable is an array. Also to be consistent we'd have to change `Annex60.Fluid.Movers.SpeedControlled_Nrpm`, which is non-backwards compatible. I also think it is easier to interpret the meaning of the parameter when using `speeds`. So I would propose to stick with the name `speeds`?

Annex60/Fluid/Movers/SpeedControlled_Nrpm.mo
mwetter: I changed it to `speeds_rpm` to indicate that these speeds have non-SI units. Similarly to the situation where we use `T_degC`.

Annex60/Fluid/Movers/Data/SpeedControlled_Nrpm.mo
```diff
@@ -1,9 +1,18 @@
...
-  parameter Modelica.SIunits.Conversions.NonSIunits.AngularVelocity_rpm N_nominal = 1500
-    "Nominal rotational speed for flow characteristic";
+  parameter Modelica.SIunits.Conversions.NonSIunits.AngularVelocity_rpm constantSpeed_rpm=
+    speeds_rpm[size(speeds_rpm, 1)]
+    "Speed set point when using inputType = Annex60.Fluid.Types.InputType.Constant";
+
+  parameter Modelica.SIunits.Conversions.NonSIunits.AngularVelocity_rpm[:] speeds_rpm = {1500}
+    "Vector of speed set points when using inputType = Annex60.Fluid.Types.InputType.Stages";
+
+  final parameter Modelica.SIunits.Conversions.NonSIunits.AngularVelocity_rpm speed_rpm_nominal=
```
Mathadon: I think it would make sense if this were not `final` since in some cases the highest speed curve does not allow to set up the `pressure` record, since data in the data sheet is incomplete due to a power limiting function. See for instance:
http://productfinder.wilo.com/repo/media/previews/00626996_0.jpg 
in this case it would be useful if the user would be able to override `speed_rpm_nominal`?

Annex60/Fluid/Movers/Data/SpeedControlled_y.mo
```diff
@@ -3,6 +3,16 @@ record SpeedControlled_y
...
 
+//  parameter Real speed_nominal(final min=0, final unit="1") = 1
+//    "Nominal rotational speed for flow characteristic";
+
+  parameter Real constantSpeed(final min=0, final unit="1")=
+    speeds[size(speeds, 1)]
+    "Speed set point when using inputType = Annex60.Fluid.Types.InputType.Constant";
+
+  parameter Real[:] speeds(each final min = 0, each final unit="1") = {1}
+    "Vector of speed set points when using inputType = Annex60.Fluid.Types.InputType.Stages";
```
Mathadon: Note to self: clarify comment that this should be 'normalised' speeds.

Annex60/Fluid/Movers/Examples/MoverParameter.mo
```diff
@@ -54,12 +54,13 @@ model MoverParameter
     redeclare package Medium = Medium,
     dynamicBalance=false,
     filteredSpeed=false,
-    speeds={0,1000,2000},
-    redeclare Annex60.Fluid.Movers.Data.Pumps.Wilo.Stratos25slash1to4 per,
-    inputType=Annex60.Fluid.Types.InputType.Constant,
-    speed=2000) "Pump with speed input"
+    redeclare Annex60.Fluid.Movers.Data.Pumps.Wilo.Stratos25slash1to4 per(
+      speeds_rpm={0,1000,2000},
+      constantSpeed_rpm=2000),
```
Mathadon: I think this causes the different unit test results you mention in #397 . The speed for which the pressure curve is defined is 1800 rpm. By changing `constantSpeed_rpm` you also change the speed for which the curve is defined, which it should not do. As illustrated by this problem I still think it is a good idea to set `speed_rpm_nominal` in the record itself, instead of assigning it the value of `speeds_rpm[size(speeds_rpm, 1)]`?

Annex60/Fluid/Movers/FlowControlled_dp.mo
```diff
@@ -25,6 +25,7 @@ model FlowControlled_dp
     "Nominal pressure raise, used to normalized the filter if filteredSpeed=true"
     annotation(Dialog(group="Nominal condition"));
 
+  // fixme: shouldn't the heads not also come from the performance record?
```
Mathadon: This still needs to be sorted out. I would not involve the performance records in this parameter.

Annex60/Fluid/Movers/FlowControlled_m_flow.mo
```diff
@@ -20,7 +20,7 @@ model FlowControlled_m_flow
   parameter Modelica.SIunits.MassFlowRate m_flow_start(min=0)=0
     "Initial value of mass flow rate"
     annotation(Dialog(tab="Dynamics", group="Filtered speed"));
-
+    // fixme: shouldn't the massFlowRates not also come from the performance record?
```
Mathadon: This still needs to be sorted out. I would not involve the performance records in this parameter.

Annex60/Fluid/Movers/Examples/MoverParameter.mo
mwetter: I agree and added `speed_rpm_nominal` and `speed_nominal`. Now the unit tests are correct again.

mwetter:
@Mathadon 
This is now ready for review. Please merge it to the `master` if you agree with all the changes.
All unit tests produce the same results.

Issue #353, opened 2015-11-13
Annex60.Fluid.Interfaces.FourPortHeatMassExchanger is not used


mwetter:
The model `Annex60.Fluid.Interfaces.FourPortHeatMassExchanger` is not used anywhere in the library, and hence it is also not part of any regression test.
I suggest to keep the model in the library as it is similar to `TwoPortHeatMassExchanger` and libraries the are based on `Annex60` are using it, for example in `Buildings.Fluid.Chillers.Carnot`.

To avoid breaking the model, a regression test needs to be added.

mwetter:
If `Carnot` where to be moved to `Annex60`, it should be called `Carnot_y`.

All should comment whether this should be moved to the `Annex60` library.

thorade:
Is this related to https://github.com/iea-annex60/modelica-annex60/issues/223 ??

mwetter:
#223 is much bigger as its main purpose is to model an actual heat pump, whereas the Carnot model is idealized and meant for applications where no data of the heat pump is available, e.g., it is solely based on thermodynamics 2nd law, plus an efficiency correction that optionally takes the load as an input.

damienpicard:
According to me, Annex60 could benefit of having the `Buildings.Fluid.Chillers.Carnot` model. I've revised the model and I've listed my comments in https://github.com/lbl-srg/modelica-buildings/issues/471. Once these comments are adressed and if everybody agrees, the Carnot model can be added.

damienpicard:
As discussed in lbl-srg/modelica-buildings/issues/471, I did the following changes:
. Add a Carnot heat pump, and rename these models to Carnot_y (similarly than for the heater/cooler) to avoid a clash in names.
. remove actualStream in computing the states that are used to compute the Carnot efficiency.

I also added the example from the building library and i've updated the reference result.

I've pushed the changes on iea-annex60/modelica-annex60/tree/issue353_addCarnot_y

mwetter:
@damienpicard 
I will also integrate the `Chiller.Carnot_TEva`, `HeatPumps.Carnot_TCon` and `HeatPumps.Carnot_y`.

I will also correct the COP computation so that the useful heat is the heat of the evaporator for the chiller, and the heat of the condenser for the heat pump.

[...]

PR #400, opened 2016-01-20
Issue353 add carnot y

damienpicard:
Add carnot machine to master.


mwetter:
@damienpicard : I added Carnot heat pumps and chillers that take the leaving water temperature or the compressor speed as an input signal. I started with your models, and then refactored them as we have now 4 models that share similar functionalities. I also added many validation tests.
All unit tests work fine. Can you please review and then merge to the master.

This will close #353

damienpicard:
@mwetter `Annex60.Fluid.Chillers.BaseClasses.PartialCarnot_y` and  `Annex60.Fluid.Chillers.BaseClasses.PartialCarnot_T` are very similar but due to the fact that the first extend from `Interfaces.FourPortHeatMassExchanger` and the other not oblige you to have quite some code duplication.

I wonder if it would not make more sense to use the implementation of `Annex60.Fluid.Chillers.BaseClasses.PartialCarnot_T` for both models and redeclaring `con` with `HeatExchangers.HeaterCooler_u`? If not, making an extra partial with only the parameters of `Interfaces.FourPortHeatMassExchanger` could also improve the current code.

damienpicard:
otherwise, the models are good.

mwetter:
@damienpicard Thanks for pointing this out. I will try to
 - [ ] use `Annex60.Fluid.Chillers.BaseClasses.PartialCarnot_T` as the base class for both models.

I think this should yield a clearer implementation.

mwetter:
@damienpicard 
I refactored the heat pumps and chiller model so that there is less code duplication. The implementation actually got simpler. Thanks for this comment.

Can you please review https://github.com/iea-annex60/modelica-annex60/commit/2a1e8c147e2f35a8c8a3e27cebc6004232e26c58 so we can merge it to the master branch?

PR #409, opened 2016-01-27
Issue408 modelica compliance

### LINKED ISSUE #408: Correct wrong Modelica syntax
Some classes contain Modelica syntax that is not valid and does therefore not compile with JModelica. This issue is to correct these problems.

Also some units, primarily in `Sensors` and `Sources` are not set, which will also be corrected on this branch.

mwetter:
This will close #408.
All unit tests work.


thorade:
Looks good to me (just reading the code, no testing)

PR #414, opened 2016-02-15
Update examples and fix DoublePipe #264

bramvdh91:
I have updated the example section. I think the event comparisons are no longer needed since the component with delay tracking inside HeatLoss component has been moved to archive. Therefore, I moved these examples to Archive as well.

I noticed issues with the DoublePipe because of an updated `PartialFourPort`. This partial model distinguishes between Medium1 and Medium2 and also two allowFlowReversal flags. I tried to implement the option of having two media, but I doubt it is useful for twin pipes, since we assume the same medium is flowing through the pipes in opposite direction and equal velocity anyway. Problems pop up in defining different heat capacity values for the two sides and setting the standard thermodynamic state (only for one of the Media).
I propose to set allowFlowReversal2 and Medium2 as dummy parameters and not use them in the rest of the component:
```
extends Annex60.Fluid.Interfaces.PartialFourPort(
   final Medium2 = Annex60.Media.Water,
   final allowFlowReversal2 = True);
```
What do you think?

@marcusfuchs , can you check this request?


carlesRT:
>What do you think?

If this is the case, the media should be the same then go ahead. But it seems to be an unwritten rule, If the model has 1 medium -> Medium is used. If the model has 2 Medium -> Medium1 and Medium2 are used. So,  I would set Medium1= Medium and Medium2 = Medium and  declare a new Medium.

bramvdh91:
I agree. I will make these changes and reopen the pull request.

bramvdh91:
Updated pull request.

marcusfuchs:
Thank you @bramvdh91 for the updated examples. I noticed the following:

- `Annex60.Experimental.Pipe.Archive.Examples.DoublePipe.IndependentSupplyAndReturn` cannot find the `doublePipe` component. Should this be changed to the current double pipe model? Or is the example no longer needed? Then we should delete it entirely.
- In `Annex60.Experimental.Pipe.Types.ThermalResistanceLength`, `Annex60.Experimental.Pipe.Types.ThermalCapacityPerLength` and `Annex60/Experimental/Pipe/Examples/Comparisons/Events/DelayInHeatloss.mo` the html-formating `<span style=\"font-family: MS Shell Dlg 2;\">` is not necessary and could be removed
- The `package.order` in `Annex60.Experimental.Pipe.Examples.Comparisons` does not list the `Events` package, therefore Dymola does not show the examples `DelayInHeatloss.mo` and `DelayInPipe.mo`. Should those be deleted?

bramvdh91:
Hey @marcusfuchs , thanks for reviewing.
About the archived example packages: I think we can delete them. They were meant to compare implementations with multiple delay operators and a single one, and we decided to throw the first ones out. (so both ``Annex60.Experimental.Pipe.Archive.Examples.DoublePipe.IndependentSupplyAndReturn`` and ``Annex60.Experimental.Pipe.Examples.Comparisons``.
About the HTML tag, I'm not sure how this happened, but I use Dymola 2016's editor so I don't have to worry about the syntax, this probably explains unnecessary tags.
I will implement your suggestions!

bramvdh91:
Another update. I added @marcusfuchs 's proposed changes and finalized updating docstrings in the examples package.

marcusfuchs:
Looks good, thanks @bramvdh91 !

Issue #247, opened 2015-05-22
Conservation of quantities


Mathadon:
We currently do not have (many) unit tests that check if ```MixingVolumes``` conserve mass and energy. I'll make some suggestions that we can have a look at.

Mathadon:
Can we have a look at 
```Annex60.Fluid.MixingVolumes.Validation.TracesubstanceConservation```
and 
```Annex60.Fluid.MixingVolumes.Validation.MoistureMixingConservation```
in https://github.com/Mathadon/modelica-annex60/tree/issue247_conservationTests

The first test fails because ```Annex60.Fluid.Sensors.TraceSubstancesTwoPort``` seems to contain a bug:
```
CMed = s*inStream(port_b.C_outflow);
```
should be
```
CMed = s*inStream(port_a.C_outflow);
```
After fixing this the model still fails. This is because the mass flow rate calculation in ```Annex60.Fluid.Sensors.TraceSubstancesTwoPort``` does not take into account that a part of ```m_flow``` is water vapour. Adding water vapour therefore also increases the mass flow rate of the CO2.

The second test fails because stream connector mixing points assume that ```m_flow``` in a fluid connector is the dry air mass flow rate. 

Both problems can be fixed by setting 
```
  if massDynamics == Modelica.Fluid.Types.Dynamics.SteadyState then
    0 = mb_flow;
  else
    der(m) = mb_flow;
  end if;
``` 
in
```
Annex60.Fluid.Interfaces.ConservationEquation
```
and 
```
port_a.m_flow + port_b.m_flow = 0;
```
in
```
Annex60.Fluid.Interfaces.StaticTwoPortConservationEquation
```. 
This also relates to issue #234 .

Mathadon:
I have not yet implemented the bug fixes so we can test this first and then decide on how to solve it.

mwetter:
@Mathadon 
I slightly revised the models on `issue247_conservationTests` of the `Annex60` repository.
The test setups are good, but there also need to be tests that check the implementation of the dynamic balance. 

Can you please implement how you think the mass balance should be treated, add `.mos` scripts for the new examples and clean up the examples (such as naming convention. expanding the drawing area so the model fits into it etc).

There will also need to be a documentation that if moisture is added, then the trace substance concentration does not change as it is mass fraction per unit mass of dry air and hence adding moisture does not affect the mass flow rate `m_flow`.

Mathadon:
[...]

PR #427, opened 2016-03-06
Issue247 conservation testsv2

Mathadon:
This adds new unit tests for #247 

It also fixes conservation of mass/quantities for mixing volumes. The new implementation contains the following lines of code:
```
    port_b.Xi_outflow = -(inStream(port_a.Xi_outflow)*port_a.m_flow + mXi_flow) * m_flowInv_b;
    port_a.Xi_outflow = if allowFlowReversal then -(inStream(port_b.Xi_outflow)*port_b.m_flow + mXi_flow) * m_flowInv else Medium.X_default[1:Medium.nXi];
```
the consequence is that when `m_flow=0`, `Xi_outflow` becomes zero since `m_flowInv=0`. This was not the case before and results in different unit test results in `Annex60.Fluid.MixingVolumes.Examples.MixingVolumePrescribedHeatFlowRate`. I'm unsure how to proceed with this behaviour. I see following options:
- accept this new behaviour
- use the old equations and accept that mass is not conserved
- write a custom regularisation that has `port_b.Xi_outflow = inStream(port_a.Xi_outflow)` for `m_flow = 0`

This pull request also adds new outputs in `PartialMixingVolume`. These can also be used for checking conservation of energy.


Mathadon:
I did not yet update unit tests for the model that gives different results.

mwetter:
@Mathadon : I added a few more validation cases in `Fluid.MixingVolumes.Validation`, and discovered a bug in the steady state balance. The following needs to be fixed:
- [ ] The model `Fluid.MixingVolumes.Validation.MixingVolumeReverseFlowMoisturePrescribedHeatFlowRate` computes wrong results during reverse flow. The current reference results are correct (as I generated the correct results prior to setting `prescribedHeatFlowRate=true`), but if you run the regression tests, you see that this model produces the wrong result with `prescribedHeatFlowRate=false`. The error is in `StaticTwoPortConservationEquation`, probably caused by `m_flowInv` or `m_flowInv_b`.
- [ ] In the steady state balance, `m_flowInv` should be renamed `m_flowInv_a`.
- [ ] The graphics of some of the new validation models needs to be cleaned up to align components with the grid. Please don't resize components to make them small, and avoid that connection lines are on top of the icons. You can always increase the graphic pane if you run out of space. See for example `MoistureMixingConservationDynamicBalance`

mwetter:
@Mathadon : I corrected the bug, but there are still open issues (see above). I think we should assume `part_a.m_flow == -port_b.m_flow`, unless we can demonstrate that this approximation leads to too big an error. I added various fixme's that need to be addressed if we assume them to be different. The implementation would get rather complex, and I am not convinced that we gain anything useful from it.

Mathadon:
This is on my to do list but I'm abroad until thursday so it may take a while.

Do you suggest to keep the old implementation then or to use some in-between solution?

mwetter:
I suggest an in between solution that uses one m_flowInv. I will try to
have a look at it as the current version has a bug and I need a solution
for the Buildings release

Mathadon:
@mwetter (I think) I fixed the layout of the examples. See pull request!

The current implementation does not really conserve mass, while the unit tests suggest it does. The fact that we set `simplify_mWat_flow = true` by default causes the assert to pass, but otherwise it does not. We should probably deal with this differently? Otherwise there is no use for this pull request since it does not fix anything.

Also, the current implementation we use is
```
port_b.Xi_outflow = (inStream(port_a.Xi_outflow)*port_a.m_flow + mXi_flow) * m_flowInv;
```

if we do not use `m_fLowInv_b` then it is better to change this back to the old implementation:
```
port_b.Xi_outflow = inStream(port_a.Xi_outflow) + mXi_flow * m_flowInv;
```
since it does not have a problem for `m_flow = 0 `.

mwetter:
@Mathadon : Thanks for cleaning up the examples.

I agree that 
```
port_b.Xi_outflow = (inStream(port_a.Xi_outflow)*port_a.m_flow + mXi_flow) * m_flowInv;
```
needs to be changed, which I just did. I oversaw that. Thanks for catching this.

I disagree that this pull request does not fix anything. The new regression test `MixingVolumeTraceSubstanceReverseFlowPrescribedHeatFlowRate` shows that the old implementation has an error in computing `volSte.C[1]` if the mass flow rate is negative. This is fixed in the new version.

The info section of `StaticTwoPortConservationEquation` states 

> This causes an error in the mass balance of about <i>0.5%</i>, but generally leads to
simpler equations because the pressure drop equations are then decoupled from the
mass exchange in this component.

So, I am fine with this. I attempted an implementation that conserves mass, but it gets really complicated. Do you have a better implementation that is still efficient? As the conservation of mass is only violated for the case with moisture added, I don't see an issue in this small error. It can only occur in air flow systems, in which ducts leak anyway which is an effect we do (not yet) consider. See http://ducts.lbl.gov/calducts.htm

I merged `issue300_regStep_integration` which is in review by @marcusfuchs in https://github.com/iea-annex60/modelica-annex60/pull/434
The merge was done because we need `regStep` also in the steady state conservation equation.
Hence, this should only be merged when @marcusfuchs merged https://github.com/iea-annex60/modelica-annex60/pull/434

Regression test results have been updated.

Mathadon:
Oh yes, I overlooked the error for the `C` variables. So that is fixed indeed.

The only concern I have is then that the 'validation model' 'proves' that we conserve mass exactly, while this is not the case. So the documentation should maybe specify this?

I'm ok with this implementation, no need for the complicated one (I do not have a better solution).

mwetter:
I fixed the documentation in a7b6b5d

PR #449, opened 2016-04-11
Issue433 lim pid

### LINKED ISSUE #433: Annex60.Controls.Continuous.LimPID triggers events when it saturates
`Annex60.Controls.Continuous.LimPID` triggers events when the output limiter becomes active.
This considerably slows down the simulation of `Buildings.Examples.DualFanDualDuct`, and probably of other models as well. These events can be avoided by setting `strict=true`.

The code that is affected by this change is in the limiter of the MSL, 
```
  if initial() and not limitsAtInit then
     y = u;
     assert(u >= uMin - 0.01*abs(uMin) and 
            u <= uMax + 0.01*abs(uMax),
           "Limiter: During initialization the limits have been ignored.\n"+
           "However, the result is that the input u is not within the required limits:\n"+
           "  u = " + String(u) + ", uMin = " + String(uMin) + ", uMax = " + String(uMax));
  elseif strict then
     y = smooth(0, noEvent(if u > uMax then uMax else if u < uMin then uMin else u));
  else
     y = smooth(0,if u > uMax then uMax else if u < uMin then uMin else u);
  end if;
```
This issue is to see if we should set `strict=true` as the default value.

Below is a chart that compares the number of events and the CPU time. Run 2 is with `strict=false` while run 3 is with `strict=true`.
closedloop

mwetter:
This closes #433. Only small differences have been seen in the regression tests.


mwetter:
@mlauster I delete this pull request as there is already another development branch `issue433_strict` that should be used.

PR #453, opened 2016-04-12
Issue435 radiator

### LINKED ISSUE #435: Reduce nonlinear system of equations in radiator model
In Dymola 2016 FD01, the nonlinear system of equation for the radiator model `RadiatorEN442_2` increased compared to Dymola 2016. An isssue has been filed with dassault. This ticket is to implement a work-around received from Dassault.

mwetter:
This will close #435.
The change is backward compatible.
All unit tests work.


marcusfuchs:
Interesting effects here. Looks good!

PR #456, opened 2016-04-12
Issue433 strict

### LINKED ISSUE #433: Annex60.Controls.Continuous.LimPID triggers events when it saturates

mwetter:
This closes #433 

All regression tests work.


Annex60/Controls/Continuous/PIDHysteresis.mo
```diff
@@ -52,28 +52,32 @@ model PIDHysteresis
...
   Modelica.Blocks.Logical.Hysteresis hys(
-    pre_y_start=pre_y_start,
-    uLow=eOff,
-    uHigh=eOn) "Hysteresis element to switch controller on and off"
+    final pre_y_start=pre_y_start,
+    final uLow=eOff,
+    final uHigh=eOn) "Hysteresis element to switch controller on and off"
     annotation (Placement(transformation(extent={{-30,50},{-10,70}})));
   Modelica.Blocks.Logical.Switch swi
     annotation (Placement(transformation(extent={{60,-10},{80,10}})));
```
mlauster: In PIDHysteresisTimer, you set this to `zer(final k=0)`, @mwetter . Is there a reason why it is different for this model here?

mwetter:
@mlauster : there is no reason, for consistency, it should be the same in both models. I changed this in https://github.com/iea-annex60/modelica-annex60/pull/456/commits/fbb38728b5b663741d9b70ec7d298e4f1c487e31, in which I also made some block protected. This way, the same blocks are protected in `PIDHysteresis` and in `PIDHysteresisTimer`. This change is however not backwards compatible.

PR #466, opened 2016-04-25
Issue464 jmodelica

mwetter:
This pull request adds more fixes for JModelica.


Mathadon:
You relaxed some error computations. Does this mean that JModelica does/did not satisfy the error bound and dymola does?

mwetter:
@Mathadon: I simply changed the test so it checks for relative error rather than absolute error. Previously, we checked `assert(abs(x-y) < 1E-2, "Model has an error.");` but since `x` is in the order of 1E6, this would require an error below 1E-8, which is asking for very high precision, and won't be satisfied with the default solver tolerance.

PR #473, opened 2016-05-21
Issue471 velocity sensor

### LINKED ISSUE #471: Add sensor for flow velocity?
For some of our applications, it would be useful to show the flow velocity of the medium. Is there interest in having a flow velocity sensor in the library?

marcusfuchs:
This closes #471.


Annex60/Fluid/Sensors/Examples/FlowVelocity.mo
```diff
@@ -0,0 +1,78 @@
+within Annex60.Fluid.Sensors.Examples;
+model FlowVelocity "Test model for the flow velocity sensor"
+  import Annex60;
```
mwetter: Don't use `import Annex60;` as there is no need to import everything.

Annex60/Fluid/Sensors/Examples/FlowVelocity.mo
```diff
@@ -0,0 +1,78 @@
...
+  extends Modelica.Icons.Example;
+
+  package Medium = Annex60.Media.Air "Medium model";
+
+  parameter Modelica.SIunits.Area crossSection = 0.1*0.1/4*Modelica.Constants.pi
```
mwetter: I suggest to rename 'crossSection` to 'A`. Then it is consistent with the `Dampers.Exponential` and with the `Orifice` which both use `A` for the face area.

Annex60/Fluid/Sensors/FlowVelocity.mo
```diff
@@ -0,0 +1,122 @@
+within Annex60.Fluid.Sensors;
+model FlowVelocity "Ideal sensor for flow velocity"
```
mwetter: I suggest to rename it the model to `Velocity` which is shorter and in my view, it is clear that it is the velocity of the flow as the sensor is in the `Fluid` package.

mwetter:
@marcusfuchs 
Can you please address the comments above.

marcusfuchs:
@mwetter 
I addressed all comments and also fixed the `defaultComponentName` annotation.

mwetter:
@marcusfuchs 

I removed 
```
Annex60/Experimental/Pipe/Examples/UseCases/TypeS_Development/UCPipeS02AD_DpM.mo
Annex60/Experimental/Pipe/Examples/UseCases/TypeS_Development/UCPipeS02AD_MSL_DpM.mo
```
which do not belong in this pull request.

I also changed the name of the sensor output signal from `velocity` to `v` to use a shorter name (easier to display in plots) and to be consistent with the following models:
* `DoorDiscretized` which uses `v`
* `Damper` which uses `v_nominal` and `use_v_nominal`
* `PowerLawResistance` which uses `v`

I also added the reference result.

Please let me know if you agree with the changes, in which case it can be merged.

marcusfuchs:
@mwetter 
I agree with the changes and sorry for the pipe examples, you are right, they should not be in this PR.

Issue #475, opened 2016-05-24
Duplicate model in Psychometrics?


Mathadon:
It seems as if `Annex60.Utilities.Psychrometrics.WetBul_pTX` and `Annex60.Utilities.Psychrometrics.TWetBul_TDryBulXi` are more or less identical models. Can we remove either of the two? I suggest the former.

mwetter:
From my point of view, `Annex60.Utilities.Psychrometrics.WetBul_pTX` can be deleted as all computations can also be done using `Annex60.Utilities.Psychrometrics.TWetBul_TDryBulXi`.
The only capability that `Annex60.Utilities.Psychrometrics.TWetBul_TDryBulXi` adds is to output `X` at the wet bulb temperature, which can also be computed using `X_pTphi`.

@thorade , @mlauster : Is UdK Berlin and Aachen OK with removing `Annex60.Utilities.Psychrometrics.WetBul_pTX`? If not, then the fix of #474 also needs to be applied to this model.

mwetter:
@thorade , @mlauster 
Can you please comment on this so we can process the issue.

mlauster:
Removing should be fine for Aachen.

thorade:
Removing one of the two is fine for UdK.

PR #476, opened 2016-05-24
Issue474 wetbulbtemperature

### LINKED ISSUE #474: Enthalpy of evaporation of water
In `Annex60.Utilities.Psychrometrics.Constants` we currently have
```
constant Modelica.SIunits.SpecificEnthalpy h_fg = 2501014.5
    "Enthalpy of evaporator of water";
```
which seems to be inconsistent with the often quoted enthalpy of evaporation of 2260 kJ/kg. Is this correct?

Edit:
h_fg = 2501 kJ/kg is the enthalpy of evaporation at 100 degrees C. The specific enthalpy of saturated steam is 2257 kJ/kg. This is based on http://www.engineeringtoolbox.com/steam-vapor-enthalpy-d_160.html . So I think the value of h_fg is correct?
However, the example below then still gives different results. Is there possibly an error in the implementation of Annex60.Utilities.Psychrometrics.TWetBul_TDryBulPhi?

Mathadon:
This fixes #474 
note that I did not fix the code in `Annex60.Utilities.Psychrometrics.WetBul_pTX` since this should be removed in #475


mwetter:
@Mathadon 

I changed
```
assert(abs(wetBulPhi_TWetBul-wetBulXi_TWetBul)<1e-6, "Inconsistent implementation of wetBulPhi and wetBulXi.");
```
to
```
assert(abs(wetBulPhi_TWetBul-wetBulXi_TWetBul)<1e-2, "Inconsistent implementation of wetBulPhi and wetBulXi.");
```
as the test is at higher accuracy than the solver tolerance.

I also moved the assertions into a separate block to avoid mixing graphical modeling with textual modeling in the equation section (which is part of the style guide).

I cleaned up the model a bit.

When running the regression tests, I got a different result for `Annex60_Fluid_Sensors_Examples_TemperatureWetBulb.txt`.
Can you please verify and check in the new result.

If the tests run and you are fine with the above changes, I can merge it to the master.

Mathadon:
I agree with the changes!

I allowed a change by dymola in 		cab5d0f since an annotation seemed to be in the wrong place.

The change in reference result was because I only ran unit tests for Annex60.Utilities.Psychrometrics. I now also ran Annex60.Utilities.Sensors and I (also) get different results. I accepted the changes and pushed them in 1d50df337544242c3340559dbb4db004f75f89ae.

mwetter:
@Mathadon 
Actually, the misplaced annotation can be removed altogether, which I did in 76e7021.
This commit also makes the new block protected as there is no need to have it public.

PR #486, opened 2016-06-26
Issue485 active beam

### LINKED ISSUE #485: add active beam model
Add the active beam model from the `Buildings` library that was developed by @amaccarini  through
https://github.com/lbl-srg/modelica-buildings/issues/499

See also https://bitbucket.org/berkeleylab/iea-annex-60-admin/wiki/Meetings/2016-05-expertMeeting/activity_1.1_breakout

mwetter:
@mlauster Who at Aachen would be interested in reviewing this pull request, which would close #485. See also the notes from Miami at https://bitbucket.org/berkeleylab/iea-annex-60-admin/wiki/Meetings/2016-05-expertMeeting/activity_1.1_breakout


mlauster:
I asked Marcus Schumacher, but he is on holiday, I hopefully get answer within the next weeks if he can do that.

mwetter:
@mlauster : Can Marcus Schumacher review this, or shall we assign it to someone else?

mlauster:
Sorry, this got out of my attention. He didn't seem really interested in reviewing, so we should choose somebody else.

mwetter:
Is there someone from Aachen, Leuven or Berlin (or other Annex 60 member) who can review this code? It was developed by @amaccarini and reviewed by me for the `Buildings` library.

Mathadon:
We had a discussion in Miami whether or not the active beam model should supply heat through a heat port or by explicitly modelling the air that's drawn from the zone and by adding the heat into this air stream. I'd like to pick this up again.

With the current implementation `air_b.h_outflow` is not correct. If the user would connect a temperature sensor to it, the wrong temperature would be outputted. I would therefore be in favour of replacing the heat port connection by a fluid connection, or to provide both options.

amaccarini:
True, air_b.h_outflow is not correct when thinking about the real operation
of active beam units. In a previous work I modeled a specific active beam
unit by explicitly modeling the induced air from the room. The water
stream supplied heat to the induced air form the room in a heat exchanger.
Then, induced (and now heated) air and primary air from the plenum mixed
together and the resulting air stream was supplied to the room. Therefore,
in this model, the heat to the room was supplied entirely by the air.
Please see http://link.springer.com/article/10.1007/s12273-015-0236-5

The choice of using a different approach for the newly developed active
beam model reflects the modeling strategy used in E+. Recently, E+
introduced a new model for active beam systems where heat was supplied to
the room separately (air and water). In addition, if we want to model the
correct value of air_b.h_outflow, we need to know the induction ratio at
the current operating conditions. This can be a non-trivial task since
manufacturers usually do not provide empirical equations for the induction
ratio.

Unfortunately I was not in Miami, but we can further discuss this topic
next month in Corsica.

2016-09-06 10:55 GMT+02:00 Filip <notifications@github.com>:

mwetter:
I would also prefer to keep the current design, e.g., not have a separate air path for the induced air. The rationale is as follows (copied from the UserGuide):

> All heat flow rate that is added to or extracted from the room is transmitted through the heat port heaPor. Hence, this model does not cool the supply air between the ports air_a and air_b. Rather, it adds this heat flow rate to the heat port heaPor. The rationale for this implementation is that the beam transfers heat by convection directly to the room, and by induction of room air into the supply air. As this split of heat flow rate is generally not known, and because the amount of inducted air is also unknown, it was decided to transfer all heat through the heat port heaPor. This also avoids having to add an extra air flow path for the air induced from the room. 

If there are concerns with this approach, we can revisit, but then we would need to know the induction ratio which to my knowledge is not published by manufacturers. Getting the induction ratio wrong may also cause the air to be in the fog region (if we underestimate the amount of induction during cooling mode) which may cause issues if users use a medium that is different from `Annex60.Air.*`.

Mathadon:
okay!

Annex60/Fluid/HeatExchangers/ActiveBeams/Data/BaseClasses/TemperatureDifference.mo
```diff
@@ -0,0 +1,49 @@
+within Annex60.Fluid.HeatExchangers.ActiveBeams.Data.BaseClasses;
+record TemperatureDifference "Record for temperature difference"
+  extends Modelica.Icons.Record;
+  parameter Real r_dT[:](each min=0, each final unit="1")
+   "Normalized temperature difference, e.g., temperature difference at user-selected operating points divided by nominal temperature difference. Must be positive.";
```
mlauster: This line could be parted into two for better versioning control

Annex60/Fluid/HeatExchangers/ActiveBeams/UsersGuide.mo
```diff
@@ -0,0 +1,155 @@
+within Annex60.Fluid.HeatExchangers.ActiveBeams;
+package UsersGuide "User's Guide"
+  extends Modelica.Icons.Information;
+  annotation (preferredView="info",
+  Documentation(info="<html>
+<p>
```
mlauster: Have you already run html checks?

Annex60/Fluid/HeatExchangers/ActiveBeams/Validation/NumberOfBeams.mo
```diff
@@ -0,0 +1,215 @@
...
+      Annex60.Fluid.HeatExchangers.ActiveBeams.Data.Trox.DID632A_nozzleH_lenght6ft_heating
+      perHea,
+    nBeams=nBeams,
+    energyDynamics=Modelica.Fluid.Types.Dynamics.SteadyState)
+                   "Active beam"
+    annotation (Placement(transformation(extent={{-14,-92},{14,-68}})));
+  Modelica.Blocks.Sources.Step step(
+    height=-0.094,
+    offset=0.094,
+    startTime=2000) "chilled water mass flow rate"
```
mlauster: Use capital letter for first character in comment

mwetter:
@mlauster : I addressed all of the above points and rerun the unit tests, which also includes the html check.

PR #490, opened 2016-07-07
Issue487 water documentation

carlesRT:
Text was changed but  I did not add plots... I could do it if there is some template or hint about the configuration so all plots looks the same.


mwetter:
@carlesRT 
It would be nice if you can add the plots. The existing ones were generated using
`Annex60/Resources/Images/Media/Water/{plotDensity.py, plotCp.py}`

carlesRT:
@mwetter I added the plots. I think it is ready to merge.

mwetter:
@carlesRT 
Thanks. I made some revision to the documentation and will merge it to the master.

PR #491, opened 2016-07-07
Issue482 rom validation

### LINKED ISSUE #482: add validation criteria and missing source for figures for ROM VDI 6007
@mlauster 

- [x] Can you please add the source file that was used to generate the figures in `Resources/Images/ThermalZones/ReducedOrder/RC`. If they are from a full paper that you may not want to share as LaTeX, it would be good to add at least these sections that generate the figures
- [x] Also, on Dymola 2017, I get different reference results for`Resources/ReferenceResults/Dymola/Buildings_ThermalZones_ReducedOrder_Validation_VDI6007_TestCase12.txt`
Maybe this is due to the change in numerics between Dymola versions, and the interpolation done by the regression tests. In either case, it would be good to add an `assert` if VDI prescribes a certain accuracy. Otherwise, it will be very cumbersome to ensure that once validated test cases remain within the required tolerance from one version of a tool to another, and also among different tools.

mlauster:
closes #482


Annex60/ThermalZones/ReducedOrder/Validation/VDI6007/BaseClasses/package.mo
```diff
@@ -0,0 +1,5 @@
+within Annex60.ThermalZones.ReducedOrder.Validation.VDI6007;
+package BaseClasses "Base classes for VDI 6007 validation cases"
+  extends Modelica.Icons.BasesPackage;
+
```
mwetter: Documentation is missing

mwetter:
@mlauster 
Can you please verify and update the reference results. I did some changes (see revision log), mainly with regard to documentation and the new validation assertion, but I do get one more state variable in most test cases, and also different results, such as shown below.
Please review and update before I merge.

testcase2

mlauster:
@mwetter , thanks for polishing the models. The new state events are related to the `mean` block that contains a derivative, so this is okay. The different results are also related to the `mean`, as I compared degree Celsius (display unit) in the past and now need to compare Kelvin (internal unit) using `mean` (the values have no unit after that block, so I can't switch to degC). For that, `reference.y` is now in Kelvin.
Could you update the reference results? I couldn't get the regression tests running so far. I'm fine with merging afterwards.

mwetter:
@mlauster I updated the reference results. From my view, you can go ahead and merge it.

Issue #489, opened 2016-07-06
Refactor to export weather resource file as FMU


mwetter:
This issue should refactor the weather data reader to allow the export of models which contain weather file as FMUs.

Todo:
- [x] Merge this to the master branch.

mwetter:
@tsnouidui: I created branch `issue489_weatherFMU`

mwetter:
@tsnouidui Please correct `BoundaryConditions/WeatherData/BaseClasses/getAbsolutePath.mo` which differs between this development branch and the `master` of `Buildings`

tsnouidui:
@mwetter I made a local diff between this branch and the master of Buildings and got

< within Annex60.BoundaryConditions.WeatherData.BaseClasses;
---
> within Buildings.BoundaryConditions.WeatherData.BaseClasses;
40c40
< Now the function works from any directory as long as the <code>Annex60</code> library
---
> Now the function works from any directory as long as the <code>Buildings</code> library

Do you get/expect something different?

mwetter:
Sorry, my local copy used the old cached state of the branch. The files are identical (except for the top-package name).

mwetter:
This is now on the `master`.

PR #492, opened 2016-07-08
Issue489 weather fmu

tsnouidui:
This addresses #489.


tsnouidui:
@mwetter Line length has been reduced.

mwetter:
This does not yet pass the `Buildings` regression tests and hence should not be merged: https://github.com/lbl-srg/modelica-buildings/pull/540#issuecomment-236203641

mwetter:
@marcusfuchs : Can you please review and merge? This now passes the tests also for the `Buildings` library and is integrated in `Buildings:master`.

marcusfuchs:
@mwetter : Looks good and can be merged from my point of view. I just changed one instance of "uri" in a description string to "URI", as all other references in the documentation and descriptions also used "URI". In addition, my text editor removed a few trailing whitespaces. If that is fine with you, either you or I could merge this, otherwise I could also revert the whitespace changes.

mwetter:
@marcusfuchs : Thanks for the review and changes, I will merge it.

Issue #494, opened 2016-07-18
LimPID with integrator reset


PMehrfeld:
I want to enhance `Annex60.Controls.Continuous.LimPID`. When a boolean input becomes true, the integrator should be reset.
The problem is easy to solve as long as you are not in pedantic mode.
But to satisfy pedantic criteria I can only see the solution to copy the whole source code of `Modelica.Blocks.Continuous.LimPID` instead of extending it. Because then the integrator can be replaced by an own integrator model including a boolean input to reset/reinit() the output `y.`

I will show the two possibilities in a new fork.

PMehrfeld:
It would be nice to get some feedback about the two options you can see in my fork: [issue494_resetIntPID](https://github.com/PMehrfeld/modelica-annex60/tree/issue494_resetIntPID)

So, what do you think is the best?
Option a) Leave LimPID as extension from standard library, but check in pedantic mode leads to erros.
Option b) Duplicate LimPID form standard library and create own integrator.
Option c) Neither a) nor b) and let everyone model this by himself/herself if someone need a reset of the integrator.
Option d) A better modelling approach than a) that will also meet the requirements of pedantic mode.

It would be nice to receive some feedback.

Mathadon:
Hi!
I did not check all of your code but I think b) would be best as long as the additional inputs / complexity can be disabled by default such that the default implementation stays the same. 

The question arises however if we should do 

```
reinit(y,y_inint)
```

or 

```
reinit(y, something else)
```

c) is an option as well since we do not want to overload the user with complex options. Good documentation should however be sufficient to deal with this.

PMehrfeld:
Thanks a lot for your recommendations!

I now decided to choose Option b) and use the edge() function. Also you can choose the value to reinit with.
I hope the documentation is detailled enough.

PR #495, opened 2016-07-18
Issue494 optional integrator reset for PID

PMehrfeld:
#494


Mathadon:
We'll have to check this with @mwetter too, so let us get back to you!

PMehrfeld:
Alright. Thanks a lot!

mwetter:
@PMehrfeld 

I have a few questions/comments to your implementation.

1. Is the integrator anti-windup as implemented in the original controller not sufficient? If you indeed have a case for having to use a reset, then we can certainly add this, but I am asking as I have not yet encountered this need and hence want to avoid adding complexity if there is no need for such a feature.
1. I would rename `resetI` to `reset` as the integrator is anyway the only thing that can be reset, so no need to use `I` in the name.
1. `IntegratorWithReset` should be moved to `Utilities.Math` as it may be useful for other applications, and it is not a controller by itself (and hence should not be in `Controls.Continuous`.
1. Is it really needed to have `intResetValue`? Why would someone use a different value than `xi_start`?
1. To use consistency with other parameter names, please rename `withResetIntegrator` to `use_reset`.
1. The documentation talks about rising edge, but the implementation uses `edge`.
1. The code `when edge(resetIPassThrough.y) then` should be in an `if use_reset` and `use_reset` would need `annoation(Evaluate=true)` so that (hopefully) tools will remove this event if not needed.
1. Examples or validation tests need to be added before it can be merged to the master (we need test for all features to ensure that we don't break what has been tested already).

PMehrfeld:
@mwetter 
Thank you very much for your feedback.

1. The anti-windup feature only drives down the integrator when the boundary limits are reached/exceeded. Even then there is no hard reset to `y_start` (or any other value). Please correct me, if have overlooked something.
Regarding an exemplaric case: For example, a plant/system should be shut down when a certain trigger/boolean becomes true, and when it restarts the integration must start from zero again.
Anyway, the model `IntegratorWithReset` can be very useful in diverse applications.

2. OK -> Done

3. Absolutely!! -> Done

4. This was Filip's suggestion (see #494 ). It might be useful for some people.

5. OK -> Done

6. To my knowledge the `edge()` function means a rising edge. A test proved this. Additionally, a closer look to the block `Modelica.Blocks.MathBoolean.RisingEdge` says: 
```
block RisingEdge
  "Output y is true, if the input u has a rising edge, otherwise it is false (y = edge(u))"
[...]
equation 
  y = edge(u);
end RisingEdge;
```

7. OK -> Done

8. OK -> Done with `Controls.Continuous.Examples.LimPIDWithReset` and `Utilities.Math.Examples.IntegratorWithReset`.
The remaining applications where `LimPID` is used should still work, as default parameter values disable the integrator reset.

mwetter:
@PMehrfeld Thanks for the changes. I will review but it may take a bit longer due to travel this and next week. I merged your changes to https://github.com/iea-annex60/modelica-annex60/tree/issue494_resetIntPID and will commit my revision to this branch.

PMehrfeld:
@mwetter I have comitted into my fork the discussed changes:
1. `LimPID` output is set to 0 if boolean input `reset `has a rising edge.
2. I implemented the Annex60 types `Annex60.BoundaryConditions.Types.DataSource.Input` and `.Parameter` in the model `IntegratorWithReset`. Additionally, I use the `_internal ` variables.

I also revised the 2 examples and added scripts to `Annex60\Resources\Scripts\Dymola\...`.
Unfortunately, I could not manage to get the python script `runUnitTests.py` work properly.

Can I do any further work or does this satisfy your requirements?

PR #499, opened 2016-07-29
added custom implementation of extractor in PartialFlowMachine

### LINKED ISSUE #498: Bug in Extractor in Movers when using Euler
This issue is to fix https://github.com/lbl-srg/modelica-buildings/issues/472.

There seems to be an issue when multiple events occur within 1 fixed time step interval. The extractor implementation in the pump is then not correctly converting the input of the block into its output. See ticket above for a figure. 
I've been having issues with this time and again without being able to create a clear working (bugged) example so I hope to fix this without adding a unit test for the bug.

Mathadon:
This closes #498 
I ran unit tests in Annex60.Fluid.Movers, which yielded the same results.


mwetter:
@Mathadon 
I made `Extractor` protected so that it can eventually be removed.
I will also run all the other regression tests as we need to run them all prior to merging on the master.

PR #502, opened 2016-08-03
Issue501 week day

### LINKED ISSUE #501: Compute weekdays
I propose to add the functionality to compute weekdays. This requires a link to be made between the Modelica variable `time` and our calendar system. Therefore there must be a way for the user to specify what it means if time = 0. For instance time = 0 may mean that it's either 
1) new year 2014,
2) new year 2015,
etc..
3) new year 1970 (corresponding to unix time stamp = 0)
4) some custom date

Since the concept of time is linked to the input file I propose to add it in `Annex60.BoundaryConditions.WeatherData.ReaderTMY3`. This way it can also be integrated easily into IDEAS.

I'll propose an implementation that is disabled by default.

This relates to https://github.com/open-ideas/IDEAS/issues/534

Mathadon:
This closes #501 
Added one unit test and result.

I did not make changes to other models. I.e. this model is currently not integrated into other models.


Annex60/Utilities/Time/CalendarTime.mo
```diff
@@ -0,0 +1,279 @@
...
+    annotation (Placement(transformation(extent={{100,-10},{120,10}})));
+  discrete Modelica.Blocks.Interfaces.IntegerOutput month "Month of the year"
+    annotation (Placement(transformation(extent={{100,10},{120,30}})));
+  Modelica.Blocks.Interfaces.IntegerOutput day "Day of the month"
+    annotation (Placement(transformation(extent={{100,30},{120,50}})));
+  Modelica.Blocks.Interfaces.IntegerOutput hour "Hour of the day"
+    annotation (Placement(transformation(extent={{100,50},{120,70}})));
+  Modelica.Blocks.Interfaces.RealOutput minute "Minute of the hour"
+    annotation (Placement(transformation(extent={{100,70},{120,90}})));
+  // fixme: why is there not a second of the minute?
```
Mathadon: I think this calculation would generate too many events, hence the 'minute' is a Real instead of an integer number.

Annex60/Utilities/Time/CalendarTime.mo
```diff
@@ -0,0 +1,279 @@
...
+      epochLastMonth :=epochLastMonth + (if i == 2 and isLeapYear[yearIndex]
+         then 1 + dayInMonth[i] else dayInMonth[i])*3600*24;
+    end if;
+  end for;
+
+equation
+  // compute unix time step based on found offset
+  unixTimeStamp = tim + timOff;
+
+  // fixme: Both "when" constructs need to be rewritten.
```
Mathadon: I was unaware of this.. Thanks!

Annex60/Utilities/Time/CalendarTime.mo
```diff
@@ -0,0 +1,279 @@
...
+      (if pre(month)==2 and isLeapYear[yearIndex]
+        then 1 + dayInMonth[pre(month)] else dayInMonth[pre(month)])*3600*24;
+  end when;
+
+  // compute other variables that can be computed without using when() statements
+  daysSinceEpoch = integer(floor(unixTimeStamp/3600/24));
+  // fixme: I see why you want to use Real for minutes and I think this is a good design.
+  // However, I would make weekDay, day and hour also Reals. Then, we have one design principle
+  // (e.g., day, hour, minute are all Real and users need to convert to integer if they need an integer)
+  // rather than having a special case for minute
```
Mathadon: Do you suggest to just change the variable to Real or also remove the floor() functions etc that ensure that the actual output is a round number?

Annex60/Utilities/Time/CalendarTime.mo
Mathadon: This does not seem to work since 'time' passes through the input variable 'tim'. The fact that `tim` is a linear function of `time` does not seem to be detected.
An option would be to use 'time' instead, but this would only avoid 76 state events for 1e8 seconds of simulation time.
55000 more events are generated by the `floor()` and `rem()` functions etc, but they do not seem to be able to generate time events at all. 
Even this model generates state events:
```
model test
  Real test;
equation
  test=floor(time);
end test;
```

Annex60/Utilities/Time/Examples/CalendarTime.mo
```diff
@@ -0,0 +1,52 @@
+within Annex60.Utilities.Time.Examples;
+model CalendarTime
+  extends Modelica.Icons.Example;
+  // fixme: comment of calendarTime2016 should probably be changed
```
Mathadon: I am not sure what you mean by this comment?

mwetter: @Mathadon The comment is currently

> ... assuming time=0 corresponds to new year 2016"

This can be changed because if we set `final Real monthRef=1` etc., then `time=0` is always a new year, or maybe you simply wanted to stress the year of 2016?

Mathadon: Yes, I think so :p Since you have the option to choose from multiple years.

PR #517, opened 2016-09-19
Issue516 dymola2017 fd01

### LINKED ISSUE #516: verify that library works for Dymola 2017 FD01 beta2
This issue is to correct models as needed for Dymola 2017 FD01 beta2.

mwetter:
This will close #516 

All models work in Dymola 2017 FD01 beta2 with the pedantic model check.


marcusfuchs:
Can't check with the beta, but changes look good to me.

PR #547, opened 2016-10-14
Added missing jar file

### LINKED ISSUE #533: invalid path in ReaderTMY3 documentation
The documentation of `BoundaryConditions.WeatherData.ReaderTMY3` contains a command to convert `.epw`-files to `.mos`-weatherfiles. The associated java archive is not part of Annex60 yet (it is part of Buildings), thus the command does not work. Should we integrate the archive into Annex60 or modify the documentation? It is not possible to link only to Buildings, the merge-script from BuildingsPy will change this path when merging into another library.

mwetter:
This closes #533.

All that was missing was the jar file in the `Resources/bin` directory.


mlauster:
Thanks for that!

PR #548, opened 2016-10-17
Issue451 splitter fixed resistance dp m

### LINKED ISSUE #451: Fluid.FixedResistances.Examples.SplitterFixedResistanceDpM used in doc but is not in library
The package information of `Annex60.Fluid.FixedResistances` has a link to `Annex60.Fluid.FixedResistances.Examples.SplitterFixedResistanceDpM`, but this model is only in `Buildings`, not in `Annex60`.

The model should be either added to `Annex60` or removed from the documentation. As we use this model very often in `Buildings`, I suggest to add it. It is similar to the 3 way valves, except that it has in each flow leg a `FixedResistanceDpM` model.

@mlauster, @Mathadon, @thorade do you (speaking for your libraries) think this should be added, needs more discussion, or are you against adding it?

mwetter:
This closes #451


Mathadon:
Thanks for adding the model!

I have some more suggestions:
 - In `Annex60.Fluid.FixedResistances.SplitterFixedResistanceDpM` I propose to put 'transition to laminar' after 'nominal condition' in the parameters box, or even move it to and 'Advanced' tab since these parameters do not significantly affect results. This relates to https://github.com/iea-annex60/modelica-annex60/issues/480.
 - In the revision history you wrote 2017 instead of 2016.

I'd like to use this model in the exercise so feel free to merge it once you corrected the revision history! We can discuss the first bullet this weekend.

mwetter:
I reordered the parameters and corrected the revision date.
I did not move `deltaM` to the advanced tab, as it depends on the value of `use_dh` which is not on the advanced tab.

PR #555, opened 2016-10-23
Added assertion for check of energy balance

### LINKED ISSUE #470: Asserts for Q_flow in MixingVolume when m_flow=0
We should add an assert that verifies if energy is conserved for zero flow if the mixing volume is configured to be steady-state energy balance.
This may detect certain singularities instead of throwing an error in the algebraic loop solver, which may result in the problems discussed in #445 . 

The assert should check if the heat flow rate is such that the corresponding temperature increase stays below a certain bound.

mwetter:
This will close #470


Mathadon:
`Annex60.Fluid.HeatExchangers.Validation.ConstantEffectiveness` fails with 

```Model error - division by zero: (hex.Q1_flow/hex.bal1.vol.cp_default) / (sin_1.ports[1].m_flow) = (-0) / (0)```

This should error not occur since the implementation is

```
    assert(if abs(m_flow) < 10*Modelica.Constants.small
      then 
        abs(Q_flow) < 1E-10
      else 
        abs(Q_flow/cp_default/m_flow) < 200,
   "Energy may not be conserved for small mass flow rates. This model may require prescribedHeatFlowRate = false.");
```
adding `noEvent()` does not work.

dsmodel.c contains the following code:

```
AssertModelica(IF Less(fabs(DYNX(W_,24)),"abs(sin_2.ports[1].m_flow)", 1E-59,
  "1E-59", 7) THEN Less(fabs(DYNX(W_,192)),"abs(hex.Q1_flow)", 1E-10,"1E-10", 5)
   ELSE Less(fabs(divmacro(divmacro(DYNX(W_,192),"hex.Q1_flow",DYNX(W_,308),
  "hex.bal2.vol.cp_default"),"hex.Q1_flow/hex.bal2.vol.cp_default",DYNX(W_,24),
  "sin_2.ports[1].m_flow")),"abs(hex.Q1_flow/hex.bal2.vol.cp_default/sin_2.ports[1].m_flow)",
   200,"200", 8),"(if abs(sin_2.ports[1].m_flow) < 1E-59 then abs(hex.Q1_flow) < 1E-10 else abs(hex.Q1_flow/hex.bal2.vol.cp_default/sin_2.ports[1].m_flow) < 200)",
   "Energy may not be conserved for small mass flow rates. This model may require prescribedHeatFlowRate = false.");
```

it seems that even though the division by zero occurs in the 'else' branch, it is still evaluated, while this need not happen. 
For m_flowInv in `inverseXRegularized` we use the same kind of construct to guard against division by zero. This function generates the following C code (in the same model):

```
DYNX(W_,222) = IF DYNX(W_,106) > 5E-07 OR DYNX(W_,106) < -5E-07 THEN divmacro(1,
  "1",DYNX(W_,106),"sin_1.ports[1].m_flow") ELSE IF DYNX(W_,106) < 2.5E-07 AND 
  DYNX(W_,106) > -2.5E-07 THEN 4000000000000.0*DYNX(W_,106) ELSE 
  (PushModelContext(1,"Annex60.Utilities.Math.Functions.BaseClasses.smoothTransition(sin_1.ports[1].m_flow, 5E-07, 2000000.0, -30000000.0, 476000000000000.0, -2.888E+21, 8.544E+27, -1.216E+34, 6.656E+39)")
  Annex60_Utilities_Math_Functions_BaseClasses_smoothTransition(DYNX(W_,106), 
  5E-07, 2000000.0, -30000000.0, 476000000000000.0, -2.888E+21, 8.544E+27, 
  -1.216E+34, 6.656E+39));
```

This looks more or less the same so I have no idea why this code does not work..

**Comment by mwetter** [comment]
@Mathadon You will need `noEvent` to avoid both branches to be evaluated during event iteration.

**Comment by Mathadon** [comment]
`noEvent` did not work when I tested it. Looks like a bug then?

**Comment by mwetter** [comment]
@Mathadon : With `noEvent`, the branch that seems to trigger a division by zero in your test case should not be executed. So, this looks like a bug.

**Comment by Mathadon** [comment]
I tried to reproduce the issue but I was unable to. Adding noEvent now works. I think I did something wrong last time.

**Comment by Mathadon** [comment]
from the unit tests I get outputs like

```
*** Warning: Reference file Annex60_Fluid_FMI_ExportContainers_Validation_FMUs_ThermalZoneSimpleAir3.txt has different FMU statistics for 'InitialUnknowns'.
*** Warning: Reference file Annex60_Fluid_FMI_ExportContainers_Validation_FMUs_ThermalZoneSimpleAir3.txt has different FMU statistics for 'Outputs'.
             Rewrite file?
             Enter: y(yes), n(no), Y(yes for all), N(no for all):
```
but I think this is unrelated? This may be because my buildingspy may be outdated?

Other than that this is ready to be merged.

Issue #558, opened 2016-10-26
Bug in valve models


Mathadon:
The current valve implementations do not guard against negative control signals. When the control signal becomes negative, the k value becomes negative, which causes the valve to behave like a pump.

This should be guarded against using an `assert()`.

mwetter:
This is on the master.

PR #559, opened 2016-10-27
added assert for phi>0 in valve partial

Mathadon:
this is for #558


Annex60/Fluid/Actuators/BaseClasses/PartialTwoWayValve.mo
```diff
@@ -102,6 +108,14 @@ each valve opening characteristics has different parameters.
 revisions="<html>
 <ul>
 <li>
+October 27, 2016, by Filip Jorissen:<br/>
+Added assert for <code>phi > 0</code>.
```
mwetter: Don't use `>`, see http://www.w3schools.com/html/html_entities.asp

Annex60/Fluid/Actuators/BaseClasses/PartialTwoWayValve.mo
```diff
@@ -29,7 +32,10 @@ protected
 initial equation
   assert(dpFixed_nominal > -Modelica.Constants.eps, "Require dpFixed_nominal >= 0. Received dpFixed_nominal = "
         + String(dpFixed_nominal) + " Pa.");
-
+equation
+  if assertPhiPos then
+    assert(phi>0, "Negative valve control signals are not allowed, but found phi=" + String(phi));
```
mwetter: Zero, which is not negative, is also not allowed. Hence, a better comment is "Valve control signal needs to be bigger than zero, received "

Annex60/Fluid/Actuators/BaseClasses/PartialTwoWayValve.mo
```diff
@@ -19,6 +19,9 @@ partial model PartialTwoWayValve "Partial model for a two way valve"
   input Real phi
     "Ratio actual to nominal mass flow rate of valve, phi=Kv(y)/Kv(y=1)";
 protected
+ constant Boolean assertPhiPos = true
```
mwetter: I suggest to remove this as its value can only be changed by models that extend this model, but not by a user because it is protected.
Also, `Evaluate=true` is not needed for `constant` as constants are evaluated at compile time.

Mathadon:
Fixed suggestions!

PR #574, opened 2016-11-04
Issue552 start values

### LINKED ISSUE #552: Removing initialisation values from pressure drop components
Many pressure drop components contain an entry in the parameter window for setting the start value of `m_flow` and `dp`. We should remove these because they distract attention from the more important parameters `m_flow_nominal` and `dp_nominal`.

mwetter:
This closes #552.

I also moved the computation of `dp` from the `equation` section to the variable declaration so that the model can be used for graphical modeling without having to mix graphical models with textual equations in the `equation` section, which is something we want to avoid per our style guide.


Mathadon:
thanks for fixing this!

Issue #584, opened 2016-11-10
add more realistic control to SimpleHouse


mwetter:
This model has on/off control for the damper, and also the pump is continuously operating.
This branch is to update the model to be more realistic as it will also be used in the final report.

mwetter:
This is now on the master

PR #585, opened 2016-11-10
Updated SimpleHouse for #584

mwetter:
This will close #585


mwetter:
- [ ] The supply air should also be changed to be connected to the outside temperature. I will work on that.

mwetter:
@Mathadon It is now ready for review and merging by you.

PR #598, opened 2016-11-21
Issue472 fluid.fmi

### LINKED ISSUE #472: Inclusion of Buildings.Fluid.FMI
This issue is to discuss whether `Buildings.Fluid.FMI`should be included in the Annex60 library. 
This is a package which Activity 1.2 was planning to reference in the report. Best might be if the reference was pointing to the Annex60 library as opposed to the Buildings library.
This might also be relevant for potential future work done in the IBPSA project 1.

mwetter:
This closes #472


Annex60/Fluid/FMI/Adaptors/ThermalZone.mo
```diff
@@ -0,0 +1,380 @@
...
+<p>
+Block that converts a vector input for the water mass fraction <code>Xi</code>
+to a scalar output <code>X</code>.
+This is needed for models in which a scalar input signal <code>Xi</code> that
+may be conditionally removed is to be connected to a model with a vector
+input <code>X</code>, because the conversion from scalar to vector
+needs to access the conditional connector, but conditional connectors
+can only be used in <code>connect</code> statements.
+</p>
+</html>"));
```
marcusfuchs: @mwetter 
I meant to suggest to have all the documentation for the top-level model and the internal blocks in one place at the top-level, as some users may overlook the internal block documentation when they read the documentation of the top-level model. But having a second look at it, I also see your point in putting the further information to the internal blocks, keeping the top-level documentation easier to read. I'll merge as is.

PR #607, opened 2016-12-03
Issue480 fixed resistance

### LINKED ISSUE #480: DeltaM and ReC variables
We currently have the following parameters in `Annex60.Fluid.FixedResistances.FixedResistanceDpM`:

```
  parameter Boolean use_dh = false "Set to true to specify hydraulic diameter";
  parameter Modelica.SIunits.Length dh=1 "Hydraulic diameter";
  parameter Real ReC(min=0)=4000 
    "Reynolds number where transition to turbulent starts";
  parameter Real deltaM(min=0.01) = 0.3 
    "Fraction of nominal mass flow rate where transition to turbulent occurs";
```

I have two remarks:
- The only use of these parameters is to determine when the regularisation around zero flow is activated, which has a limited effect on model results. Still these parameters are prominently present in the parameter window, suggesting their values should be chosen well. I propose to move those parameters to a different tab.
- The default values of these parameters (mostly deltaM) result in the regularisation region being activated quite quickly, already at 30% nominal flow. I would suggest to reduce this value drastically to 1e-4 or similar since there is really no reason to have such a broad regularisation region. Moreover it leads to 'wrong' results. This would require to also lower the minimum value of deltaM.
- `ReC` and `dh` give the impression that the flow function/model considers the existence of a 'turbulent' regime (and hence also a non-turbulent regime), which is not true since the flow is always turbulent, even in the regularisation region. Moreover it requires physical insight to know which diameter the user should provide, which can be confusing. Therefore we could even remove these variables.

mwetter:
This closes #480 

As this is non-backward compatible, @Mathadon @mlauster @thorade (or anyone from your team) needs to agree within the next 2 weeks if there are objections with proceeding with the final review and merging the pull request (see https://github.com/iea-annex60/modelica-annex60/wiki/Workflow-for-code-changes#non-backward-compatible-changes).

The following changes have been done, and all changes have also been tested and integrated with `Buildings` (on the master branch). I also revised `FixedResistances.{Examples,Validation}` as many models had sparse to no documentation and the formatting was very inconsistent (some date back to 2007).

The model Annex60.Fluid.FixedResistances.FixedResistanceDpM has been refactored. Now, if the hydraulic diameter is not yet known, one can use the simpler model Annex60.Fluid.FixedResistances.PressureDrop, otherwise the model Annex60.Fluid.FixedResistances.HydraulicDiameter may be used. With this refactoring, also the model Annex60.Fluid.FixedResistances.SplitterFixedResistanceDpM has been renamed to Annex60.Fluid.FixedResistances.Junction and parameters that use the hydraulic diameter have been removed. 

Renamed Annex60.Fluid.FixedResistances.FixedResistanceDpM to Annex60.Fluid.FixedResistances.PressureDrop and removed the parameters use_dh, dh and ReC. For Dymola, the conversion script will update models. If a model needs to be used that allows specifying dh and ReC, then the new model Annex60.Fluid.FixedResistances.HydraulicDiameter can be used. 

Renamed Annex60.Fluid.FixedResistances.SplitterFixedResistanceDpM to Annex60.Fluid.FixedResistances.Junction and removed the parameters use_dh, dh and ReC. For Dymola, the conversion script will update models. If a model needs to be used that allows specifying dh and ReC, then use Annex60.Fluid.FixedResistances.Junction with dp_nominal = 0 (which removes the pressure drop) and use Annex60.Fluid.FixedResistances.HydraulicDiameter at each fluid port. 

Here is the entry of the conversion script that we use for Buildings:
```
// Pressure drop
convertModifiers({"Annex60.Fluid.FixedResistances.FixedResistanceDpM"},
                  {"dh"},
                  fill("", 0),
                  true);

convertModifiers({"Annex60.Fluid.FixedResistances.FixedResistanceDpM"},
                  {"ReC"},
                  fill("", 0),
                  true);

convertModifiers({"Annex60.Fluid.FixedResistances.FixedResistanceDpM"},
                  {"use_dh"},
                  fill("", 0),
                  true);

convertClass("Annex60.Fluid.FixedResistances.FixedResistanceDpM",
             "Annex60.Fluid.FixedResistances.PressureDrop");

// Flow junction
convertModifiers({"Annex60.Fluid.FixedResistances.SplitterFixedResistanceDpM"},
                  {"dh"},
                  fill("", 0),
                  true);

convertModifiers({"Annex60.Fluid.FixedResistances.SplitterFixedResistanceDpM"},
                  {"ReC"},
                  fill("", 0),
                  true);

convertModifiers({"Annex60.Fluid.FixedResistances.SplitterFixedResistanceDpM"},
                  {"use_dh"},
                  fill("", 0),
                  true);

convertClass("Annex60.Fluid.FixedResistances.SplitterFixedResistanceDpM",
             "Annex60.Fluid.FixedResistances.Junction");
```


Annex60/Fluid/FixedResistances/Validation/PressureDropsExplicitLinearized.mo
```diff
@@ -0,0 +1,30 @@
+within Annex60.Fluid.FixedResistances.Validation;
```
Mathadon: good test, thx!

Mathadon:
looks good! thanks for these changes and for the script

mwetter:
@mlauster @thorade A friendly reminder that you have one more week to raise objections for this non-backward compatible change.

thorade:
Thanks for the reminder and sorry for answering late. BuildingSystems adds only two models to the Annex60 models that use FixedResistanceDpM and I will file an issue with our library to update them.

mlauster:
Hm, I thought I already gave my opinion to that, but github proves the opposite.
We might already have some models using `FixedResistanceDpM`, but we will follow the way of UdK-VPT and file an issue to update our models. So go ahead!

mwetter:
Thanks for the feedback.
@Mathadon : Does your "looks good" mean you already reviewed the models and they can be merged, or does this simply mean you agree with the change? If it is the latter, could you please review?

Mathadon:
I did review the code. Sorry for being so cryptic!

mwetter:
@Mathadon No problem, and thanks for taking time to review.

PR #626, opened 2016-12-14
Issue602 polynomial

### LINKED ISSUE #602: remove derivative annotation in Annex60.Utilities.Math.Functions.polynomial
This annotation causes a numerical derivative to be used in `translateModel("Buildings.Fluid.HeatPumps.Validation.ReciprocatingWaterToWater");` on branch `issue587_lowLiftHeatPump`, commit aa250631103d962c767ebe9471e22f541cf89a56

Also, the example should be removed and the model tested to make sure no numerical derivatives are introduced in the regression tests.

mwetter:
This closes #602.
All regression tests work, also for the Buildings library, without any numerical Jacobian.


Mathadon:
Do you understand why it did not work with the derivative annotation?

mwetter:
I am not sure if it was caused by the `zeroDerivative=a` annotation. We tested it by removing the derivative annotation and the model had no numerical Jacobian anymore.

PR #636, opened 2016-12-21
Issue623 rom assertions refactored2

### LINKED ISSUE #623: Assert sensitive to state variable in JModelica
In all these models there is an assert
`if time >= assEqu.t0 and time <= assEqu.t1 or time >= assEqu.t2 and time <= assEqu.t3 or time >= assEqu.t4 and time <= assEqu.t5 then
  assert(noEvent(abs(assEqu.u1 - assEqu.u2) < assEqu.threShold), "Inputs differ by more than threShold" + "\n" + "  time       = " + String(time) + "\n" + "  u1         = " + String(assEqu.u1) + "\n" + "  u2         = " + String(assEqu.u2) + "\n" + "  abs(u1-u2) = " + String(abs(assEqu.u1 - assEqu.u2)) + "\n" + "  threShold  = " + String(assEqu.threShold));
 end if;`
which makes JModelica fail:
Annex60.ThermalZones.ReducedOrder.Validation.VDI6007.TestCase2
Annex60.ThermalZones.ReducedOrder.Validation.VDI6007.TestCase4
Annex60.ThermalZones.ReducedOrder.Validation.VDI6007.TestCase6
Annex60.ThermalZones.ReducedOrder.Validation.VDI6007.TestCase7
Annex60.ThermalZones.ReducedOrder.Validation.VDI6007.TestCase11

The problem is similar to the problem discussed in #564.
Could the assert be removed and the assert be checked in the result verification instead?

mwetter:
This closes #623


mwetter:
@mlauster 
You will need to put variables that have different orders of magnitudes in separate subplots or plot windows. Otherwise, one does not see their value in the test scripts:
testcase1
Please make these changes before I can merge.
I will also run the tests for JModelica as I put in an assertion again, but this time in the `terminal()` section which should work.

PR #639, opened 2017-01-02
added example model using FourPortHeatMassExchanger

### LINKED ISSUE #353: Annex60.Fluid.Interfaces.FourPortHeatMassExchanger is not used

Mathadon:
and unit test

this closes #353


Annex60/Fluid/Interfaces/Examples/FourPortHeatMassExchanger.mo
```diff
@@ -0,0 +1,79 @@
...
+  connect(ram_m_flow.y, sou1.m_flow_in) annotation (Line(points={{-79,30},{-66,
+          30},{-66,18},{-60,18}}, color={0,0,127}));
+  connect(ram_T.y, sou1.T_in) annotation (Line(points={{-79,10},{-70,10},{-70,
+          14},{-62,14}}, color={0,0,127}));
+  annotation (Icon(coordinateSystem(preserveAspectRatio=false)), Diagram(
+        coordinateSystem(preserveAspectRatio=false)),
+    Documentation(info="<html>
+<p>
+This example model demonstrates the used of the
+<code>FourPortHeatMassExchanger</code> model.
```
mwetter: Use a hyperlink, otherwise we cannot detect if a model name becomes invalid.

Mathadon: How does this detection work exactly? Or when does it happen?

mwetter:
@mathadon: The function buildingspy.development.refactor.move_class will
only replace the full name, and wrong hyperlinks can be detected in the
html generated from Dymola, whereas a wrong class name inside
<code>...</code> is hard to detected.

PR #658, opened 2017-01-18
Issue657 splitter validation

### LINKED ISSUE #657: change validation test due to tool-dependent choice when using inStream operator
There is a difference between JModelica and Dymola in the choice of eps for the inStream operator (see Modelica specification 3.3 p. 179) which affects the calculation for near zero or zero flows. In JModelica the smallest of the flow nominals is used as epsilon for all flows in the implementation whereas Dymola has specific epsilons for each of the flows.
 
This difference affects the verification of Annex60.Fluid.FixedResistances.Validation.FlowJunctionSteadyStateNoPressureDrop where the nominals for spl.res1.m_flow, spl.res2.m_flow, spl.res3.m_flow are not equal. To solve this issue, either:

1. Remove the signals senTem1.T, senTem2.T and senTem3.T from the verification
2. Adjust the nominal values so that spl.res1.m_flow, spl.res2.m_flow and spl.res3.m_flow have the same nominals

mwetter:
This closes #657 .

The reference results have been updated.


mwetter:
@Mathadon Can you please review and merge.

PR #661, opened 2017-01-20
Fix quantity and year

tbeu:


mwetter:
@tbeu Thanks for the pull request.
@Mathadon Can you please review and merge if the change in your code is OK.

PR #667, opened 2017-02-13
Update ConstantEffectiveness.mo

carlesRT:
Independently of inflow occurs on "a1 or b1"  and "a2 or b2" counterflow configuration is assumed.


carlesRT:
There is no need to add this remark... The fact that constant effectiveness is used should be sufficient.

Issue #687, opened 2017-03-22
set min=max=0 zero for m_flow in sensors


Mathadon:
Single port sensors could be further simplified since their mass flow rate is always zero.

I.e. 
```
Modelica.Fluid.Interfaces.FluidPort_a port(redeclare package Medium=Medium, m_flow(min=0));
```
should be changed into
```
Modelica.Fluid.Interfaces.FluidPort_a port(redeclare package Medium=Medium, m_flow(min=0, max=0));
```

PR #688, opened 2017-03-22
set m_flow(max=0) in PartialAbsoluteSensor

Mathadon:
for #687

I ran unit tests in `Fluid.Sensors` which yielded 1 false negative.


Mathadon:
Travis is so cool..

PR #701, opened 2017-04-04
Issue50 new delay mod

bramvdh91:
Implement new delay tracker with separate memory variable for forward and reverse flow, eliminating complicated tests


bramvdh91:
Wrong pull request, this can be ignored/deleted

Issue #700, opened 2017-04-04
Documentation ReducedOrder: Add reference to TEASER


PRemmen:
Since TEASER is an open-source project with a stable version and can be used to calculate all parameters and generate code for all four reduced order models contained in IBPSA library, I suggest to add a reference of TEASER to the documentation of `ReducedOrder` package description.

Repository: https://github.com/RWTH-EBC/TEASER

Documentation: http://rwth-ebc.github.io/TEASER/

Any objections? :)

mwetter:
@PRemmen Adding such a reference would be good. Can you please make a pull request against the branch `issue700_romDocumentation`.

Also, on http://rwth-ebc.github.io/TEASER/ I suggest you list the four libraries as we want end-users to use any of the four libraries directly rather than the IBPSA library.
The section is:

> export of individual dynamic simulation models for Modelica libraries AixLib and Modelica IBPSA library.

PR #704, opened 2017-04-05
Add reference to TEASER repository, #700

PRemmen:
This adds a reference to the TEASER repository for parameter calculation, solves #700


IBPSA/ThermalZones/ReducedOrder/package.mo
```diff
@@ -29,6 +29,11 @@ package ReducedOrder
     smooth=Smooth.None)}), Documentation(info="<html>
     <p>This package contains models for reduced building physics of thermal
     zones and accompanying models for consideration of solar radiation and
-    radiation transmission through windows. </p>
+    radiation transmission through windows.</p>
+    <p>The python package TEASER <a
+    href=\"https://github.com/RWTH-EBC/TEASER\">
+    https://github.com/RWTH-EBC/TEASER</a> can be used to calculate reduced
+    order parameters and generate ready-to-tun models.
```
mlauster: Typo, "ready-to-run"

mlauster:
We should add this sentence to `RC.UsersGuide` as well, I think this is the major source of information and most users will have a look here (if they read the docs at all).

mlauster:
@mwetter, all tests are failing, probably because it's issued from outside ibpsa/modelica ,@PRemmen is not yet part of it. As this PR only includes changes on documentation, I vote for merging it. Objections from your side?

thorade:
Peter created the PR against master. In the top right corner there is a button `Edit`. You can change the branches used for the Pull Request, change it so that the PR is not against master, but against the issue branch here on the repository (it has the same name). You can safely merge Peters code to the issue branch and then create a new Pull Request, asking to merge the IBPSA/issue-branch into IBPSA/master. That PR should run without problems.
Or, add Peter to the IBPSA organization.

PRemmen:
@thorade thanks for thehint. I think it was supposed to be like this any way :+1: 

@mlauster I changed base so you can hopefully savely merge into `issue700_romDocumentation` and then create a PR for master by yourself

PR #721, opened 2017-04-12
Remove default experiment attributes

### LINKED ISSUE #720: Remove default stopTime
This issue is to remove the default `StopTime=1.0` from the `experiment` annotation of Modelica models

tsnouidui:
@Mathadon Could you please review and merge. This hasen't changed the unit tests results
This closes #720.


Mathadon:
@tsnouidui looks good but travis won't let me merge due to the [ci skip]. 
It's probably easiest to add a dummy commit and let the unit tests run just to get this out of the way?

tsnouidui:
@Mathadon I just did.

tsnouidui:
@mathadon Please wait before merging.
I came across this post https://github.com/lbl-srg/modelica-buildings/issues/250 and I am checking with @mwetter if we indeed need to remove the `StopTime` here? Based on what I am reading we actually want the opposite unless it is now obsolete for OpenModelica.

mwetter:
@sjoelund : Is the annotation
```
experiment(Tolerance=1e-6)
```
sufficient for OpenModelica to run a model in its regression tests, or does it need a `StopTime`, e.g.,
```
experiment(Tolerance=1e-6, StopTime=1.0)
```

mwetter:
@sjoelund Is `experiment(Tolerance=1e-6)` sufficient for OpenModelica's regression tests? Please see above.

sjoelund:
OpenModelica does indeed require a stoptime to perform its regression testing:
```mo
loadString("
model M
  annotation(experiment(StopTime=1.0, Tolerance=1e-6));
end M;
model N
  annotation(experiment(Tolerance=1e-6));
end N;
");
isExperiment(M); // true
isExperiment(N); // false
```

Otherwise, how would you know which timescale or stop-time to use?

tsnouidui:
Thanks @sjoelund . @mwetter I should in this case keep the `StopTime` right? I will also change BuildingsPy to check if `StopTime` exists in future checked mo files.

mwetter:
@tsnouidui Unless the stop time is different from 1, please add `StopTime=1` as the Modelica 3.3 specification does indeed not define a default for `StopTime`. (For the record, there is also no default for `StartTime` in the specification, but Dymola, OpenModelica and JModelica seem to assume 0 without a warning, maybe because this is the only meaningful default.)

tsnouidui:
@mwetter Shouldn't we rather reject the model as part of unit testing in travis and let the modeler add this to the file. I will otherwise end up modifying mo files which in turn will require to create a new branch with the changes, run tests, let someone review, and merge to the master?

mwetter:
@tsnouidui : Can you make a script that fixes all models in the library once, and afterwards we can flag this as an error when we get a new model that has not this annotation. But fixing the 600+ tests and getting them to a valid format is easier done by a script.

tsnouidui:
@mwetter Luckily, all models of the Buildings and IBPSA library have `StopTime` defined...

mwetter:
@tsnouidui I will close it as it is no longer needed.

tsnouidui:
Thanks!

Issue #730, opened 2017-04-18
Quintic Hermite spline interpolation


Mathadon:
Quintic Hermite spline interpolation expressions are presented on page four of https://www.rose-hulman.edu/~finn/CCLI/Notes/day09.pdf 

We should add this to the utilities package to allow C2 continuous regularisations.

mwetter:
Closed as it is on the master.

PR #735, opened 2017-04-19
added quintic Hermite spline model for #730

Mathadon:
added function, example, unit test

result:
screen shot 2017-04-19 at 14 07 01

transition is between `-time` and `exp(0.5*time)` in range [1, 2.6]


Mathadon:
The peak in the derivative that you see on the figure seems to be infinitely narrow. I'm not sure why it pops up.. Adding `noEvent()` fixes the problem. Should we report this to Dymola?

Mathadon:
New result after some changes:
screen shot 2017-04-19 at 14 16 04

Issue #737, opened 2017-04-20
Non-smooth min() function in PartialEffectiveness


Mathadon:
In `IBPSA.Fluid.HeatExchangers.BaseClasses.PartialEffectiveness` we currently have 
```
CMin_flow = min(C1_flow, C2_flow);
```
which is C0 and could lead to problems when this equation ends up in an algebraic loop.

I'll add an implementation using `smoothMin`

Mathadon:
As suggested in #736 we could add something like:

```
  if smoothness == IBPSA.Types.Smoothness.C0 then
    CMin_flow = min(C1_flow, C2_flow);
  elseif smoothness == IBPSA.Types.Smoothness.C1 then
    CMin_flow = Modelica.Fluid.Utilities.cubicHermite(...);
  else
    CMin_flow = Modelica.Fluid.Utilities.quinticHermite(...);
  end if;
```

I'm just not sure how to treat `C1_flow=C2_flow=0`.. Since this leads to a division by zero.

Mathadon:
Actually, this seems to work rather well for me:

```
CMin_flow = C1_flow + C2_flow - ((C1_flow)^10 + (C2_flow)^10)^(1/10);
```

this returns 0 when `C1_flow` or `C2_flow` or both are zero. Since both arguments should always be positive, I think we can use it in this case. I think the largest relative error this produces is 2^(1/10). Which in this case is 7%. The power 10 can be increased/reduced based on the required accuracy.

@mwetter would this be a good approach?

Mathadon:
Moreover I propose to replace

```
  C1_flow = abs(m1_flow)*
    ( if allowFlowReversal1 then 
           fra_a1 * Medium1.specificHeatCapacityCp(state_a1_inflow) +
           fra_b1 * Medium1.specificHeatCapacityCp(state_b1_inflow) else 
        Medium1.specificHeatCapacityCp(state_a1_inflow));
  C2_flow = abs(m2_flow)*
    ( if allowFlowReversal2 then 
           fra_a2 * Medium2.specificHeatCapacityCp(state_a2_inflow) +
           fra_b2 * Medium2.specificHeatCapacityCp(state_b2_inflow) else 
        Medium2.specificHeatCapacityCp(state_a2_inflow));
```
by
```
  C1_flow = 
      if allowFlowReversal1 then 
           abs(m1_flow)*(fra_a1 * Medium1.specificHeatCapacityCp(state_a1_inflow) +
[...]

Issue #736, opened 2017-04-20
C2 continuity


Mathadon:
I'm experiencing problems when optimising Annex60-based models that I think are related to models not being C2 continuous. Functions such as `smoothMax` can (hopefully) be updated to be C2 continuous, but this will most likely increase the number of operations required for evaluating the function.
I was therefore thinking about the option to introduce new C2-continuous transition functions, but not to use them by default. We could then create a parameter (e.g. in `pressureIndependent`) that allows the user to choose between a C1 (cubicHermite), C2 (quinticHermite) or even C0 (normal min/max function) implementation. 
Good idea?

Mathadon:
Actually it looks like `SpliceFunction` is already C2, if not C-infinite.

Following example
```
model SpliceFunction
  Real y "Function value";
  Real dy;
  Real dy2;
  Real dy3;
equation 
  y=IBPSA.Utilities.Math.Functions.spliceFunction(pos=time^2, neg=-exp(time), x=time-0.4, deltax=0.2);
  dy= der(y);
  dy2=der(dy);
  dy3=der(dy2);

end SpliceFunction;
```
leads to:
![screen shot 2017-04-20 at 12 45 55](https://cloud.githubusercontent.com/assets/5970879/25249708/abf3b070-25c7-11e7-90f3-710d2a90b0f3.png)

Is my analysis correct that this transition is (infinitely) smooth?

mwetter:
The only way to check if `spliceFunction` is more than C1 is to take its derivatives at `x +/- deltax` and compare them with the functions that are used for `pos` and `neg`. I would rely more on a pencil and using arbitrary functions for `pos` and `neg` than on a plot.

Mathadon:
From following python script:

```
from sympy import *

scaledX = Symbol('scaledX')
pos = Symbol('pos')
neg = Symbol('neg')

y = (tanh(tan(asin(1)*scaledX))+1)/2

out = y*pos(scaledX) + (1-y)*neg(scaledX)

print out.subs(scaledX, 0.999999).evalf()
print out.subs(scaledX,-0.999999).evalf()
for i in range(1,6):
    print str(i) + "-th Derivative"
    print diff(out,scaledX,i).subs(scaledX,0.999999).evalf()
    print diff(out,scaledX,i).subs(scaledX,-0.999999).evalf()
```
where I analytically compute the i-th derivative of `IBPSA.Utilities.Math.Functions.spliceFunction`:

```
[...]

PR #738, opened 2017-04-20
using smoothMin in IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffe…

Mathadon:
…ctiveness.mo

this is for #737


IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectiveness.mo
```diff
@@ -83,7 +83,7 @@ equation
            fra_a2 * Medium2.specificHeatCapacityCp(state_a2_inflow) +
            fra_b2 * Medium2.specificHeatCapacityCp(state_b2_inflow) else
         Medium2.specificHeatCapacityCp(state_a2_inflow));
-  CMin_flow = min(C1_flow, C2_flow);
+  CMin_flow = IBPSA.Utilities.Math.Functions.smoothMin(x1=C1_flow, x2=C2_flow, deltaX=CMin_flow_small);
```
mwetter: I am not convinced that this is not giving large spikes if C1_flow = 0 and C2_flow transitions through 0. Then, you can get an undershoot of min(C1_flow, C2_flow), leading to a large ratio of CMin_flow/min(C1_flow, C2_flow), and therefore large QMax_flow/min(C1_flow, C2_flow), which in turn leads to large outlet temperatures near zero flow. This is why I refrained from using `smoothMin` but rather used `min`.

Mathadon: A better solution could then be to use a cubic/quintic Hermite spline? Since these should not have an overshoot or undershoot?

Mathadon:
As suggested in #736 we could add something like:

```
  if smoothness == IBPSA.Types.Smoothness.C0 then
    CMin_flow = min(C1_flow, C2_flow);
  elseif smoothness == IBPSA.Types.Smoothness.C1 then
    CMin_flow = Modelica.Fluid.Utilities.cubicHermite(...);
  else
    CMin_flow = Modelica.Fluid.Utilities.quinticHermite(...);
  end if;
```

I'm just not sure how to treat `C1_flow=C2_flow=0`.. Since this leads to a division by zero.

mwetter:
Let's close this as handling singularities in `Buildings.Fluid.HeatExchangers.BaseClasses.epsilon_ntuZ` is quite involved.

PR #752, opened 2017-04-25
Reimplemented check for energy conservation

### LINKED ISSUE #741: too sensitive assert in StaticTwoPortConservationEquation
I'm getting following warning (after having extended the assert output):

```
noEvent((if abs(adsolairs.bureaux2.heatingOut.m_flow) < 1E-60 then 
abs(adsolairs.bureaux2.heater.Q1_flow) < 1E-10 else 
abs(adsolairs.bureaux2.heater.Q1_flow/adsolairs.bureaux2.heater.bal2.vol.cp_default/adsolairs.bureaux2.heatingOut.m_flow) < 200))
The following was detected at time: 1404170000
Energy may not be conserved for small mass flow rates. 
This model may require prescribedHeatFlowRate = false. Found 
Q_flow=-7.41038e-18 
abs(Q_flow)=  7.41038e-18 
abs(m_flow) = 9.75149e-43
```

`m_flow` is solved from an algebraic loop and should be zero, but instead it is just very small. This value is used to compute `Q_flow` through `CMin_flow = C1_flow + C2_flow - ((C1_flow)^10 + (C2_flow)^10)^(1/10);`. This last computation subtracts `C2_flow - (C2_flow^6 + 0)^(1/6)` with `C2_flow` nonzero, which is relatively large due to numerical noise, although still very small (7e-18). The assert is correctly triggered since this heat flow rate corresponds to a huge temperature difference (~1e20K).

I propose to change

```
    assert(noEvent(if abs(m_flow) < Modelica.Constants.small
      then 
        abs(Q_flow) < 1E-10
      else 
        abs(Q_flow/cp_default/m_flow) < 200);
```
into
```
    assert(noEvent(if abs(m_flow) < 1e-10
      then 
        abs(Q_flow) < 1E-10
      else 
        abs(Q_flow/cp_default/m_flow) < 200);
```
to solve this.

mwetter:
This closes #741 if all tests pass.

### DIFF
--- a/IBPSA/Fluid/Interfaces/StaticTwoPortConservationEquation.mo
+++ b/IBPSA/Fluid/Interfaces/StaticTwoPortConservationEquation.mo
@@ -144,12 +144,8 @@ equation
   end if;
 
   if prescribedHeatFlowRate then
-    assert(noEvent(if abs(m_flow) < Modelica.Constants.small
-      then
-        abs(Q_flow) < 1E-10
-      else
-        abs(Q_flow/cp_default/m_flow) < 200),
-   "Energy may not be conserved for small mass flow rates. This model may require prescribedHeatFlowRate = false.");
+    assert(noEvent( abs(Q_flow) < 200*cp_default*max(m_flow_small/1E3, abs(m_flow))),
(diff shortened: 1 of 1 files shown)


Mathadon:
@mwetter this can be merged

PR #754, opened 2017-04-25
Issue156 pressure independent valve

### LINKED ISSUE #156: Pressure-independent valve model
The Annex60/Building library does not seem to contain a model for a pressure-independent valve (example: http://www.belimo.ch/pdf/e/r2p_2_e.pdf). These valves supply a desired mass flow rate regardless of the supplied pressure (as long as the pressure is high enough).
I intend to make such a model. Is this something that I should make a pull request for?

Mathadon:
this closes #156


IBPSA/Fluid/Actuators/Valves/TwoWayPressureIndependent.mo
```diff
@@ -12,69 +12,115 @@ model TwoWayPressureIndependent "Model of a pressure-independent two way valve"
...
+           actual=y_smooth,
+           simplified=dp_nominal_pos*m_flow/m_flow_nominal_pos);
+    end if;
+  else
+    if from_dp then
+      m_flow=y_smooth;
+    else
+      dp=y_smooth;
+    end if;
+  end if;
```
mwetter: `y_smooth = if from_dp then m_flow else dp`, hence the units of `y_smooth` change depending on the parameter `from_dp`. I don't know if this is valid Modelica and hence suggest to change it so that you have a different variable for each case. Then, you can specify its units when you declare the variable.

Mathadon: I wasn't sure about this either. I'll change it.

PR #756, opened 2017-04-26
added tolerance for #755

Mathadon:
this is for #755


tsnouidui:
@mwetter Please merge this to the Buildings master when you get the chance to fix the problem there too.

PR #757, opened 2017-04-27
Add non-zero start attributes

tbeu:
A discrete variable that is used with its pre-value as array index requires a non-zero start attribute. Otherwise the pre-value is set to zero (default start value) resulting in an array-index-out-of-bounds error (in SimulationX).


tbeu:
Hm, why does Travis CI all fail?

thorade:
Two reasons, I think: 
you are not a member of IBPSA and docker won't work for you, 
and Travis has some problems currently:
https://www.traviscistatus.com/ says

tbeu:
Should non-IPBSA-members commit using [ci skip] then?

thorade:
Does not matter too much, it fails quickly, with message "Docker login failed".
The workflow, as I understand it, is that your PR will be merged to a branch here on the repository (you can change the base branch, too, in the top right corner) by an IBPSA member with sufficient rights to create new branches, and then in a second step it will be merged into master.

mwetter:
@tbeu Thanks for the pull requests. CI only works for organization members. Please submit your PR as before and then we move it onto another branch for running the CI before committing. Please do not use [ci skip] as this will prevent our tests to run too, unless we make another commit.

mlauster:
@mwetter ans @thorade, is this failing of CI tests for non-members only an issue of License/Password security? Then you could possibly encrypt them as I read here:
https://docs.travis-ci.com/user/deployment/pypi/
I didn't take a deeper look, don't know if it helps in any way.

thorade:
Non-members should not be able to use the license, because otherwise anybody could use it to do anything in their fork or in their own repository.
Once all tests pass with open source tools, that could be an option.
Or maybe switch to Jenkins instead of Travis? And use a computer with more cores and more RAM? buildingspy runs tests in parallel, so using 32 cores instead of 4 would result in a speedup of factor ~8.

mlauster:
And encrypting the license so that only Travis can read it (and only in combination with our repo) does not solve that problem?

tbeu:
That is the usual way to keep private settings (like passwords) enclosed.

tbeu:
And that only works for the original repo, but not for forks.

mlauster:
For to that topic:

https://docs.travis-ci.com/user/encrypting-files/

mlauster:
But that page states that
"Please note that encrypted files are not available for pull requests from forks."
So possibly that does not solve this problem.

thorade:
Yes, but then pull requests would still fail, as I understand it!?
So that would be a different way to achieve a similar result.
Setting environment variables from the Travis homepage seems to be easier.

Issue #759, opened 2017-04-27
water index computed using initial algorithm


Mathadon:
In `IBPSA.Utilities.Psychrometrics.TWetBul_TDryBulXi` the Medium water index is computed using

```
initial algorithm
   iWat:=-1;
     for i in 1:Medium.nX loop
       if Modelica.Utilities.Strings.isEqual(string1=Medium.substanceNames[i],
                                             string2="Water", caseSensitive=false) then
         iWat :=i;
       end if;
     end for;
```

this is causing problems when doing optimisations since the water medium index is unknown at translation time.

This can be resolved using 

```
parameter Integer iWat = sum( { (if Modelica.Utilities.Strings.isEqual(string1=Medium.substanceNames[i],string2="Water", caseSensitive=false) then i else 0) for i in 1:Medium.nX})
```

but that in turn leads to an error when checking the model:

`
Index 1 of Xi[iWat] is 0, which is outside of range 1 to 0.
`

This can be resolved by setting as default medium `IBPSA.Media.Air`instead of `Modelica.Media.Interfaces.PartialCondensingGases `.

I propose to make these changes.

thorade:
might be related: #675 and #689

mwetter:
@Mathadon Would 
```
parameter Integer iWat(min=1, start=1) = ...
```
help?

Mathadon:
That works too! But it's not really a proper solution, is it?

Mathadon:
But then dymola complains:

```
Index 1 of Xi[iWat] is 1, which is outside of range 1 to 0.
```
and, more surprisingly:
```
The parameter wetBul.iWat has the attribute fixed = false
indicating its value to be calculated from initial conditions.
Error: However, it is used to specify a structural property and
it must be possible to evaluate it at translation.
```
somehow dymola (and apparently also JModelica) does accept this without the `start` and `min` attributes.

mwetter:
If this causes issues we can't work around, I would be fine with changing the protected parameter to a public constant of the from
``` modelica
  constant Integer iWat=1 "Index of water in medium composition vector"
    annotation (HideResult=true);
```
and then add an assert to make sure that water is at the right index.

However, you still have the same problem with `IBPSA.Fluid.Sensors.TraceSubstances` where such a fix won't work.

Mathadon:
That's an idea too.

What is wrong with the solution that I proposed, though?

mwetter:
[...]

PR #761, opened 2017-05-01
changed computation of iWat in TWetBul_TDryBulXi for #759

Mathadon:
for #759

### DIFF
--- a/IBPSA/Utilities/Psychrometrics/TWetBul_TDryBulXi.mo
+++ b/IBPSA/Utilities/Psychrometrics/TWetBul_TDryBulXi.mo
@@ -41,9 +41,12 @@ protected
   Modelica.SIunits.MassFraction XiSatRefIn
     "Water vapor mass fraction at saturation, referenced to inlet mass flow rate";
 
- parameter Integer iWat = sum( { (if Modelica.Utilities.Strings.isEqual(string1=Medium.substanceNames[i],string2="Water", caseSensitive=false) then i else 0) for i in 1:Medium.nX})
-    "Index of water in medium composition vector";
-initial algorithm
+ parameter Integer iWat = sum({(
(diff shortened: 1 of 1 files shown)


IBPSA/Utilities/Psychrometrics/TWetBul_TDryBulXi.mo
```diff
@@ -4,8 +4,7 @@ block TWetBul_TDryBulXi
   extends Modelica.Blocks.Icons.Block;
 
   replaceable package Medium =
-    Modelica.Media.Interfaces.PartialCondensingGases "Medium model"
-                                                            annotation (
+    IBPSA.Media.Air "Medium model"                          annotation (
```
mwetter: Why is this change needed? Neither Dymola nor JModelica needs it.

Mathadon:
If I don't change it, then I get:
screen shot 2017-05-01 at 15 26 56

mwetter:
@Mathadon : Is this a problem, as you check the block with unassigned medium? The following works in Dymola 2017 FD01 with the pedantic mode:
```
checkModel("IBPSA.Utilities.Psychrometrics.Examples.TWetBul_TDryBulXi");
```
We have the same issue with any model in `.Fluid` unless a medium is assigned.

Mathadon:
I get the error in the screenshot and 
```
Advanced.PedanticModelica:=true;
checkModel("IBPSA.Utilities.Psychrometrics.TWetBul_TDryBulXi");
 = false
```
with `Medium = Modelica.Media.Interfaces.PartialCondensingGases` and no errors with `IBPSA.Media.Air`. 

So I understand you don't get these errors in 2017FD01? So I should revert back to the old Medium?

Mathadon:
Yes, when I check the block with the default Medium then I get that error.

The example is fine. The old implementation didn't have that error so I wanted to avoid it, but I did indeed not realise that there are more models that have this effect when the medium is unassigned.

I'll revert the medium change.

PR #764, opened 2017-05-03
Issue271 linear damper

### LINKED ISSUE #271: add valve and damper with linear relationship between flow and lift for given authority
For valves and air dampers, the mass flow rate increases sharply at low lifts, in particular, but not only, if the authority is low. This causes a large control gain at low lifts, which in turn can cause instability in the feedback control that may require a compensation at the input signal of the valve. Such instabilities can significantly affect computing time.

This ticket is to implement either a simple valve and air damper model, or to add to the existing models a parameter that causes the mass flow rate to be a linear function of the opening signal for the given valve (or damper) authority.

mwetter:
This closes #271


mwetter:

PR #766, opened 2017-05-03
Inherit function icon

tbeu:


tbeu:
OK, #768 is the updated PR.

PR #767, opened 2017-05-03
Fix inheritance

thorade:
this replaces #766 
as only icons are added, the tests should pass fine


thorade:
@tbeu it seems the changes in 
IBPSA/Controls/SetPoints/OccupancySchedule.mo
break the simulation & tests

tbeu:
OK, #768 is the updated PR.

IBPSA/Controls/SetPoints/OccupancySchedule.mo
```diff
@@ -35,6 +35,8 @@ protected
   output Modelica.SIunits.Time tNonOcc "Time when next non-occupancy starts";
 
 encapsulated function switchInteger
+  import Modelica;
```
mwetter: Why do we need to import Modelica?

IBPSA/Controls/SetPoints/OccupancySchedule.mo
```diff
@@ -45,6 +47,8 @@ algorithm
 end switchInteger;
 
 encapsulated function switchReal
+  import Modelica;
```
mwetter: Same question as above.

IBPSA/Controls/SetPoints/OccupancySchedule.mo
thorade: Because the function is encapsulated, without the import the `extends Modelica.Icons.Function;` fails

thorade: That was why the first travis build failed

mwetter:
I see.

PR #768, opened 2017-05-03
Inherit function icon

tbeu:


thorade:
I will also merge this into the `fix-inheritance branch`, thus updating PR #767

PR #772, opened 2017-05-05
Tbeu fix typos

thorade:
this replaces #771


thorade:
Typos in comments and docstrings,
travis passes

PR #773, opened 2017-05-05
Issue770 flow machine dp system integrated

### LINKED ISSUE #770: Mover with remote pressure tracking
When using static pressure reset controllers, the pressure at a remote point is controlled. To model this we currently need a control loop. Setting the remote pressure directly instead, would allow for some algebraic loops to be simplified. 
Therefore it would make sense to create a model that sets the remote pressure directly. I'll add such a model.

Mathadon:
@mwetter this integrates the functionality that we discussed yesterday. I decided to integrate this as an option in `FlowMachine_dp` instead of making a new mode, which can be confusing for the user. An alternative implementation that does use a separate model is on 9578913 . 
This closes #770 
I did not update the user guide since I put the documentation for the new parameters into the `FlowMachine_dp` model documentation. Is that ok?


Mathadon:
@mwetter before you merge this, I just thought of something:
Should `dp_actual` output the 'actual' pressure difference or the value corresponding to `dp_in`. These two values are not the same in the new model and I suppose the former choice would be appropriate?

Mathadon:
@mwetter this is ready to be merged if you agree with the changed connection of `dp_actual`.

mwetter:
@Mathadon `dp_actual` should be the actual pressure difference over the fan.

mwetter:
@Mathadon I added a list of points in the issue that need to be addressed before merging.

Mathadon:
@mwetter I addressed the comments!

mwetter:
@marcusfuchs Can you please review this PR? Thanks.

IBPSA/Fluid/Movers/BaseClasses/IdealSource.mo
```diff
@@ -4,8 +4,11 @@ model IdealSource
   extends IBPSA.Fluid.Interfaces.PartialTwoPortTransport(show_T=false);
 
   // Quantity to control
-  parameter Boolean control_m_flow "= false to control dp instead of m_flow"
+  parameter Boolean control_m_flow "= false to m_flow"
     annotation(Evaluate = true);
+  parameter Boolean control_dp = not control_m_flow "= false to control dp"
```
marcusfuchs: @Mathadon I think the comments for `control_m_flow` and `control_dp` are not easy to understand for the user. I would expect that `control_dp` should be `true` to control `dp`, not `false`. And `"= false to m_flow"` could use a verb.

mwetter: @marcusfuchs I improved the documentation. Can you please merge if all tests succeed.

mwetter:
@marcusfuchs Can you please merge if you are OK with the revisions.

marcusfuchs:
@mwetter Perfect, I'll merge.

PR #776, opened 2017-05-08
Updated run unittests

### LINKED ISSUE #774: runUnitTests.py
@mwetter and others: Is there any reason why the `runUnitTests.py` we have on the IBPSA repo differs from the one on the Buildings library repo? Should we update the one on the IBPSA repo to be the same as the one on Buildings? Otherwise the one on IBPSA won't be checking for instance the validity of mo/mos files?

tsnouidui:
This closes #774.@mathadon: Could you please review and merge? It is the same file as on the Buildings branch.

### DIFF
--- a/bin/runUnitTests.py
+++ b/bin/runUnitTests.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 #######################################################
 # Script that runs all unit tests or, optionally,
-# only checks the html syntax.
+# only checks the html syntax or the validity of
+# the simulation parameters of the models
 #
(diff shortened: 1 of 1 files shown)


Mathadon:
@tsnouidui I'm actually not that familiar with the details of buildingspy. So I couldn't tell if this makes sense :p does this need a 'serious' review?

tsnouidui:
@mathadon no problem. I will let @mwetter review. It should be all fine since it is a copy of the one used on Buildings.

PR #779, opened 2017-05-11
Issue763 prescribed outlet state

### LINKED ISSUE #763: Allow XSet as an input for PrescribedOutletState
The model `PrescribedOutletState` only takes as an input `TSet`. We should also allow `XSet` in order to model a humidifier that takes `XSet` as an input. The section that computes `dhSetAct` should be moved to a model-internal function so that it can be used to compute the raise in enthalpy and the raise in mass fraction. (This was not done in https://github.com/ibpsa/modelica-ibpsa/issues/39 as there was no need yet.)

This change will required using `PartialTwoPort` as the base class.

Parameters `use_{T,X}Set = true` could be used to enable or disable a mode and its associated function calls. As by default a new input connector will be present, the change is not backward compatible.

Note that by adding a capacity limit, this will allow something similar to a setpoint manager in EnergyPlus for temperature and humidity (see http://bigladdersoftware.com/epx/docs/8-0/engineering-reference/page-100.html). For example if we set `Q_flow_maxHeat=Modelica.Constants.inf` (the default) and `Q_flow_maxCool = 0`, we could write `h_out = smoothMax(h_in, hSet)` if the model is steady-state.

mwetter:
This closes #763 and is not backward compatible.

We would need one reviewer (any volunteer?), and all library developers to agree or raise their concerns within 2 weeks.

As this change also includes two new humidifier models, I moved `IBPSA.Fluid.MassExchangers.Humidifier_u` to `IBPSA.Fluid.Humidifiers.Humidifier_u`.

Below is the snippet of the conversion script that we used for this change.
```
convertClass("Buildings.Fluid.Interfaces.PrescribedOutletState",
             "Buildings.Fluid.Interfaces.PrescribedOutlet");


convertClass("Buildings.Fluid.HeatExchangers.HeaterCooler_T",
             "Buildings.Fluid.HeatExchangers.PrescribedOutlet");

convertModifiers("Buildings.Fluid.HeatExchangers.HeaterCooler_T",
                 {"m_flow_start"}, fill("", 0), true);
convertElement("Buildings.Fluid.HeatExchangers.HeaterCooler_T",
               "Q_flow_maxCool", "QMin_flow");
convertElement("Buildings.Fluid.HeatExchangers.HeaterCooler_T",
               "Q_flow_maxHeat", "QMax_flow");

convertClass("IBPSA.Fluid.MassExchangers.Humidifier_u",
             "IBPSA.Fluid.Humidifiers.Humidifier_u");


convertModifiers("Buildings.Fluid.MassExchangers.Humidifier_u",
                 {"use_T_in"}, fill("",0), true);
convertModifiers("Buildings.Fluid.MassExchangers.Humidifier_u",
                 {"T"}, fill("",0), true);

```


mlauster:
Which models are concerned by the non-backwards compatibility? I think it is worthwhile having that non-backwards compatibility to gain new modelling possibilities, I think they shouldn't be a problem for `AixLib`,  But I will take that into the group.

mwetter:
@mlauster To see which models are affected, it is easiest to look at the conversion script:
```
convertClass("Buildings.Fluid.Interfaces.PrescribedOutletState",
             "Buildings.Fluid.Interfaces.PrescribedOutlet");
```
means the name of this class changes.

```
convertModifiers("Buildings.Fluid.HeatExchangers.HeaterCooler_T",
                 {"m_flow_start"}, fill("", 0), true);
```
means that this parameter no longer exists.

```
convertElement("Buildings.Fluid.HeatExchangers.HeaterCooler_T",
               "Q_flow_maxCool", "QMin_flow");
```
means that this parameters was renamed to `QMin_flow`, etc.

See also the Dymola manual version 2 for more information about the conversion commands.

mwetter:
@mlauster @thorade @Mathadon : The two week review period for non-backward compatible changes is over, but I have only heard a tentative OK from Moritz. Please raise any objections by tomorrow or let me know if you need more time than that.

Mathadon:
For us this should be ok!

Issue #783, opened 2017-05-16
from_dp examples


Mathadon:
We should have some more simple examples that clearly demonstrate what can be done using from_dp.

mwetter:
Closed as this is on the master branch.

Issue #782, opened 2017-05-16
revise flow functions for easier analytic manipulation


Mathadon:
Variable `k` could be eliminated from the flow functions, which may lead to smaller algebraic loops. This issue aims to figure out how large the impact of such a revision would be.

Mathadon:
Some analysis from what I remember from a few months ago:
I did some tests and, although this works, the problem seems to be that the flow functions must have exactly the same arguments: dp_turbulent or m_flow_turbulent (depending on the value of `from_dp`). The problem is that we use m_flow_turbulent as an argument and we compute dp_turbulent internally. So unless we want to duplicate the computation of dp_turbulent everywhere, this won't be practical. Finally, the alias variable elimination of the tool should work very well for this.

mwetter:
Closed for now unless someone starts working on it again.

PR #784, opened 2017-05-16
added some example models where the influence of from_dp on translati…

Mathadon:
…on statistics is demonstrated

this is for #783 

@mwetter I suppose we want to put this on the master? This can also be useful to benchmark #782


mwetter:
@Mathadon 
Can you please make the unit tests work before I review and also add info to package.mo
Looking at the (broken) models, I also wonder if the test can be further simplified, e.g., are so many parallel branches needed? Wouldn't it be easier for the tool developer (and us) to have something like ParallelFrom_dp_ideal, ParallelFrom_dp_true, ParallelFrom_dp_true where ideal has the ideal (manually tuned) settings for from_dp, and the others have either true or false everywhere. This way, we could see if tools generate the same statistics as the *_ideal case.

Mathadon:
@mwetter I accidentally used some models from my own package, which caused the models to fail. So the failed tests were not related to the number of parallel/series branches. They aren't that complex.

Clearly there are different ways of doing setting up these examples. However, I reasoned as follows: each example essentially contains the same flow network, but with different boundary conditions or `from_dp` settings. Ideally, tools can find the same analytic solution to these problems regardless of what variables are prescribed and what the values of `from_dp` are. So it makes sense to put all of them in the same example such that it's easy to compare their statistics. Since I added the different from_dp combinations in each example, the ideal solution is also contained within the example.

I'm also not sure whether it makes sense to create *_ideal cases since for more complex examples we might not even know what the ideal configuration is and it probably depends on what simplifications the tools support.

mwetter:
But this way, we don't know _where_ the nonlinear equations come from. It could be from any of the 4 parallel examples. And it is also not clear what the expected minimum number of nonlinear equations is, e.g., what is the best case that a tool should strife for? Do we need 10 series, or would 2 suffice? For tool developers to be able to analyze the problem, we should make it as simple as possible.

Mathadon:
@mwetter  this is ready to review

PR #791, opened 2017-05-31
Issue787 fix substance x

JayHuLBL:
Fixed the incorrect substance X.


mwetter:
@mlauster Can you please review and merge if you approve so that we have a non-LBNL reviewer in the merge process.

PR #795, opened 2017-06-05
Tighten tolerance for JModelica

### LINKED ISSUE #794: Tolerance for JModelica verification
This is to tighten the tolerance of some models so they pass JModelica verification.

tsnouidui:
The tolerance needs to be tightened for JModelica verification. This closes #794.


tsnouidui:
@mwetter,  Does travis include the test to check if .mos and .mo files are well formed? 

Running 

`../bin/runUnitTests.py --validate-experiment-setup`

yields to 

`ValueError: 
Found mo file=./Fluid/Examples/Performance/PressureDrop/SeriesParallelDpOptimised.mo without experiment annotation.`

There seems to be a number of models without experiment annotations.
It will be go to set-up travis to automatically catch that so we don't have to do it manually.

@Mathadon Could you please fix the model(s) which seems to be one of your commits? The `mo` should essentially reflect what is in the `mos` file since JModelica doesn't use the `mos` but rather use the information encoded in the `experiment` annotation of the `mo` file.

mwetter:
@tsnouidui 
Please set this test up for IBPSA, it should already be included in Buildings.

tsnouidui:
@mwetter Any instructions on how to do this? Do I only need to replicate 

`https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Resources/Scripts/travis/Makefile` with `Buildings` changed to `IBPSA`?

tsnouidui:
I submitted a pull request to include the test for the experiment setup. Since @Mathadon is away, and the models he committed will cause travis to fail once the pull request is integrated in the master, I will try to fix them.

PR #801, opened 2017-06-07
Issue800 three way valves

### LINKED ISSUE #800: ThreeWay valves' documentation and code need to be corrected
This issue is to address https://github.com/lbl-srg/modelica-buildings/issues/795

mwetter:
This closes #800.

I also revised the test for the two way valves and moved it to the `Validation` package.


mlauster:
@mwetter , it's hard to evaluate the validation cases, but I assume they do what they are intended for, I will merge that!

PR #803, opened 2017-06-08
Issue802 mover icons

### LINKED ISSUE #802: fix mover icons
The real output is very close to the fluid port which often yields the gui the pick the wrong connector. Also, there is text over the control input signal. This issue is to correct these graphical problems.

mwetter:
This closes #802.
Changes were only made to the annotation to move connectors and make sure all are labelled correctly.


marcusfuchs:
In my opinion, these are helpful changes.

PR #804, opened 2017-06-19
Issue698 carnot dt

### LINKED ISSUE #698: Add approach temperature to Carnot.mo
This issue is to integrate https://github.com/lbl-srg/modelica-buildings/issues/653 as otherwise, the COP of the heat pump is unreasonably high if there is a small temperature lift. Adding an approach temperature is also justified as the Carnot efficiency is for the refrigerant temperatures rather than the water temperatures.

mwetter:
This closes #698.


mwetter:
@Mathadon The changes look good to Felix but he did not simulate the new model.
Could you please review and merge if it is OK.

IBPSA/Fluid/Chillers/BaseClasses/Carnot.mo
```diff
@@ -50,6 +50,14 @@ partial model Carnot
     "Pressure difference over evaporator"
     annotation (Dialog(group="Nominal condition"));
 
+  parameter Modelica.SIunits.TemperatureDifference TAppCon_nominal(min=0) = if cp1_default < 1500 then 5 else 2
```
Mathadon: It would make sense to let the name reflect that the parameter is a temperature _difference_?
E.g. use `dT` or `TDif`

IBPSA/Fluid/Chillers/BaseClasses/Carnot.mo
```diff
@@ -50,6 +50,14 @@ partial model Carnot
...
+    "Temperature difference between refrigerant and working fluid outlet in condenser"
+    annotation (Dialog(group="Efficiency"));
+
+  parameter Modelica.SIunits.TemperatureDifference TAppEva_nominal(min=0) = if cp2_default < 1500 then 5 else 2
+    "Temperature difference between refrigerant and working fluid outlet in evaporator"
```
Mathadon: It would make sense to let the name reflect that the parameter is a temperature difference?

mwetter:
@Mathadon I suggest to leave it as the approach temperature is always a difference, and to keep the variable name shorter. The approach temperature is the difference between the leaving process and the entering service fluid.

PR #806, opened 2017-06-29
Issue18 move

mwetter:


mwetter:
I will merge it to the development branch even though there are many packages that should not have changes, such as `Fluid.Sensors`.

PR #811, opened 2017-07-31
pr810 eta

thorade:
this replaces #810, sorry for two PRs

Using MSL SIunits where possible instead of defining new units should enhance readability.

Note: The MSL definition has a min value defined
```modelica
type Efficiency = Real (
      final quantity="Efficiency",
      final unit="1",
      min=0);
```

If that was the reason why the MSL unit was not used one could also change the min value at instantiation:
```modelica
Modelica.SIunits.Efficiency etaHyd(min=-1) = eff.etaHyd "Hydraulic efficiency";
```


mwetter:
@thorade We used to have limits set as in `Modelica.SIunits.Efficiency` but JModelica violated them slightly due to numerical noise from the iterative solvers, causing the JModelica verification to indicate a failed test. This is why we removed them. I suggest we close this pull request without merging.

PR #824, opened 2017-09-20
Issue805 pipe

bramvdh91:
@mwetter I would like to ask you to review the new single pipe model. Tried to adhere to the Style Guide as far as this was possible without making the unittests fail.

However, when I am running the unittests of other folders than the one I added, I get the following change: 
```
*** Warning: ConservationEquation.mat: Translation statistics for initialization changed for nonlinear, but results are unchanged.
 Old = 1
 New = 0, 1
```
I checked that this is also the case when I run the same in the master branch. Does this indicate a problem?


bramvdh91:
I'm still struggling with the unit tests. Those for `IBPSA.Fluid.PlugFlowPipes` are working; when I run all unit tests of `IBPSA.Fluid` I get a lot of changed reference results (although I did not change anything there); the unit tests for all of the library get stuck in some other folder.

bramvdh91:
I've cleaned up the html documentation in order to fix part of the travis checks. Still the unit tests for the added models produce warnings due to different results; I'm afraid I'm not able to solve this on my computer as I seem to get different reference results for a lot of unit tests (running 64 bit Dymola 2018 on Win10).

thorade:
The results depend on the Dymola version, but also OS and compiler

mwetter:
@bramvdh91 : I updated the reference results for Dymola 2017FD01 on Linux, which is what travis is using. I will further review today and tomorrow.

thorade:
@bramvdh91 Just a trivial sidenote: You can also do multiple commits locally, and then push those commits all at once to github. That would also trigger just one single Travis build, for the last commit.

bramvdh91:
@thorade I will pay attention to this in the future, sorry for the inconvenience.
@mwetter I tried to revise the code according to the suggestions in your mail. Okay for me to move `IBPSA.Fluid.PlugFlowPipes.PlugFlowPipe` inside `..FixedResistances`, although I would argue that these models implement more than just the hydraulic resistance. Maybe it could help to rename FixedResistances to something that reflects both hydraulic and thermal losses in a transport pipe?

bramvdh91:
Checks failed again because the validation results have changed slightly, but I'm not sure what's the cause.

bramvdh91:
@mwetter regarding c8efaaa: I'm wondering how much extra memory is required because of these extra `spatialDistribution()` (if any?). 
This also made me think that perhaps it is simpler to also propagate the entrance time like this instead of in the separate `TimeDelay` component. That way we would avoid the extra mass flow rate sensor and the connections would become a bit simpler.

mwetter:
@bramvdh91 : There is no extra overhead. If a medium has nC=0 or nXi=0 (e.g., water) than these equations are removed. The `spatialDistribution` operator needs to be vectorized as I did, rather than using a vector of `C_outflow` as arguments.

For efficiency, the propagation for reverse flow could conditionally be removed if `allowFlowReversal=false`.

mwetter:
@bramvdh91 : I am almost done with the revisions but have these two questions/requests:

1. Should we delete `PlugFlowCore` and use these models directly in `PlugFlowPipe`? In my view, there is no reason to have this additional layer that only separates the mixing volume from the other parts of the model.

2.  Why is `MSLATI*Nodes` present? Shouldn't there at least be the same model but with the new pipe? What is the point of this model if it does not test a pipe that we use? Please remove it, or if it stays in the library, please

- [ ] Explain in the info section what its purpose is.
- [ ] Correct the wrong hyperlinks in the info section.
- [ ] Clean up the connections in these models as the connections are quite messy. (You may be able to clean up one only, and then copy all the "connect" statements to the other model.)

bramvdh91:
@mwetter 

- ad 1: These separate levels were introduced to allow compatibility with @marcusfuchs' uesmodels code; here the fluid volumes are added at pipe junctions (or not, depending on the chosen settings). Therefore, we needed a "core" model without added volumes.
- ad 2: I see your point that it is not logical to add these examples in the library as such. However, in the EnConMan paper we presented a comparison between our plug flow model and the MSL with different discretisations.  We could provide these two examples separately as a resource to the paper, or we could just leave them out altogether.

mwetter:
@bramvdh91 : Regarding 1, that is fine then. It would be good to make a remark under an `<code>Implementation</code>` section in the info section.

Regarding 2, I suggest to either leave them if you also provide the implementation with the new pipe model so that a comparison is possible, or remove them.

bramvdh91:
@mwetter 
Regarding 2: Perhaps I'm missing your point here, but the plug flow implementation is given in ``IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.PlugFlowAIT``. Only one such example is needed because there is no discretization.

mwetter:
@bramvdh91 : I now realized that the 3 models are identical except for the pipe. I added documentation in https://github.com/ibpsa/modelica-ibpsa/pull/824/commits/edafb54af241155b4a29b2b7bab8f73aed55c73f
Hopefully this is now more obvious to users.

(Please also use shorter lines so that diffs can be seen, that the whole line can be seen in github's display, and that editing is easier on editors that don't automatically wrap lines.)

mwetter:
@bramvdh91 : can you please address the comment from 15 days ago so that we can merge this after one final review by one of the main developers (because we did some many revisions, it would be good to have another review)

bramvdh91:
@mwetter Sorry for that, I was not aware that I needed to do more changes. 
I wrapped the long documentation lines and expanded documentation where I thought it was still needed. Was there another comment you were referring to?

mwetter:
@bramvdh91 : Thanks for the changes. From my point of view, this is ready to merge.

@marcusfuchs : Can you please have a last review and then merge as part of my review, I did substantial refactoring.

marcusfuchs:
@mwetter Yes, I can do the review and merge. I'll start with fixing the documentation causing the tests to fail.

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowAIT.mo
```diff
@@ -0,0 +1,392 @@
...
+    experiment(
+      StopTime=603900,
+      Interval=900,
+      Tolerance=1e-006),
+    Documentation(info="<html>
+<p>
+The example contains
+experimental data from a real district heating network.
+</p>
+<p>The pipes' temperatures are not initialized. Therfore, results of
```
marcusfuchs: Typo in `Therfore`

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/MSLAIT.mo
```diff
@@ -0,0 +1,491 @@
...
+          {78,62},{74,62}}, color={0,0,127}));
+  connect(col1.port_b, prescribedTemperature.port) annotation (Line(points={{
+          112,46},{214,46},{214,-142},{94,-142}}, color={191,0,0}));
+  annotation (
+    Diagram(coordinateSystem(preserveAspectRatio=false, extent={{-160,-160},{
+            220,200}})),
+    experiment(
+      StopTime=603900,
+      Tolerance=1e-006,
+      __Dymola_Algorithm="Dassl"),
```
marcusfuchs: Of the 3 models for comparison, only this one explicitly sets a solver choice for Dymola. Should we have this in either none or all of the models?

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowULg.mo
```diff
@@ -0,0 +1,213 @@
+﻿within IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes;
+model PlugFlowULg "Validation against data from Université de Liège"
+  extends Modelica.Icons.Example;
+  // R=((1/(2*pipe.kIns)*log((0.0603/2+pipe.dIns)/(0.0603/2)))+1/(5*(0.0603+2*pipe.dIns)))/Modelica.Constants.pi
```
marcusfuchs: Is this comment still needed?

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowULg.mo
```diff
@@ -0,0 +1,213 @@
...
+    "Initial temperature at pipe outlet";
+  parameter Modelica.SIunits.SpecificHeatCapacity cp_default=
+      Medium.specificHeatCapacityCp(state=sta_default)
+    "Heat capacity of medium";
+  parameter Medium.ThermodynamicState sta_default=Medium.setState_pTX(
+      T=Medium.T_default,
+      p=Medium.p_default,
+      X=Medium.X_default) "Default medium state";
+
+  Fluid.Sources.MassFlowSource_T WaterCityNetwork(
```
marcusfuchs: Missing docstrings

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowULg.mo
```diff
@@ -0,0 +1,213 @@
...
+  annotation (
+    Documentation(info="<html>
+<p>
+The example contains
+experimental data from a real district heating network.
+</p>
+<p>
+This model compares the results with the original Modelica Standard Library pipes.
+</p>
+<p>The pipes' temperatures are not initialized. Therfore, results of
```
marcusfuchs: Typo in `Therfore`

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowULg.mo
```diff
@@ -0,0 +1,213 @@
...
+</p>
+<h4>Test bench schematic</h4>
+<p><img alt=\"Schematic of test rig at ULg\"
+src=\"modelica://IBPSA/Resources/Images/Experimental/ULgTestBench.png\"/> </p>
+<h4>Calibration</h4>
+<p>
+There are some uncertainties about the heat loss coefficient between pipe and
+surrounding air as well as regarding the heat conductivity of the insulation
+material.
+With the <a href=\"modelica://IBPSA.Experimental.Pipe.Data.PipeDataULg150801\">
```
marcusfuchs: This is still an old hyperlink

IBPSA/Fluid/Interfaces/PartialFourPortParallel.mo
```diff
@@ -0,0 +1,81 @@
...
+    "Fluid connector b2 (positive design flow direction is from port_a2 to port_b2)"
+    annotation (Placement(transformation(extent={{110,-70},{90,-50}})));
+
+  annotation (
+    preferredView="info",
+    Documentation(info="<html>
+<p>
+This model defines an interface for components with four ports
+in which flows occur in parallel.
+The parametera <code>allowFlowReversal1</code> and <code>allowFlowReversal2</code>
```
marcusfuchs: Typo in `parametera`

IBPSA/Fluid/Interfaces/PartialFourPortParallel.mo
```diff
@@ -0,0 +1,81 @@
...
+may be used by models that extend this model to treat flow reversal.
+</p>
+<p>
+This model is identical to
+<a href=\"modelica://Modelica.Fluid.Interfaces.PartialTwoPort\">Modelica.Fluid.Interfaces.PartialTwoPort</a>,
+except for the fowllowing: </p>
```
marcusfuchs: Typo in `fowllowing`

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/Data/FlowRateGeneric.mo
```diff
@@ -0,0 +1,50 @@
+within IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.Data;
```
marcusfuchs: It seems like this dataset is never used in the current codebase. Should we include it in the merge?

IBPSA/Fluid/FixedResistances/BaseClasses/PlugFlowCore.mo
```diff
@@ -0,0 +1,263 @@
...
+
+  IBPSA.Fluid.FixedResistances.BaseClasses.PlugFlow del(
+    redeclare final package Medium = Medium,
+    final m_flow_small=m_flow_small,
+    final dh=dh,
+    final length=length,
+    final allowFlowReversal=allowFlowReversal,
+    final T_start_in=T_start_in,
+    final T_start_out=T_start_out)
+    "Model for temperature wave propagation (fixme: could this be part of the del instance?)"
```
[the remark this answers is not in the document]
bramvdh91: Not sure either, I think this refers to splitting out the hydraulic resistance from the former PlugFlowAdiabatic (current PlugFlow model in BaseClasses). I have removed it.

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/MSLAIT.mo
```diff
@@ -0,0 +1,491 @@
+within IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes;
+model MSLAIT
+  "Validation pipe against data from Austrian Institute of Technology with standard library components"
+  extends Modelica.Icons.Example;
+
+  /*TODO: change nNodes for pipes. For fair comparison, n should be adapted to
```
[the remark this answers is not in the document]
bramvdh91: The comparison between one element per meter and 2 elements per pipe was made, but apart from a reduced calculation time, no big differences were found. Removing `TODO`.

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowAIT.mo
```diff
@@ -0,0 +1,392 @@
...
+  parameter Boolean allowFlowReversal=true
+    "= true to allow flow reversal, false restricts to design direction (port_a -> port_b)";
+  parameter Modelica.SIunits.MassFlowRate m_flow_nominal=1
+    "Nominal mass flow rate, used for regularization near zero flow";
+  parameter Modelica.SIunits.Time tauHeaTra=6500
+    "Time constant for heat transfer, default 20 minutes";
+
+  parameter Modelica.SIunits.Length thickness=0.0032 "Pipe wall thickness";
+
+  Fluid.Sources.MassFlowSource_T Point1(
```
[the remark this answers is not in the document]
bramvdh91: Which docstrings are you referring to? Comparing to the parallel examples, nothing seems to be missing.

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/MSLAIT.mo
bramvdh91: Added this in the other two models as well.

bramvdh91:
Thanks for the review, @marcusfuchs!

I addressed the remaining review comments.

marcusfuchs:
Perfect, thanks @bramvdh91 - I'm really happy to merge this!

PR #828, opened 2017-09-30
added polynomial valve model

### LINKED ISSUE #821: More detailed equal percentage valve
On page 5 of https://www.downloads.siemens.com/download-center/Download.aspx?pos=download&fct=getasset&id1=A6V10416372 . The flow characteristic for a specific Siemens valve is outlined. This equal-percentage three-way valve consists of a linear valve and an equal-percentage valve where the equal-percentage valve also has a linear part for small valve openings. I'm seeing some model mismatch in my validations that is probably caused by this. It would be good if we could extend existing models to support this.
Do we add a second equal percentage valve, or do we provide an option in the existing one? At the same time we could consider moving some of the valve parameters (rangeability, valve leakage) to the advanced section.

Mathadon:
this closes #821


IBPSA/Fluid/Actuators/Valves/TwoWayPolynomial.mo
```diff
@@ -0,0 +1,43 @@
...
+
+  parameter Real[:] c "Polynomial coefficients, starting with fixed offset";
+
+protected
+  Real pol_y = sum(c.*{y_actual^i for i in 0:size(c,1)-1}) "Polynomial of valve control signal";
+
+equation
+  // We provide some slack on these bounds to avoid the assert failing when
+  // small integration errors are made, e.g. in the filter.
+  assert((pol_y>=-0.2 and pol_y <=1.2),
```
mwetter: As `y_actual` can vary between 0 and 1, you better add a test in the initial equation section for these limits.
Please also add a test of the form `pol_y(y_actual[i]) < pol_y(y_actual[i+1])` for a sequence `y_actual[i] < y_actual[i+1]` to test that the curve is increasing (or better for low size(c,1), add an analytical test.

IBPSA/Fluid/Actuators/Valves/TwoWayPolynomial.mo
```diff
@@ -0,0 +1,43 @@
...
+annotation (
+defaultComponentName="val",
+Documentation(info="<html>
+<p>
+Two way valve with polynomial opening characteristic.
+The polynomial coefficients are defined using parameter <code>c</code>.
+The elements of c are coefficients for increasing powers of y,
+starting with the power 0, which corresponds to a fixed offset.
+</p>
+<p>
```
mwetter: Please add why a user would choose this valve to avoid users being confused what to select. Maybe this should be in the user guide as well?

IBPSA/Fluid/Actuators/Valves/TwoWayPolynomial.mo
```diff
@@ -0,0 +1,43 @@
+within IBPSA.Fluid.Actuators.Valves;
+model TwoWayPolynomial "Two way valve with polynomial characteristic"
+  extends IBPSA.Fluid.Actuators.BaseClasses.PartialTwoWayValveKv(
+    phi=l + pol_y*(1 - l));
+
+  parameter Real[:] c "Polynomial coefficients, starting with fixed offset";
```
mwetter: Do you have an example setting for `c` for a few valves? This should better be in a data record so that a library of valve coefficients, together with their leakage and kV_s values, can be provided to the user.

Mathadon:
Thanks for the feedback Michael! I'll need some time to process this though.

IBPSA/Fluid/Actuators/Valves/TwoWayPolynomial.mo
Mathadon: I currently have only one example.

Using records makes sense, but not if we only do it for the polynomial model? Then all valves should be refactored. Unless you want to use the polynomial model as a 'one size(/model) fits all' model which is used to build up a database of valve implementations, including linear valves etc.
In this case I'd propose to make a new model `TwoWay` or `TwoWayRecord` that extends `TwoWayPolynomial` and that declares the record and propagates relevant parameters.

IBPSA/Fluid/Actuators/Valves/TwoWayPolynomial.mo
Mathadon: I extended the description. There is currently no overview of the two/three way valve models in the user guide, so there was no obvious location where to mention the new valve model. Is this something we'd like to add?

IBPSA/Fluid/Actuators/Valves/TwoWayPolynomial.mo
Mathadon: @mwetter : please see my reply :)

IBPSA/Fluid/Actuators/Valves/TwoWayPolynomial.mo
Mathadon: @mwetter : please see my reply :)

mwetter:
@Mathadon 
The outer `i` has no effect as there is an inner `i`:
```
  for i in 0:nP loop
    phi_test:=sum(c .* {(i/nP)^i for i in 0:size(c, 1) - 1});
    assert(phi_test-phi_test_prev>0, "The provided valve polynomial coefficients 
    do not lead to a strictly increasing characteristic. This is not allowed.");
  end for;
```
Also, why not just testing the derivative at all support 100 points? This would save you a few multiplication and subtractions, and it could be done in an `initial equation` section.

We can wait with using records as this would be a larger refactoring.

Mathadon:
@mwetter I think I fixed all remarks :)

PR #843, opened 2017-10-25
Issue839 initialize p

### LINKED ISSUE #839: Change initialize_p from a constant to a parameter
In some models in `Buildings` that have staggered volumes, `initialize_p` is computed based on an expression that has variability parameter. This need to be corrected for models to work with an update in OpenModelica (see https://github.com/lbl-srg/modelica-buildings/issues/1013)

mwetter:
This closes #839 

There is a long discussion in https://github.com/lbl-srg/modelica-buildings/issues/1013

In summary, for models that use staggered volumes, such as `Buildings.Fluid.HeatExchangers.DryCoilDiscretized` the pressure equation need to be set differently for the first volume. This requires code such as
https://github.com/lbl-srg/modelica-buildings/blob/issue1013_initialize_p/Buildings/Fluid/HeatExchangers/DryCoilDiscretized.mo#L107

The problem is that the construct that is used here, e.g.,
``` modelica
model Test
  parameter Integer nEle = 2
    annotation(Evaluate=true);
  M m[nEle](
    b=cat(1, {true}, fill(false, nEle-1)));
  model M
    constant Boolean b;
  end M;
end Test;
```
where `M` is the `ConservationEquation` model. But because `nEle` is a parameter, this is not valid, and the latest version of `OpenModelica` refuses to translate it.

I see two options that I developed with input from @sjoelund and @tbeu :

1. Make `nEle` a parameter. This is not satisfactory as users need to change this to change the level of discretization and to specify how many heat exchanger registers are present.
2. Make `b` (e.g., `initialize_p`) a parameter. But then users will see it in the dialog window, and frankly I don't think more than a few developers know what to do with this parameter. The parameter can only be hidden if made `final`.
3. Change the Modelica specification to allow hiding parameters, or displaying constants, in a GUI, but we need a solution until such a change would be implemented.

I selected option 2 and made it `final` in `MixingVolumes.{MixingVolume,MixingVolumeMoistAir}` so that users do not see it in a GUI, but this required two new base classes in `MixingVolumes.BaseClasses` that have it *not* set to `final`. These base classes are then used in `FourPortHeatMassExchanger` so that we can build staggered volumes and initialize the pressure equations correctly.

As this adds some complexity for developers (but preserves the list of parameters for users) I think it would be good if @Mathadon , @mlauster and @thorade  can approve or raise their concerns (and if so, provide an alternate solution) before merging.
Test model are in `{IBPSA.Fluid.MixingVolumes.BaseClasses.Validation}` and a related pull request that contains the heat exchanger models is https://github.com/lbl-srg/modelica-buildings/pull/1017


Mathadon:
@mwetter option 2 makes sense. In my opinion it's not required to set `final` as long as the parameter is in the 'Advanced' tab. Or we could make a new tab 'Developer' for this purpose and add to the user guide that these parameters should not be touched..?

mwetter:
@Mathadon I was thinking about a `Developer` tab but then used `final` (and added the new classes) to have fewer tabs that may confuse the user. I would rather not put it on the `Advanced` tab as the `Advanced` tab also has other parameters that advanced users, but not developers, can set such as the `from_dp` and `show_T`.

Mathadon:
How about a developer group in the `Advanced` tab?

mwetter:
@Mathadon That could be possible too. The main question is whether we want to show `initialize_p`. However, I doubt anyone remembers how to set this or would need to change this value, which is why I opted for now not to show it.

Mathadon:
Ok, but for some other parameters it may be good to keep this in mind!

mwetter:
@mlauster  and @thorade : do you have any feedback or shall I go ahead and merge this?

thorade:
OK from Berlin.

mwetter:
@mlauster : do you plan to review or shall I go ahead and merge it?

mlauster:
I followed the discussion partly, no concerns from Aachen side. Go ahead and merge!

Issue #847, opened 2017-11-03
Algorithm section in IBPSA.Utilities.Psychrometrics.X_pTphi


Mathadon:
`IBPSA.Utilities.Psychrometrics.X_pTphi` has an algorithm section that can be written using an equation section.

PR #848, opened 2017-11-03
converted algorithm section into equation section in IBPSA/Utilities/…

Mathadon:
…Psychrometrics/X_pTphi.mo

for #847


mwetter:

PR #854, opened 2017-11-03
Issue825 verification

### LINKED ISSUE #825: set up examples so that they have flow rates above m_flow_turbulent
The example `ResistanceVolumeFlowReversal` has flow rates through the parallel resistances that are well below `m_flow_turbulent`. Hence, depending on whether a tool solves m_flow = f(dp) or dp=g(m_flow), mass flow rates in the parallel branches can differ.

This issue is to fix the model so that the flow rate is at or above `m_flow_turbulent`, in which case m_flow = f(dp) and dp=g(m_flow) gives identical results. This is needed as different tools may solve the equation differently, and hence the verification would fail. For example, Dymola 2018FD01 beta2 fails the tests because of this.

A similar problem may apply to `Simplified1`, which has these differences, produced with dymola 2018FD01 beta2:
image

@Mathadon : Can you please change `Simplified1` so that its flow rates are above `m_flow_turbulent`. I changed this for `ResistanceVolumeFlowReversal`

mwetter:
This closes #825


mwetter:
This is ready to merge when all tests pass

mwetter:
`Fluid.Examples.FlowSystem.Simplified1` fails to converge with Dymola 2017 FD01, but works with Dymola 2018 FD01 beta3. Since Dymola 2018FD01 has been released, I suggest we merge this after updating to the new version.

Issue #842, opened 2017-10-24
ReaderTMY3 does not work for data going over into the next year


AnaConstantin:
Hello everyone,

I have my own weather data, which I have adapted to a TMY3 format, in order to use the ReaderTMY3. The data covers the period of June 2016 to October 2017. The time stamp I used in the TMY3 format sets 1.01.2016 00:00 to 0s. However I get an error at 01.01.2017 00:00 and the simulation stops. Is there a way around this, as depending on my application, I might have data stretching over more than a year anyway? Or am not setting some parameter correctly?
Otherwise I'm happy to assist in extending the ReaderTMY3 to deal with such a situation.

mwetter:
The ReaderTMY3.mo file is simply reading a .mos file. The 2nd line of this file has the number of columns. Maybe you made a mistake in this entry?

AnaConstantin:
@mwetter thank you for your suggestion. Unfortunatelly it was not that. The description of the table in the second line is correct, otherwise the simulation wouldn't start.
To my mind the problem is in the  ConvertTime-Block (conTim). Once the time stamp jumps into the next year, the calendar time (calTime) used for the interpolation of the table values jumps from 31539600s to 0s and the interpolation of the table values produces values outside the set bounds which because of some assert statements lead to a break in the simulation. The way I understand it, the advantage of this component, is that it allows a repeat of the table with annual data, if the simulation should strech over one year, but it does not allow the reading of values in a table with a time stamp over 31539600s. If this is the case I might have an idea to solve it by using a Modelica.Blocks.Sources.CombiTimeTable.
However this is just speculation on my side. I'm happy to hear your thoughts. Thank you in advance for your answer.
I have attached a snippet of my weather table which exemplifies this problem. As GitHub does not support .mos format, I saved it in .txt format and has to be re-saved in .mos format to work. I have done the simulation using both Lsodar and Euler as solvers, using the newest version of the Modelica IBPSA Library and Dymola 2018. 
[ExampleFile_error.txt](https://github.com/ibpsa/modelica-ibpsa/files/1413482/ExampleFile_error.txt)

mwetter:
@AnaConstantin : You are right, the issue is caused by the `ConvertTime` block.
I made a validation case on branch `issue842_multiyear` but won't have time this week to implement a fix. I think what should be done is 

[...]

PR #856, opened 2017-11-15
Modifications for #842

AnaConstantin:
1. Added function to evaluate the time span of the weather data getTimeSpanWeatherData.mo.
2. Modified ReaderTMY3.mo to evaluate the time span and decide if the convert time block should be used or not.
3. Some minor modifications:
- changed longitude to latitude for error message in getLatitudeTMY3.mo
- added table name as parameter for ReaderTMY3.mo, in order not to have it hidden inside the model and address a more general approach to tables.

If review is passed, I can also update the documentation and the revision section of ReaderTMY3.mo

### DIFF
--- a/IBPSA/BoundaryConditions/WeatherData/BaseClasses/getLatitudeTMY3.mo
+++ b/IBPSA/BoundaryConditions/WeatherData/BaseClasses/getLatitudeTMY3.mo
@@ -14,7 +14,7 @@ algorithm
     IBPSA.BoundaryConditions.WeatherData.BaseClasses.getHeaderElementTMY3(
       filNam=filNam,
       start="#LOCATION",
-      name=  "longitude",
+      name = "latitude",
       position=7);
    (nexInd, lat) :=Modelica.Utilities.Strings.Advanced.scanReal(
     string=element,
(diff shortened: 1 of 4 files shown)


thorade:
I'll merge this as it is, so that the code is in the IBPSA repository, so that the tests can be run on it.
If you have more changes, you can just keep committing to your branch, and then open a new pull request from your issue branch into the IBPSA issue branch.

thorade:
See test results for branches here:
https://travis-ci.org/ibpsa/modelica-ibpsa/branches

thorade:
@AnaConstantin Could you have a look at why the tests are failing and try to fix it?

AnaConstantin:
Tried to fix them via #857.

thorade:
Thanks, the code will now be tested by Travis.
One very minor note: If you add the issue number #842 in the **text** of the pull request (instead of in the title) then github will generate nicer cross-references.

AnaConstantin:
Thanks for the tip. I was not aware of the fact that it does not link to the issue if the reference is in the title.
I'll check in one hour to see if the tests ran correctly this time.

thorade:
The only failure now is due to missing reference results for the new test.
Reference results (currently) have to be generated with BuildingsPY on Ubuntu 16.04 with Dymola 2017FD01. Do you know how to do that?

AnaConstantin:
Short answer: No, I do not know how to do it and have no access to a Linux machine. 
Long answer: you sure we need to do it? I suppose the example will be deleted if my (or another) solution is considered correct, since this is more of a bug issue, than an extend functionality issue, for which a new permanent example would make sense.

PR #861, opened 2017-11-21
Updated revision section

### LINKED ISSUE #860: Remove __Dymola_loadSelector in revision section
This is to replace `__Dymola_loadSelector` in the revision section, because the Buildings library release script searches for this string during its compliance checking.

mwetter:
This closes #860


thorade:
Not sure if it is best to mention the removed annotation in the revisions, because then these revision will always turn up when searching for the deprecated annotation.

mwetter:
@thorade : This is why I replaced the string `__Dymola_loadSelector` in the revision section.

thorade:
Yes of course, looks like I mixed up red and green...

PR #858, opened 2017-11-16
Issue842 multiyear

thorade:
This PR will close #842 once it is merged to the master,
it contains the code from @AnaConstantin from PR #856 and #857 and #859.

Opening the PR so that changes can easily be seen and discussed here,
Ana you can still add code to the branch.

@mwetter Should there be reference results for the new test, or was it only temporary?

PR #863, opened 2017-11-24
Example DecemberToJanuary.mo runs.

AnaConstantin:
Deleted the start values for the timeSpan parameter and just added a filename to ReaderTMY3.mo


thorade:
This is for #858

PR #871, opened 2017-12-07
ignore diagnostics and eventlog files

thorade:
[ci skip]
Does the CI skip command also work from PRs?

The files ignored are created when using the new debugging feature,
see https://github.com/UdK-VPT/BuildingSystems/issues/78#issuecomment-349312150


mwetter:
I believe [ci skip] will also skip the CI on the pull request.

Issue #869, opened 2017-12-06
Plugflow pipe fails to translate in JModelica


mwetter:
The plugflow pipe fails to translate in JModelica.

Mathadon:
What error message do you get?

mwetter:
The error is
```
In components:
    pip.heaLos_a
    pip.heaLos_b
  Calling function Medium.specificEnthalpy_pTX(): types of positional argument 3 and input X are not compatible
    type of 'inStream(port_a.Xi_outflow)' is Real[0]
    expected type is Real[1]
```
I am not convinced that JModelica properly resolves the medium dependencies, but the bug fix provided in the pull request works around it.

Mathadon:
Looks like a bug, indeed! Nice fix :)

PR #874, opened 2017-12-12
Issue872 jmodelica tests

### LINKED ISSUE #872: add unit tests for translation in JModelica
This branch is for adding unit tests for compilation with JModelica.

mwetter:
This closes #872 
As before, the unit tests return 0 if all is fine, or non-zero otherwise. The JModelica translation tests can be invoked with
```
../bin/runUnitTests.py -s IBPSA.Controls.Continuous.Validation -t jmodelica
Regression tests are only run for the following package:
  IBPSA.Controls.Continuous.Validation
Using 2 of 24 processors to run unit tests for jmodelica.
Number of models   : 254
          blocks   : 101
          functions: 68
Generated 2 regression tests.

Script that runs unit tests had 0 warnings and 0 errors.

See 'simulator.log' for details.
Unit tests completed successfully.

Execution time = 7.374 s
```
The file `simulator.log` also contains an entry with the compilation time.

Not yet implemented is testing the translation log for warnings, and also not yet implemented is a check for the simulation and verification of results. This will be added later.

I suggest to merge this already to avoid errors such as
https://github.com/ibpsa/modelica-ibpsa/issues/869

The time on travis increases from 26 to 38 minutes.


IBPSA/Resources/Scripts/travis/usr/local/bin/jm_ipython.sh
```diff
@@ -0,0 +1,51 @@
...
+sha_dir=`dirname ${cur_dir}`
+# If the current directory is part of the argument list,
+# replace it with . as the docker may have a different file structure
+arg_lis=`echo $@ | sed -e "s|${cur_dir}|.|g"`
+
+docker run \
+	--detach=false \
+    ${MOD_MOUNT} \
+    -v ${sha_dir}:/mnt/shared \
+	-e DISPLAY=${DISPLAY} \
```
Mathadon: do you need this line for JModelica?

mwetter: @Mathadon For travis and JModelica, the DISPLAY is probably not needed. But if present, the same script can be used to run JModelica locally in a docker and plot results from the python script invoked by jm_ipython.py. Therefore, rather than having two separate scripts, I prefer to keep the assignment.

thorade:
Very nice to also have JModelica tests!

As you mention that time increases for Travis, there are three ways to make Travis run the tests less often:
 - adding [ci skip] to the commit message
 - doing multiple commits locally and then push them all in a single commit
 - use the Travis Auto-Cancellation feature

The third option can be changed here:
https://travis-ci.org/ibpsa/modelica-ibpsa/settings
Should we turn ON the Auto cancel feature?

mwetter:
@thorade : Thanks for the hint. I enabled the auto-cancel feature as I don't think having many old tests are more useful than faster test results.

Issue #884, opened 2018-02-03
from_dp for linearized PressureDrop


Mathadon:
The `PressureDrop` component currently implements the pressure drop equation as

```
m_flow*m_flow_nominal_pos = k^2*dp;
```

this should be 

```
      if from_dp then
        m_flow = k^2*dp/m_flow_nominal_pos;
      else
        dp = m_flow*m_flow_nominal_pos/k^2;
      end if;
```
Dymola is then better able to tear the linear system of equations. In a test flow network this leads to 31 iteration variables instead of 129 for a system of 240 linear equations when using `from_dp=true`.

PR #886, opened 2018-02-03
modified implementation of PressureDrop for linearized=true

Mathadon:
now depends on from_dp
avoiding parameter divisions
see #884

### DIFF
--- a/IBPSA/Fluid/FixedResistances/PressureDrop.mo
+++ b/IBPSA/Fluid/FixedResistances/PressureDrop.mo
@@ -10,14 +10,14 @@ model PressureDrop
                   Dialog(group = "Transition to laminar",
                          enable = not linearized));
 
-  final parameter Real k(unit="") = if computeFlowResistance then
+  final parameter Real k = if computeFlowResistance then
         m_flow_nominal_pos / sqrt(dp_nominal_pos) else 0
     "Flow coefficient, k=m_flow/sqrt(dp), with unit=(kg.m)^(1/2)";
 protected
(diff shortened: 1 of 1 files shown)


IBPSA/Fluid/FixedResistances/PressureDrop.mo
```diff
@@ -17,6 +17,11 @@ protected
   final parameter Boolean computeFlowResistance=(dp_nominal_pos > Modelica.Constants.eps)
     "Flag to enable/disable computation of flow resistance"
    annotation(Evaluate=true);
+  final parameter Real coeff(unit="")=
```
mwetter: Is there a reason for `coeff` to be public rather than protected?
Is there a reason for settings its units to `""` (and what does this mean? No units, or non-specified units?)

Mathadon: `Coeff` is/was protected, so I presume you misread or mistyped?

There is no specific reason for this other than that I copied and modified:
```
final parameter Real k(unit="") = if computeFlowResistance then 
        m_flow_nominal_pos / sqrt(dp_nominal_pos) else 0
    "Flow coefficient, k=m_flow/sqrt(dp), with unit=(kg.m)^(1/2)";
```
I suppose this should be changed too then?

Mathadon: fixed in 5bbaea953be10a86fe1db6c928e4fd90237f2cd8

mwetter: You are right, `coeff` was already protected. I will merge it.

Issue #882, opened 2018-01-28
Use Xi as input for source models & removing BaseProperties


Mathadon:
Currently, the models in `IBPSA.Fluid.Sources` allow the use of an input for `X`, while in many cases `X` is not directly available. E.g. a `FluidPort` only contains the independent fractions `Xi`. I would like to add the option to use `Xi` as an input instead of `X` to avoid having to do obsolete computations to compute `X` from `Xi`. This process is error-prone and annoying since you're often working with multidimensional vectors. I can't see any downside if we add this option in an advanced tab to avoid confusion with users about the difference between using `X` and `Xi` as an input. 

A related issue: these models instantiate `Medium.BaseProperties`. I'm not sure why this is the case, possibly to the check consistency of `X`, temperatures that are out of the bounds, etc? In any case, I'd like to add the option to remove this instantiation since it introduces additional equations and/or translation time to the problem.

Any remarks before I implement this?

mwetter:
Both are worthwhile, but maybe we can move `X` to the advanced tab rather than `Xi`. `Xi` is more useful in my view.

Can you check whether removing `Medium.BaseProperties` indeed gives faster translation and simulation. I would think the overhead is small, and it allows to catch unreasonable user input.

Mathadon:
Using this example
```
model Boundary
  parameter Integer n = 200;
  Boundary_pT[n] bou(
    redeclare package Medium = Medium,
    each use_p_in=true,
    each use_T_in=true);
  Modelica.Blocks.Sources.Cosine cosine(freqHz=1, offset=1e5);
  Modelica.Blocks.Sources.Cosine cosine1(offset=290);
  replaceable package Medium = IBPSA.Media.Air;
equation 
  for i in 1:n loop
    connect(bou[i].p_in, cosine.y);
    connect(cosine1.y, bou[i].T_in);          
  end for;
end Boundary;
```

I get following translation/simulation times (seconds) for 1e5 steps.

With `BaseProperties` 
```
n=200 - 15/0.555
n=400 - 21/1.19
n=600 - 32/1.61
```

Without
```
n=200 - 5/0.0344
n=400 - 5.5/0.0328
n=600 - 7/0.0371
```

I'll make `BaseProperties` conditional with default enabled?

Mathadon:
I verified the analysis with

```
model Boundary
  parameter Integer n = 200;
  IDEAS.Fluid.Sources.Boundary_pT[n] bou(
[...]

PR #887, opened 2018-02-03
Issue882 bou xi

Mathadon:
for #882


Mathadon:
@mwetter I have some difficulties interpreting travis' results:

```
Traceback (most recent call last):
  File "../bin/runUnitTests.py", line 181, in <module>
    show_gui = args.show_gui)
  File "../bin/runUnitTests.py", line 88, in _runUnitTests
    retVal = ut.run()
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/buildingspy/development/regressiontest.py", line 2510, in run
    r = self._checkReferencePoints(ans)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/buildingspy/development/regressiontest.py", line 1737, in _checkReferencePoints
    y_sim = self._getSimulationResults(data, warnings, errors)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/buildingspy/development/regressiontest.py", line 908, in _getSimulationResults
    r = Reader(fulFilNam, "dymola")
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/buildingspy/io/outputfile.py", line 179, in __init__
    self._data_ = DyMatFile(fileName)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/buildingspy/thirdParty/dymat/DyMat/__init__.py", line 44, in __init__
    self.mat = scipy.io.loadmat(fileName, chars_as_strings=False)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/scipy/io/matlab/mio.py", line 141, in loadmat
    MR, file_opened = mat_reader_factory(file_name, appendmat, **kwargs)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/scipy/io/matlab/mio.py", line 64, in mat_reader_factory
    byte_stream, file_opened = _open_file(file_name, appendmat)
TypeError: 'NoneType' object is not iterable
make: *** [test-dymola] Error 1
/home/travis/.travis/job_stages: line 169:  4638 Terminated              travis_jigger $! $timeout $cmd
```

is this a timeout that should be resolved simply by restarting? Also, some of the JModelica tests fail, which I cannot run on this system so it is hard to figure out why. Would it make sense to extend the unit test script such that errors from Dymola/JModelica are reported to the end user?

IBPSA/Fluid/Sources/BaseClasses/PartialSource_Xi_C.mo
```diff
@@ -0,0 +1,97 @@
...
+  assert(not use_X_in or not use_Xi_in,
+    "Cannot use both X and Xi inputs, choose either of the two.");
+
+  if not use_X_in and not use_Xi_in then
+    Modelica.Fluid.Utilities.checkBoundary(Medium.mediumName, Medium.substanceNames,
+      Medium.singleState, true, X_in_internal, "Boundary_pT");
+  end if;
+
+equation
+  if use_X_in then
```
mwetter: Why is there only a test on `use_X_in` but not also on `use_Xi_in` as in the initial equation section above?

IBPSA/Fluid/Sources/Boundary_pT.mo
```diff
@@ -1,95 +1,9 @@
...
-    C_in_internal = C;
-  end if;
-  medium.p = p_in_internal;
-  medium.T = T_in_internal;
-  medium.Xi = X_in_internal[1:Medium.nXi];
-  ports.C_outflow = fill(C_in_internal, nPorts);
+  extends IBPSA.Fluid.Sources.BaseClasses.PartialSource_p;
+  extends IBPSA.Fluid.Sources.BaseClasses.PartialSource_T;
+  extends IBPSA.Fluid.Sources.BaseClasses.PartialSource_Xi_C;
   annotation (defaultComponentName="bou",
```
mwetter: You will also need to move the graphical annotation such as
```
        Text(
          visible=use_C_in,
          extent={{-164,-90},{-62,-130}},
          lineColor={0,0,0},
          fillColor={255,255,255},
          fillPattern=FillPattern.Solid,
          textString="C"),
...
        Line(
          visible=use_C_in,
          points={{-100,-80},{-60,-80}},
          color={0,0,255}),
```
to the respective base class.

IBPSA/Fluid/Sources/BaseClasses/PartialSource_T.mo
```diff
@@ -0,0 +1,39 @@
...
+equation
+  connect(T_in, T_in_internal);
+  if not use_T_in then
+    T_in_internal = T;
+  end if;
+  for i in 1:nPorts loop
+     ports[i].h_outflow  = h_internal;
+  end for;
+  connect(medium.h, h_internal);
+  annotation (
```
mwetter: This `info` section is a bit too short (as is the one of the other base classes).

IBPSA/Fluid/Sources/BaseClasses/PartialSource_Xi_C.mo
Mathadon: I thought `checkBoundary` only checked whether the values of `X` sum up to one, but this is not the case. I'll fix this!

IBPSA/Fluid/Sources/Boundary_pT.mo
Mathadon: Okay, since `Boundary_p*` and `MassFlowSource_*` now share the same code, I removed the lines that connect the inputs to the model icon, since the model icons have a different shape.

Mathadon:
@mwetter `Boundary_p` and `MassFlowSource` now share the same code for the temperature input. Therefore, the location of the input has changed for `MassFlowSource`. Do we need to update all example models accordingly (routing of the connections has to change)?

Mathadon:
I did not remove the property check since if it had a use in the past, it now does so too. However, I set `verifyInputs=false` by default in 9fe791c.

Mathadon:
@mwetter I think I addressed your comments and it looks like unit tests finally passed :)

mwetter:
@Mathadon : it would be good if you could update the location of the connection to the new connector placements. I also corrected a bug in b9221bf

Mathadon:
@mwetter I revised the connections by doing a 'find usage' on `IBPSA.Fluid.Sources.MassFlowSource_T` and `IBPSA.Fluid.Sources.MassFlowSource_h`. I manually opened and saved all models that use these two models, which seems to have adjusted the line connections as expected.

Issue #907, opened 2018-04-10
prescribedHeatFlowRate in PartialEffectiveness


Mathadon:
`IBPSA.Fluid.HeatExchangers.BaseClasses.PartialEffectiveness` computes `Q_flow` from the fluid inlet temperatures, which do not depend on `Q_flow`. Therefore this model should have `prescribedHeatFlowRate1=prescribedHeatFlowRate2=true`. Otherwise the model can produce division of zero by zero. A unit test should be added that produces this behavior and the model should be fixed.

Mathadon:
The following model produces the error:

```
model Test
  extends ConstantEffectiveness(PSin_1(height=0, offset=1E5));
end Test;
```

```
Error: Scalar system is always singular for hex.port_a1.h_outflow = (hex.Q1_flow-60499.614*sin_1.ports[1].m_flow)/((-1.0)*sin_1.ports[1].m_flow) = -0/-0
```

but it seems a bit dull to add this model just for testing how it behaves at zero flow? 

So I'll only change the default values of `prescribedHeatFlowRate1=prescribedHeatFlowRate2=true` in `PartialEffectiveness` and `IBPSA.Fluid.MassExchangers.ConstantEffectiveness`.

Mathadon:
okay, then I'll add it!

Mathadon:
fixed in #920

PR #920, opened 2018-04-30
set prescribedHeatFlowRate=true in hex and partialhex

Mathadon:
for #907


IBPSA/Resources/Scripts/Dymola/Fluid/MassExchangers/Validation/ConstantEffectivenessZeroFlow.mos
```diff
@@ -1,2 +1,2 @@
 simulateModel("IBPSA.Fluid.MassExchangers.Validation.ConstantEffectivenessZeroFlow", stopTime=1, method="dassl", tolerance=1e-6, resultFile="ConstantEffectivenessZeroFlow");
-createPlot(id=7, position={0, 0, 1250, 724}, y={"hex.epsS", "hex.epsL"}, range={0.0, 1.0, 0.71, 0.8899999999999999}, erase=false, grid=true, colors={{28,108,200}, {238,46,47}});
+createPlot(id=1, position={15, 10, 1045, 783}, y={"hex.QLat_flow", "hex.mWat1_flow", "hex.mWat2_flow", "hex.Q1_flow", "hex.Q2_flow", "hex.mWat_flow"}, range={0.0, 0.00028000000000000003, -1.2000000000000002, 1.4}, grid=true, colors={{28,108,200}, {238,46,47}, {0,140,72}, {217,67,180}, {0,0,0}, {162,29,33}});
```
Mathadon: This is ok for me. The reason for printing `eps` was that these values are not a function of time such that only two time samples are stored. So essentially I just want to test wether the model compiles correctly, without error messages. I'm not sure if that's something that we can/should test for in a more structured way?

Issue #877, opened 2018-01-05
updates to MediumColumn


Mathadon:
`IBPSA.Airflow.Multizone.MediumColumn` does not implement `allowFlowReversal` for the computation of `h_outflow`, etc. 

Furthermore, this model computes the density from the absolute pressure. The pressure can depend on the mass flow rate and thus on the density, which generates an algebraic loop. The impact of the absolute pressure on air/water density is negligible such that we could add an option to simplify the density computation by using a constant pressure for the evaluation of density.

mwetter:
The intentional use of `MediumColumn` is to compute the pressure difference due to differences in density, for example for modeling stack effect in buildings. 

I think `allowFlowReversal` would be confusing and probably error-prone. Consider two rooms, one on the 1st floor, and one on the second floor, and suppose the pressure is understood to be at the average of the respective room's height (as in `Examples.ReverseBuoyancy`). Then, you want to describe with what temperature the medium is computed, and not whether you have reverse flow. Also, flow is likely to reverse in models where buoyancy is a driving force.

Using only `rho=f(T)` rather than `rho=f(T, p)` as long as all instances of `MediumColumn` are configured consistently. Otherwise you may get flow even if all medium pressures and temperatures are equal, e.g., a perpetuum mobile. Do you see a measurable performance gain for this?

Mathadon:
@mwetter I can see why `allowFlowReversal=false` can be a bad idea since the flow can indeed reverse due to inverse temperature gradients, but then the parameter should be removed? It currently exists and only modifies the `min`,`max` attributes.

I would then propose to use `rho=f(T)` by default instead of `rho=f(T,p)` since pressure differences are typically in the order of 100 Pa, which is negligible compared to the absolute pressure of 100 kPa. I'm wondering whether the existing implementation in fact can't also introduce a perpetuum mobile.. Since the model is asymmetric: it takes the absolute pressure of one of the two ports to compute the density.
Anyway, I saw a case where this pressure-dependency leads to an algebraic loop in 1 iteration variable. Whether this is measurable depends on the case of course. I'm rather safe than sorry when it comes to avoiding algebraic loops.

mwetter:
@Mathadon : Yes, this parameter should be removed (which will be a backward incompatible change). It is also used in `PowerLawResistance` which is used by the orifice model. It looks like it got in there when the base classes were changed. In my view, we should not allow setting `allowFlowReversal = false` in this package.

[...]

PR #922, opened 2018-05-01
Issue877 medium column

Mathadon:
see #877


Mathadon:
@mwetter this is ready to be merged

mwetter:
- [ ] In ``MediumColumn``, I suggest to remove ``use_constantPressureForDensity`` as I don't see why someone would set this to ``false``. E.g., I suggest you implement the behavior for ``use_constantPressureForDensity = true``.
- [ ] I don't see why someone would set ``p_default`` to a value different from ``Medium.p_default``. What else should it reasonably be? 

Hence, if we hardcode ``p_default=Medium.p_default`` and remove the option to set ``use_constantPressureForDensity=false``, the model would get simpler. The error will be small as 1E5/(1E5-500)=1.005, which is an extreme case. Usually, dp across rooms is a few Pascal only.


Note that

* ``IBPSA.Airflow.Multizone.MediumColumn`` removes the parameter ``allowFlowReversal``
* ``IBPSA.Airflow.Multizone.BaseClasses.PowerLawResistance`` sets ``final allowFlowReversal = true`` (which was not ``final`` before). This affects ``IBPSA.Airflow.Multizone.Orifice`` and ``IBPSA.Airflow.Multizone.EffectiveAirLeakageArea``.

These are non-backward compatible changes and need to be handled according to https://github.com/ibpsa/modelica-ibpsa/wiki/Workflow-for-code-changes#non-backward-compatible-changes

Can you add a snippet for the conversion script that the libraries can then add to their conversion scripts?

Mathadon:
@mwetter I incorporated your suggestions.

The script should be:
```
convertModifiers("IBPSA.Airflow.Multizone.MediumColumn",{"use_constantPressureForDensity"}, fill("",0), true);
convertModifiers("IBPSA.Airflow.Multizone.MediumColumn",{"p_default"}, fill("",0), true);
convertModifiers("IBPSA.Airflow.Multizone.MediumColumnDynamic",{"use_constantPressureForDensity"}, fill("",0), true);
convertModifiers("IBPSA.Airflow.Multizone.MediumColumnDynamic",{"p_default"}, fill("",0), true);
``` 

Should I include this in a .mos file in the repo?

**Comment by mwetter** [comment]
@Mathadon The conversion script should be
```
convertModifiers("IBPSA.Airflow.Multizone.MediumColumnDynamic", {"allowFlowReversal"}, fill("",0), true);
convertModifiers("IBPSA.Airflow.Multizone.MediumColumn", {"allowFlowReversal"}, fill("",0), true);
convertModifiers("IBPSA.Airflow.Multizone.EffectiveAirLeakageArea", {"allowFlowReversal"}, fill("",0), true);
convertModifiers("IBPSA.Airflow.Multizone.Orifice", {"allowFlowReversal"}, fill("",0), true);
```
as it needs to list the parameter values that changed. The ones that you listed were never part of a model other than during the development on this branch.

@thorade @mlauster Can you please approve this pull request if you agree with the changes, as this is a backward incompatible change.

Mathadon:
@mwetter I think this can be merged?

mwetter:
This is ready to be merged if the tests pass.

Issue #928, opened 2018-05-24
incorrect formatting of .travis.yml


Mathadon:
Travis seems to have updated their parsing of our scripts, which leads to different results. For a more detailed discussion see https://github.com/lbl-srg/BuildingsPy/issues/208. I will update .travis.yml

mwetter:
@Mathadon : I close this as it is corrected on the master branch.

PR #929, opened 2018-05-24
updated travis script by adding quotes

Mathadon:
this fixes an issue where not all unit tests are run
see https://github.com/lbl-srg/BuildingsPy/issues/208 and #928


.travis.yml
```diff
@@ -46,7 +46,7 @@ before_install:
 install:
   - pip install --upgrade pip setuptools wheel
   - pip install --only-binary=numpy,scipy,matplotlib numpy scipy matplotlib
-  - pip install git+https://github.com/lbl-srg/BuildingsPy@master
```
mwetter: Why is this change needed?

Mathadon:
@mwetter I must have started from the wrong branch since this is indeed not intended.. sorry about that. I have reverted the change.

mwetter:
@Mathadon No problem, good thing we do code review.

PR #943, opened 2018-06-07
Issue314 storage tank

### LINKED ISSUE #314: consider adding storage tank to library
Consider adding or a merge the storage tank model from `IDEAS`, `BuilingSystems`, `Buildings` and maybe `AixLib`  to the `Annex60` library.

Next step is to discuss the four models during the phone call.
Filip to check with his co-worker about the stratification model.

Mathadon:
This closes #314

@mwetter I have copied the `Fluid.Storage` folder, copied dependencies (except `Buildings.BaseClasses.Icon` and CDL), updated file and url references, copied unit tests en reference results, adjusted the CDL example, etc. If you want to see what changes I made to the code, I committed everything pretty neatly, so you can go through the commit log commit by commit.

For the CDL example, unit tests are different since my hysteresis implementation has a 'high low', which is inverted in the `BooleanToReal` block.

There are other differences when plotting, but this may be related to the postprocessing and I don't think that this invalidates the example:
screenshot from 2018-06-07 12 03 30

Does this require a review? I did not check the code again now, but I looked at it in detail in the past.

@damienpicard fyi


mwetter:
@Mathadon :  This looks ready to merge if you agree with my last revisions.

PR #964, opened 2018-06-23
Issue962 extended assert

### LINKED ISSUE #962: extend assert error messages
Sometimes asserts are triggered but since Dymola does not necessarily output results exactly at that timestamp, these asserts can be tough to debug when the occur in exceptional circumstances. I then often have to add additional print statements in the assert message itself, like the two last lines below for `IDEAS.Fluid.Interfaces.StaticTwoPortConservationEquation`.

```
    assert(noEvent( abs(Q_flow) < 200*cp_default*max(m_flow_small/1E3, abs(m_flow))),
   "In " + getInstanceName() + ": Energy may not be conserved for small mass flow rates. 
   The implementation may require prescribedHeatFlowRate = false. 
   The heat flow rate equals " + String(Q_flow) +
   " W and the mass flow rate equals " + String(m_flow) + " kg/s")
```

Are these things that I can systematically start making pull requests for, or does this make the error messages too 'heavy'?

Mathadon:
This closes #962 

@mwetter can you review this?


Mathadon:
This now produces errors like

```
 The following error was detected at time: 1.000000000001
  In ConservationEquation.ste: 
   The heat flow rate equals -100 W and the mass flow rate equals -5.00044e-13 kg/s, 
   which results in a temperature difference 119503 K > dTMax=200 K.
   This may indicate that energy is not conserved for small mass flow rates. 
   The implementation may require prescribedHeatFlowRate = false.
  Failed condition: noEvent(abs(dyn.Q_flow) < 836800.0*max(ste.m_flow_small/1000.0, abs(ste.port_a.m_flow)))
```

PR #925, opened 2018-05-14
Issue924 csv reader writer

Mathadon:
@mwetter thanks for revising the code! To unittest the implementation, I extended the functionality of the CSV reader a bit further such that it can write files in the format of `CombiTimeTable`. This requires the correct header to be prepended to the file, which contains the number of rows, which are only available upon completion fo the simulation. Therefore the header has to be prepended to the file. I do this by reading the file to memory and then writing a new file. I'm not sure whether this is the most appropriate solution. Perhaps this functionality should be disabled by default.

PR #967, opened 2018-06-25
Issue924 csv reader writer

### LINKED ISSUE #924: csv writer
I think it would be useful to have a simple csv writer model that generates a csv file by sampling some inputs. This would result in a file type that is easier to post-process than the usual .mat files. I'm making this model so might as well add it to the library?

Mathadon:
This replaces #925 
This closes #924 

To do:
 - [x] execute `prependString` in destructor instead of `terminal()` section if JModelica unit tests fails.


Mathadon:
@mwetter you mentioned that JModelica does not execute code called by `terminal()` here https://github.com/ibpsa/modelica-ibpsa/pull/925#discussion_r189393681. However, the JModelica unit tests seem to pass, which suggests that the header is produced correctly? Otherwise the unit test would fail since the `CombiTimeTable` would be unable to read it. So I think that there is no need to move that piece of code to the destructor?

mwetter:
@Mathadon I am surprised that it works with the `terminal()`. Did you look at the file to make sure it is correct? As there are potential problems with `terminal()` (as an FMU may not know that it is the last call) I think it is better to put it in the destructor.

Mathadon:
I did not run the test locally. I'm just basing myself on the CI output. 

I'll make the change then!

Mathadon:
@mwetter I have updated the implementation. This is ready if the unit tests pass.

mwetter:
@Mathadon : I wrote the wrong package name in https://github.com/ibpsa/modelica-ibpsa/issues/924#issuecomment-388216277:
Buildings uses `Utilities.Reports`, not `Utilities.IO.Reports`, so I have to move `Buildings.Utilities.Reports.Printer` anyway. Hence, I wonder if we should use `IBPSA.Utilities.IO.Files`. This would in the future also allow to add a file reader in this package.
Please let me know and I can do the rename with `buildingspy.development.refactor.move_class`.

Otherwise it is ready to merge if all tests pass.

Mathadon:
@mwetter sounds good, feel free to modify the structure!

mwetter:

Mathadon:
@mwetter thanks for all the help with this! I think this model will prove to be very useful..!

Issue #422, opened 2016-02-24
add borefield model


mwetter:
This issue is to implement the borefield model of @damienpicard which is needed for `IDEAS` and `Buildings`. If interesting for other libraries, we can discuss to move it to the `master` when it is finished/cleaned up.

damienpicard:
@mwetter I uploaded the code on the branch #issue422_borefield.

damienpicard:
@mwetter I've added unit tests for all the examples except some examples on the uppermost level. Those take longer to initialize and to run and I'm not sure they should be added to the unit test suit. There are 2 examples with FIXME which do not compile:
```
Annex60.Fluid.HeatExchangers.GroundHeatExchangers.Borefield.BaseClasses.Aggregation.Examples.transientFrac
```
and
```
Annex60.Fluid.HeatExchangers.GroundHeatExchangers.Borefield.BaseClasses.BoreHoles.BaseClasses.Examples.SingleBoreHolesInSerie
```
 I'm not sure why, as I can use the models contained in these examples by other models without getting an error.

SebastianRoc:
Hello mwetter,

I am working on a model for a thermal ground storage with freezing the included water with Modelica at the moment. That is why I am intersted in your model, because I search for an example in which thermal storage is solved in 3D-model.
Could you provide me the file of the model?
Thanks in advance.
Sebastian

thorade:
@SebastianRoc the model and files are in the branch `issue422_borefield`.
https://github.com/ibpsa/modelica/tree/issue422_borefield
If you used git to clone the repository, you can just switch to that branch,
otherwise you can download that branch as a zip file:
https://github.com/ibpsa/modelica/archive/issue422_borefield.zip

Mathadon:
Or if that does not work, you can have a look at the IDEAS (https://github.com/open-ideas/IDEAS) library, which also contains the model in `IDEAS.Fluid.HeatExchangers.GroundHeatExchangers.Borefield.Examples`

MassimoCimmino:
@mwetter :

@damienpicard and I are resuming the development of the geothermal model. Should the development still be done on this issue and its associated branch (`issue422_borefield`)?

The broad scope of the work to be done is to (1) restructure the IDEAS model, (2) add automatic evaluation of temperature response function at initialization, (3) add medium package for propylene glycol-water (and potentially other) mixtures, (4) add validation cases from experimental and field monitored data.

mwetter:
[...]

PR #974, opened 2018-06-29
Issue422b borefield

MassimoCimmino:
As promised, this PR concerns the review of a selection of packages implemented in #422 .

The completed packages/models are the following:
- IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.GroundHeatTransfer
- IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Data
- IBPSA.Utilities.Cryptographics
- IBPSA.Utilities.Math (New functions are introduced: besselJ0, besselJ1, besselY0, besselY1, binomial, exponentialIntegralE1, factorial, fallingFactorial)

Work is still ongoing for the rest of the packages in `GroundHeatExchangers`.


icupeiro:
hey @MassimoCimmino !
Just FYI
I got some errors on my Linux machine due to case sensitive package.order
With the folder BoreHoles (the package is Boreholes) and the Type BoreHoleConfiguration (the package is BoreholeConfiguration)
Already solved in my fork...

Cheers!

MassimoCimmino:
Thank you @icupeiro

mwetter:
To to list:
- [ ] Revise IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.GroundHeatTransfer
- [ ] Revise IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Data
- [ ] Revise IBPSA.Utilities.Cryptographics
- [ ] Revise IBPSA.Utilities.Math (New functions are introduced: besselJ0, besselJ1, besselY0, besselY1, binomial, exponentialIntegralE1, factorial, fallingFactorial)
- [ ] Verify source of new C functions to make sure license is compatible with our BSD license.
- [ ] Rename `nbBor`, `nbSeg`, `nbTim{Sho,Lon,Tot}` to change `nb` to `n` as we use `n` in other models.
- [ ] Rename `p_max` to use a better abbreviation, or explain why `p_max` is the 'Number of cells per aggregation level'

damienpicard:
Thanks a lot @MassimoCimmino and @AlexLaferriere for this massive work. Impressive!
Here some remarks/feedback:
- is there a reason why you don't use a replaceable construction for the borehole model in `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.BaseClasses.partialBorefield`?
 As they both extend from `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Boreholes.BaseClasses.PartialBorehole` it should work fine and parameters could be propagated automatically by using `constrainedby`.

 I have had troubles some years ago with replaceable in the borefield model, so I'm fine with current implementation but they might have fixed the bug by now and it might work fine.

- Userguide should be at the top of package
- The revision history with my name does not reflect the major changes that I made in 2018. But it does not matter I guess.
- Why did you change `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Boreholes.BaseClasses.InternalResistancesOneUTube`? Just for graphical reason or did you change more? Revision history suggests you have made a full new model but it looks very similar to the original one.
- Same for `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Boreholes.BaseClasses.InternalResistancesTwoUTube`
- The original model `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Boreholes.Examples.BoreholeOneUTube` included the finite element cylindrical ground model. I understand why you removed it in `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Boreholes.Examples.BoreholeDynamics` but I would keep it in the other models as it illustrates a stand-alone borehole model. A user might be interested by it as it allows you to get the temperature profile in the ground (in the radial direction)
- In `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Boreholes.BaseClasses.Functions.Validation.InternalResistancesOneUTube_NegativeResistance` you mention: _In this case, the shank spacing is defined such that the pipes are close to the borehole wall, rendering the short-circuit thermal resistances negative. The capacity location x is then automatically set to zero. _. I'm not sure why x has to be zero?

I'm not in my office right now and I don't have a license for dymola, so I could not run the models. I'll still have a look at it this Thursday but everything looks fine. Thank you also for your impressive validation work! I'm curious to run the validation examples and see the results.

Let me know if you want me to handle myself some of my remarks.

AlexLaferriere:
@damienpicard Regarding the first five points:


- No, there is no particular reason, thanks for pointing this out.

- Thanks for pointing this out (though it should be noted that, either way, the User's Guide is still incomplete for the time being)

- I made some slight changes to more properly reflect the real history (by including the "old" history in the new partial classes), though in all cases the "old" internal resistance models didn't have any history for me to go by. For now what I did was simply to add "First implementation" by Damien Picard (June 2018) to these models, if this is insufficient we can always add more detailed history.

- It's just a graphical change for it be more compatible with the partial model (so that the same ports are situated in the same place). This should hopefully be clearer with the new revision history.

- Same as above


I've made the changes described above, and @MassimoCimmino will push these (along with various other changes) in the near future and will also address your two remaining points. In the meantime, these changes are also on the issue422b_borefield branch of my fork of modelica-ibpsa.

Let me know if you have any further questions or remarks.

damienpicard:
@AlexLaferriere Thanks a lot for your changes, they are 👍 

@MassimoCimmino I looked further at the validation and I have some comments for the sandbox validation: the paper does not include the heat capacity of the grout which is an important factor. Further more, the fin-effect of the borehole aluminum casting is also problematic. How did you deal with it? This points could be mentioned in the documentation.

@MassimoCimmino I also looked at the ground source model validation and I'm a bit surprised but the bump in the temperature output:
image

How do you explain it?

@MassimoCimmino The documentation from `IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.Boreholes.BaseClasses.Functions.Validation.InternalResistancesTwoUTube` mentions that it validates the model based on Bauer's paper. However, the paper does not contain the necessary data to perform validation. Is it correct that you used random parameters?

MassimoCimmino:
@damienpicard 

- **Cylindrical ground model:** I removed it from the examples so the tests for the boreholes only include models from the `Boreholes` package. Should we add a case for this in `GroundHeatExchangers.Examples`?
- **Capacity location x set to zero:** There were non-physical responses in the double U-tubes models during the early transient period (i.e. Heat flow from the colder tube to the warmer tube, making the cold tube temperature drop below the ground temperature). To dampen this effect, the best choice is to set `x=0`.
- **Sandbox validation:** Heat capacity was taken from other papers that used the sandbox data. The borehole resistance is prescribed (`use_Rb=true`) to the value reported by the sandbox paper and the internal resistances are adjusted to match `Rb`.
- **Temperature bump:** There was a bump in the heat injection rate during the experiment, likely caused by an air bubble going through the tiny pipes and messing with the PID control. You can see a smaller bump in the experimental temperature measurement.
- **Documentation of `Functions.Validation.InternalResistancesTwoUTube`:** Perhaps the wording of the documentation is not correct there. The tests do not compare to any data from the literature.

**TODO**
- [ ] Add reference to other sandbox papers
- [ ] Complete documentation of Sandbox validation to explain the prescribed `Rb`
- [ ] Documentation of `Functions.Validation.InternalResistances[...]`

MassimoCimmino:
@damienpicard Note that the borehole is not modeled in this validation case. In the experiment, the heat flow rate measured is that transferred to the fluid (There will be a delay between this and the heat transferred through the borehole wall). The borehole wall temperature was measured directly.

We should definitely still schedule a Skype meeting. Are you available some time this week?

mwetter:
@MassimoCimmino @damienpicard  @AlexLaferriere 
I committed some model revisions to this development branch, mainly regarding coding style and documentation, without any changes to the physics.

Before I further edit the models, can one of you please run the regression tests using
```
../bin/runUnitTests.py -s IBPSA.Fluid.HeatExchangers.GroundHeatExchangers
```
and update the reference results. Some reference results were already different before I made my changes, and it is better if one of you can commit the changes to make sure they are correct.

Note that I changed some temperature sensors in the examples to remove the temperature state of the sensor (if there is no reverse flow) and hence you will get fewer states in the new reference results.

I also suggest to rename
```
IBPSA.Fluid.HeatExchangers.GroundHeatExchangers to IBPSA.Fluid.HeatExchangers.Ground
```
rename
```
IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.GroundHeatTransfer to
IBPSA.Fluid.HeatExchangers.Ground.HeatTransfer
```
rename
```
IBPSA.Fluid.HeatExchangers.GroundHeatExchangers.GroundHeatTransfer.CylindricalGroundLayer
to
IBPSA.Fluid.HeatExchangers.Ground.HeatTransfer.Cylindrical
```
I can do this with a rename script but prefer to have a version where all reference results are correct and hence I can run the regression tests to make sure I don't break the code.

mwetter:
Todo:

- [ ] Address the new "fixme" in the code (related to documentation).
- [ ] Also add the source files used to create the graphics (so that others can adapt them if the model is revised).

MassimoCimmino:
I added the source files for the figures.

Related to the 'fixme's:
- I am unsure if we should automatically calculate the parameter `tLoaAgg` (time resolution of load aggregation) to be equal to the travel time of the fluid along one tube leg. It may be useful in practice to have this parameter set to a fraction of an hour, especially if other events are generated periodically on an hourly (or similar) schedule. Also, this recommendation comes from the simulation of a step temperature change at the borehole inlet. The borehole wall temperature change then converges to a `tLoaAgg`-independent value for values of `tLoaAgg` of the order of the fluid travel time. For greater values of  `tLoaAgg`, the short-term borehole wall temperature predictions change. This only becomes an issue if the simulation needs accurate predictions at the scale of minutes or below.
- @damienpicard could comment on the need of the `dynFil` parameter since it was imported from IDEAS. I do not mind removing the parameter. In my opinion the advanced user could also modify the value of the filling properties in the data records.

damienpicard:
@MassimoCimmino @mwetter sorry for the late reply. `dynFil` parameter is used to simulate the bore hole heat exchanger model in stead state (thus that the grout capacities are removed). This is necessary as we are often using the Euler integrator with integrator step of 30 s or more. The time constant of the filling, especially if vertically discretized is then too small.

damienpicard:
@AlexLaferriere this is truly a detail, but in the user guide, the picture of the borehole is not 100% correct: The resistance between the fluid and the capacity should be split into 3: convection, conduction in pipe, conduction in grout until the location of the capacity.

MassimoCimmino:
Thanks @damienpicard . We just updated the figure.

damienpicard:
@MassimoCimmino maybe we should clearly specify in the documentation that even if `R_b` is defined by the user directly, the grout and configuration parameters are still crucial as they are used for rescaling `R_a` and for the capacity values.

damienpicard:
@MassimoCimmino I made a figure for the borehole model in our paper. We could re-used it for the documentation if you want. Otherwise, the figure as you have adapted it is fine.

icupeiro:
I'd suggest to clarify better the parameter dBor against hBor. I got misled by this

mwetter:
@MassimoCimmino @damienpicard 
Are you OK with merging this, in which case we can have @Mathadon have a last look to see if my code revisions oversaw something (or added a new issue).

IBPSA/Fluid/BaseClasses/MassFlowRateMultiplier.mo
```diff
@@ -0,0 +1,65 @@
+within IBPSA.Fluid.BaseClasses;
+model MassFlowRateMultiplier "Model that multiplies the mass flow rate"
+  extends IBPSA.Fluid.Interfaces.PartialTwoPort;
+
+  parameter Real k "Gain for mass flow rate";
+initial equation
+  assert(k > Modelica.Constants.small or -k < -Modelica.Constants.small,
```
Mathadon: Is there a reason for putting this in an initial equation section? You can also add it in the equation section. In that case, I have the impression that dymola already evaluates the assert during translation when it only depends on parameters. In that way you get an error message earlier.

IBPSA/Fluid/BaseClasses/MassFlowRateMultiplier.mo
```diff
@@ -0,0 +1,65 @@
...
+equation
+    // Pressure drop in design flow direction
+  port_a.p = port_b.p;
+
+  // Mass balance (mass is not conserved by this model!)
+ port_b.m_flow = -k*port_a.m_flow;
+
+  // Specific enthalpy flow rate
+  port_a.h_outflow = inStream(port_b.h_outflow);
+  port_b.h_outflow = inStream(port_a.h_outflow);
```
Mathadon: No equations for `allowFlowReversal=false` but that's ok in this case since it's a very specific model.

IBPSA/Fluid/BaseClasses/MassFlowRateMultiplier.mo
```diff
@@ -0,0 +1,65 @@
...
+  port_b.C_outflow = inStream(port_a.C_outflow);
+
+  annotation (Documentation(info="<html>
+<p>
+This model multiplies the mass flow rate so that
+<code>0 = port_b.m_flow + k * port_a.m_flow</code>.
+</p>
+<p>
+The specific enthalpy, the species concentration and the trace substance concentration
+remains unchanged.
```
Mathadon: 'remain'

IBPSA/Fluid/BaseClasses/MassFlowRateMultiplier.mo
```diff
@@ -0,0 +1,65 @@
...
+Therefore, this model does not conserve mass or energy.
+It is used in 
+<a href=\"modelica://Buildings.Fluid.HeatExchangers.RadiantSlabs.ParallelCircuitsSlab\">
```
Mathadon: IBPSA

IBPSA/Fluid/BaseClasses/MassFlowRateMultiplier.mo
```diff
@@ -0,0 +1,65 @@
...
+Buildings.Fluid.HeatExchangers.RadiantSlabs.ParallelCircuitsSlab</a>
+to avoid having to instanciate multiple slab models in parallel, with each
```
Mathadon: instances

IBPSA/Fluid/BaseClasses/MassFlowRateMultiplier.mo
Mathadon: I verified this and the check is only evaluated at translation when `Evaluate=true`, for both cases.

MassimoCimmino:
`IBPSA.Fluid.BaseClasses.MassFlowRateMultiplier` is imported directly from `Buildings.Fluid.BaseClasses.MassFlowRateMultiplier`.

Should we modify it?

Mathadon:
> `IBPSA.Fluid.BaseClasses.MassFlowRateMultiplier` is imported directly from `Buildings.Fluid.BaseClasses.MassFlowRateMultiplier`.
> 
> Should we modify it?

I will make a suggestion

Mathadon:
approved upon merging #1029 !

Mathadon:
@mwetter can this be merged now? would be good to have the 3.0 release soon :) thanks!

MassimoCimmino:
I am OK with the current version of the code.

mwetter:
This is ready to be merged when the tests pass.

PR #976, opened 2018-07-07
added code for checking whether flow reversal occurs in PartialThreeW…

### LINKED ISSUE #963: assert for flow reversal
Many models define `allowFlowReversal` but do not check whether the flow does not revert. This can lead to model errors. In some cases I would like to check this explicitly. Therefore I propose to add an assertion check that is disabled by default in models such as `PartialTwoPort` and `PartialThreeWayResistance`. Are there objections against this?

Mathadon:
…ayResistance

This closes #963


mwetter:

PR #978, opened 2018-07-08
added nominal values for h_outflow to many FluidPorts in partials

### LINKED ISSUE #977: add nominal values for h_outflow in partials
Our `Fluid` partials currently do not define the nominal values of `h_outflow`. Properly chosing these values could affects convergence of algebraic loop solvers. Although I do not have a clear example for this, I'd like to add more appropriate values to be sure.

Mathadon:
This closes #977


mwetter:

PR #993, opened 2018-07-23
1.fixed all the fixmes 2. add the variable names into the message

SenHuang19:
1. fixed all the fixmes 
2. add the variable names into the message


mwetter:
Closed as we changed the architecture of the implementation.

PR #995, opened 2018-08-07
Correct two function paths to IBPSA library

PMehrfeld:
Function paths were still referred to Buildings library although functions exist in IBPSA.

### DIFF
--- a/IBPSA/Media/Refrigerants/R410A/pressureSatLiq_T.mo
+++ b/IBPSA/Media/Refrigerants/R410A/pressureSatLiq_T.mo
@@ -24,9 +24,9 @@ protected
 
 algorithm
   // Independent variable
-  x := Buildings.Utilities.Math.Functions.smoothMax(1-T/TCri, 1e-4, 5e-3) - x0;
+  x := IBPSA.Utilities.Math.Functions.smoothMax(1-T/TCri, 1e-4, 5e-3) - x0;
   // Pressure of saturated liquid refrigerant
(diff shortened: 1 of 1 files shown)


Mathadon:
@PMehrfeld how did you spot this? Because apparently the unit tests did not fail on this, so I presume that the function is not used in the library unit tests..

PMehrfeld:
@Mathadon : I did a check (in Dymola) of the `Media` package in `IBPSA` and `AixLib` to compare the number of warnings (and errors). And since I do not have `Buildings` in my Modelicapath, Dymola threw errors.

Mathadon:
ok, in that case I think we better remove the functions since they seem unused..

Mathadon:
@MassimoCimmino  @mwetter do you agree with the suggestion to remove the functions?

MassimoCimmino:
@Mathadon I agree. It is not required (and not used) by the compressor model.

I also took the time to scan the library for "`Buildings.`" and found no other instances of calls to `Buildings`, but found many instances in hyperlinks of the documentation (The text, not the links themselves). You can see some of them in `IBPSA.Airflow.Multizone.UsersGuide`.

Should we open an issue to clean up these (or is this voluntary)?

Mathadon:
@mwetter I suggest to remove the functions?

mwetter:
Removing the unused functions is fine, but this pull request will then need to be marked as backwards incompatible (unless the functions are protected).

PR #1002, opened 2018-08-28
Refactored model to add temperature gradient in soil

mwetter:
This changes the far soil temperature such that it has a temperature gradient along the depth of the borefield.

Todo:
- [x] `IBPSA.Fluid.HeatExchangers.Ground.Validation.Sandbox` gives different results. This is likely because `groTemRes` is now vectorized and but it still assumes the heat flux at the heat port is for the whole borefield length, but with the new implementation, it is only for the segment of the borefield.
- [x] Verify that the validations produce the same results (except for minor differences due to different initial fluid temperatures)


mwetter:
@MassimoCimmino : This is a first implementation that allows a vertical temperature gradient. There is still some more work (see todo above). I plan to work on this on Wed and Thu.
I think adding a vertical temperature gradient makes sense as this will be often needed for deep boreholes, and due to the connector `TSoi` being removed it would require a backward incompatible change.

mwetter:
Todo:
- [ ] Use only one instance `TGroRes` (as was done in the original code).
- [ ] Refactor `TGroRes` to take as an input `Q_flow` and produce as an output the temperature difference relative to the initial condition. This can then be superimposed to the undisturbed ground temperature.

PR #1020, opened 2018-09-12
Issue1019 if smooth transition

dhblum:
Changes if-statement to if-expression.


thorade:
Do you know `sign(x)` ?
I think it should work here, too. Not sure whether it generates events.

Mathadon:
Better check first whether `sign(x)` is supported by JModelica.

thorade:
I just tested it and it is supported. Also, it is specified in the language specification.

dhblum:
casADi doesn't seem to like the use of sign, I get the following error when I change the implementation to:

` y := sign(x) * (a + aX*(b + aX*(c + aX*(d + aX*(e + aX*f)))));`

```
Java error occurred: 
Exception in thread "main" java.lang.UnsupportedOperationException: Cannot convert expression to MX: sign(x)
```

Issue #937, opened 2018-06-05
issues with DoorDiscretised


Mathadon:
I made a simplified version of `IBPSA.Airflow.Multizone.DoorDiscretizedOpen` in IDEAS since `DoorDiscretizedOpen` is a bit too detailed in my opinion. @damienpicard suggested to compare my implementation with `DoorDiscretizedOpen`. From my understanding it's basically a combination of the stack effect and Bernoulli, combined with a discharge coefficient, which is also what I used in IDEAS, but without the fluid ports and without the discretisation.

I wrote down a number of questions/remarks regarding the implementation while I was figuring out what the model does exactly. 
 - For a temperature difference of 0K, the mass flow rate is non-zero due to the regularisation that is used:
```
dVAB_flow[i] = dV_flow[i]*
      IDEAS.Utilities.Math.Functions.smoothHeaviside(x=dV_flow[i], delta=
      VZer_flow/nCom) + VZer_flow/nCom;
```
![opendoor](https://user-images.githubusercontent.com/5970879/40964356-309f3674-68ab-11e8-849e-b4e7447dbcb0.png)
This is a violation of the second law, which is frowned upon!
I'm not sure why this regularisation is required in the first place? The model somewhere mentions that zero flow should be avoided, but I don't see why. Possibly this could simply be removed.
 - The same non-zero flow implementation implies that the flow in a port never reverses. This is not physical, since the flow direction at the top of a door depends on the sign of the temperature difference. This may lead a user to take wrong conclusions.
 - I don't see why this model would require the implemented error control? I think this can be removed.
 - The model uses a discretisation `nCom=10` by default, but for a simple test the difference in flow rates with `nCom=2` is only 4%. We could reconsider the default value.
 - It is not that clear what the parameters `hA` and `hB` mean. I presume that these are the heights corresponding to the pressures of respectively port_a1/port_b2 and port_b1/port_a2. We could reformulate this to make it a bit clearer.
 - Html is incorrect in some of the base classes, e.g.:
```
<li><i>February 8, 2005</i> by Michael Wetter:<br/>
       Released first version.
``` 
is not closed by `</li>`. Somehow this is not caught by tidylib?

 - The computation of the velocity `v` is in my opinion obsolete and leads to many divisions by parameters since `v[i]=dV_flow[i]/dA`. Some other variables like `vAB`, `VAB_flow`, `dpAB`, `v_top`, etc can either be inlined or removed such that fewer variables are saved.

[...]

PR #1021, opened 2018-09-13
Set allowFlowReversal to true. For #937

mwetter:
This corrects the min=0 violation discussed in https://github.com/ibpsa/modelica-ibpsa/issues/937#issuecomment-421066013


mwetter:
@Mathadon Is this ready to be merged?

Issue #855, opened 2017-11-08
filename too long


thorade:
Windows seems to limit filenames (including path) to 260 characters by default.
@vishen89 today was not able to clone to `C:\Users\Vishen\Desktop\Modelica\BuildingSystems` but instead had to clone to `C:\Modelica\BuildingSystems`
There is no immediate action required, just posting it here in case anyone else runs into the same problem.

![filename_too_long](https://user-images.githubusercontent.com/314571/32550197-7e019970-c48c-11e7-9bd7-bcae4f3ef5f0.PNG)

ainderfurth:
Agreed, I am running into the same issue. Especially when the library is a submodule in another project.

Is there a possibility to make filenames in IBPSA/Resources/ReferenceResults/Dymola/ shorter?

PR #1025, opened 2018-09-21
Renamed test as name is too long for JModelica CI

mwetter:
The reference result names are more than 260 characters, which causes problems in the JModelica CI testing. See also https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#maximum-path-length-limitation

This applies the following script to shorten the names
```python
import buildingspy.development.refactor as r

r.move_class("IBPSA.Fluid.Geothermal.Borefields.BaseClasses.Boreholes.BaseClasses.Functions.Validation.InternalResistancesOneUTube_NegativeResistance",
             "IBPSA.Fluid.Geothermal.Borefields.BaseClasses.Boreholes.BaseClasses.Functions.Validation.InternalResistancesOneUTubeNegative")
r.move_class("IBPSA.Fluid.Geothermal.Borefields.BaseClasses.Boreholes.BaseClasses.Functions.Validation.InternalResistancesTwoUTube_NegativeResistance",
             "IBPSA.Fluid.Geothermal.Borefields.BaseClasses.Boreholes.BaseClasses.Functions.Validation.InternalResistancesTwoUTubeNegative")
```


thorade:
This might also become a problem when cloning with git on Windows, see #855 
BuildingSystems/BuildingSystems is a couple of letters longer than modelica-ibpsa/IBPSA.

mwetter:
@thorade Best would be to make in .travis a test for the length of the reference result file names, and compute it for BuildingSystems which is the library with the longest name.
We could do this on a separate issue/branch so that the JModelica tests work while we fix it in a robust way.

PR #1030, opened 2018-09-25
Fix compilation

tbeu:
Need to include stdlib.h for declaration of malloc


Mathadon:
Creating pull request from our repo for access rights.

PR #1042, opened 2018-10-07
added custom implementation for realToString for #1041_precision

### LINKED ISSUE #1041: CSV writer precision
The CSV writer prints only 6 significant digits by default, which is a problem when using unix time stamps since this results in entries such as `1.51476e+09`, consequently rounding time to about 10000s. This should be fixed.

Mathadon:
this closes #1041


Mathadon:
superseded by #1043

PR #1046, opened 2018-10-26
printing instance name when temperature assert triggers in media

Mathadon:
closes #1045

PR #1047, opened 2018-10-28
Issue1045 media assert separate min max

### LINKED ISSUE #1045: include model name in Media asserts
The `BaseProperties` temperature check does not report in what model the temperature limits are exceeded. This can be improved for easier debugging.

mwetter:
This closes #1045.

I suggest this change instead of https://github.com/ibpsa/modelica-ibpsa/pull/1046.

Please see #1045 for the rationale.

### DIFF
--- a/IBPSA/Media/Air.mo
+++ b/IBPSA/Media/Air.mo
@@ -60,7 +60,7 @@ package Air
       "Temperature difference used to compute enthalpy";
   equation
     assert(T >= 200.0 and T <= 423.15, "
-Temperature T is not in the allowed range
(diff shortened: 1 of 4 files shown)


Mathadon:
@mwetter this is ok for me but I'd be in favor of also adding #1046

mwetter:
@Mathadon : You are right, the full instance name needs to be added for JModelica to write it in the assertion. I added it in https://github.com/ibpsa/modelica-ibpsa/pull/1047/commits/a5fce8923ffb9a94e767d87a3cf8aa64c39944f4

This pull request now integrates #1046.

PR #1049, opened 2018-10-28
Changed assertion from error to warning. For #1048

### LINKED ISSUE #1048: In StaticTwoPortConservationEquation, trigger a warning rather than an error in energy balance check
In `StaticTwoPortConservationEquation`, the check for the energy balance
```modelica
  if prescribedHeatFlowRate then
    assert(noEvent( abs(Q_flow) < dTMax*cp_default*max(m_flow_small/1E3, abs(m_flow))),
   "In " + getInstanceName() + ":
   The heat flow rate equals " + String(Q_flow) +
   " W and the mass flow rate equals " + String(m_flow) + " kg/s,
   which results in a temperature difference " +
   String(abs(Q_flow)/ (dTMax*cp_default*max(m_flow_small/1E3, abs(m_flow)))) +
   " K > dTMax=" +String(dTMax) + " K.
   This may indicate that energy is not conserved for small mass flow rates.
   The implementation may require prescribedHeatFlowRate = false.");
  end if;
```
is heuristic and hence should trigger a warning rather than an error that stops the simulation.

mwetter:
This closes #1048


mwetter:
@Mathadon I addressed your comment in https://github.com/ibpsa/modelica-ibpsa/pull/1049/commits/88a9d8d7c716ebf4d6bdbed0c4cabb441d97326d

PR #1063, opened 2018-12-21
updated default properties

### LINKED ISSUE #1062: Correct default properties of sandstone
@damienpicard 

In class ``IBPSA.Fluid.Geothermal.Borefields.Data.Soil.SandStone``, the default properties are not appropriate, e.g. density ``dSoi=540 kg/m3`` means it is lighter than water. I suggest to use values  from Table 4 in [this](http://www.15000inc.com/wp/wp-content/uploads/Geothermal-Heat-Pump-Design-Manual.pdf) reference as default, which are:
- ``kSoi=2.5 W/(m.K)``, which is ``1.44 Btu/h.ft.F`` (range from ``0.3`` to ``1.9`` in the table)
- ``dSoi=1800 kg/m3``, which is ``112 lb/ft3`` (range from ``80`` to ``120``  in the table)
- ``cSoi=1200 J/(kg.K)``, along with the ``kSoi`` and ``dSoi``, it gives thermal diffusivity of ``1.07 ft2/day`` (range from ``0.36`` to ``1.5``  in the table)

The table is reprinted from ASHRAE guideline ``Ground-Source Heat Pump Systems: Design of Geothermal Systems for Commercial and Institutional Buildings``.

JayHuLBL:
This closes #1062.


mwetter:
@JayHuLBL Please see the broken unit tests and correct them.
I also assigned @damienpicard as a reviewer as he implemented the original model, and because we want one non-LBNL reviewer if we change the code.

damienpicard:
@JayHuLBL thank you for reporting this anomaly. 

Notice that the model `IDEAS.Fluid.Geothermal.Borefields.Validation.Sandbox` is a validation with an experiment using sandstone. The properties used for the experiment are in `IDEAS.Fluid.Geothermal.Borefields.Validation.BaseClasses.SandBox_Soil`:
```
cSoi=1275 J/(kg.K),
dSoi=2000 kg/m3
```

The volumetric heat capacity you propose is slightly less than the one used in the experiment (2 160 000 J/kg.m3 instead of 2 550 000 J/kg.m3), but this is fine for me.

However, the updated reference results are strange to me. What is causing the changes in the validation models?

mwetter:
@JayHuLBL : Make sure you use Dymola 2019 and not 2019 FD01 to generate the reference results.

IBPSA/Resources/ReferenceResults/Dymola/IBPSA_Fluid_Geothermal_Borefields_BaseClasses_Boreholes_BaseClasses_Examples_InternalResistancesOneUTube.txt
```diff
@@ -1,14 +1,14 @@
...
-TPip2.T=[3.131499938964844e+02, 3.131499938964844e+02]
+TWal.T=[2.981499938964844e+02, 2.981499938964844e+02]
```
damienpicard: @mwetter and @JayHuLBL is the different organization of the reference file due to the different Dymola version?? This would be pretty annoying...

mwetter: @damienpicard : Variable names won't change between Dymola versions, but the numeric can slightly change.

damienpicard: @mwetter I meant that the variable `time` moved from line 9 to line 13.

mwetter: @damienpicard That could be fixed by sorting the output in BuildingsPy when rewriting the reference results, which seems currently not to be the case.

damienpicard: @mwetter this might be useful to avoid overloading the git history with unnecessary changes.

mwetter:
@JayHuLBL : Do you have a reason for why `TPip2.T` is now `TWal.T`. Is this change intentional?

@damienpicard : the change in order is only because some variables did not pass the test and hence the file is rewritten. Otherwise, if all results are within tolerance, then the reference result file would remain unchanged.

JayHuLBL:
@mwetter The `TPip2.T` is still in the file, in line 12. I will update the reference results with Dymola 2019.

JayHuLBL:
@mwetter @damienpicard 
I reruned the unit test with Dymola 2019. The reference results have no change from the one with Dymola 2019FD1.

PR #1077, opened 2019-01-09
Removed parameter m_flow_nominal for #970

### LINKED ISSUE #970: Remove parameter m_flow_nominal in Airflow.Multizone.MediumColumnDynamic
The parameter `m_flow_nominal` can be converted to a constant so that it no longer need to be set by the user. It is only used to assign it in `vol`, where it is only used for regularization if the model is steady-state. But this is practically never the case as the only purpose of this model, compared to `MediumColumn`, is to add dynamics.

A reasonable setting would be
```modelica
// Set m_flow_nominal to a default of about 10 air changes per hour
constant Modelica.SIunits.MassFlowRate m_flow_nominal = V/360
  "Default value for nominal mass flow rate, which is only used for regularization if the model is configured as steady-state";
```

mwetter:
This closes #970 

The entry for the conversion script is
```
convertModifiers("IBPSA.Airflow.Multizone.MediumColumnDynamic", {"m_flow_nominal"}, fill("",0), true);
```

### DIFF
--- a/IBPSA/Airflow/Multizone/BaseClasses/DoorDiscretized.mo
+++ b/IBPSA/Airflow/Multizone/BaseClasses/DoorDiscretized.mo
@@ -7,8 +7,6 @@ partial model DoorDiscretized
 
   parameter Modelica.SIunits.PressureDifference dp_turbulent(min=0) = 0.01
     "Pressure difference where laminar and turbulent flow relation coincide. Recommended: 0.01";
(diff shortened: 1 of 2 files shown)


mwetter:
@Mathadon 
@mlauster 
@nytschgeusen 
Can you please indicate whether you agree with this backward incompatible change.

mlauster:
Fine for Aachen

Issue #1072, opened 2019-01-04
Refactor Fluid.Sources


mwetter:
This issue is to refactor the `Fluid.Sources` package to avoid repeated use of the same base class. Currently, JModelica issues the warning:
```
In component bou:
  The component nPorts is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.
Warning at line 11, column 3, in file '/mnt/home/mwetter/proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA/Fluid/Sources/BaseClasses/PartialSource.mo',
In component bou:
  The component verifyInputs is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.
Warning at line 15, column 3, in file '/mnt/home/mwetter/proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA/Fluid/Sources/BaseClasses/PartialSource.mo',
In component bou:
  The component ports is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.
Warning at line 24, column 3, in file '/mnt/home/mwetter/proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA/Fluid/Sources/BaseClasses/PartialSource.mo',
In component bou:
  The component flowDirection is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.
Warning at line 26, column 3, in file '/mnt/home/mwetter/proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA/Fluid/Sources/BaseClasses/PartialSource.mo',
In component bou:
  The component p_in_internal is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.
Warning at line 29, column 3, in file '/mnt/home/mwetter/proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA/Fluid/Sources/BaseClasses/PartialSource.mo',
In component bou:
  The component Xi_in_internal is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.
Warning at line 32, column 3, in file '/mnt/home/mwetter/proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA/Fluid/Sources/BaseClasses/PartialSource.mo',
In component bou:
  The component X_in_internal is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.
[...]

PR #1088, opened 2019-01-26
Removed base properties

mwetter:
This removes the `Medium.BaseProperties`, see https://github.com/ibpsa/modelica-ibpsa/issues/1072#issuecomment-457789800


mwetter:
I close this as the BaseProperties are used to verify the inputs, and because Air has not constant T_min and T_max, there is no way to verify the inputs without using the BaseProperties.
By default, the BaseProperties are removed, so there is no overhead for this implementation.

PR #1091, opened 2019-01-31
Issue1087 jmodelica simulation

mwetter:
This adds more changes for the JModelica regression tests, including a configuration file `conf.json`.


IBPSA/Resources/Scripts/BuildingsPy/conf.json
```diff
@@ -0,0 +1,26 @@
+[
```
mlauster: What is this script for? To exclude specific tests from regression? Or is it more an example how to use this new fancy stuff?

Issue #1092, opened 2019-01-31
Temperature States in Plug Flow Pipe


MichaMans:
Hi everyone,

for an easier usage of the PlugFlowPipe Model, I would suggest to implement the show_T boolean for state calculations at the ports.

For more detail:

**What is the problem?**

At the moment there is no direct implementation of showing the temperature states of the PlugFlowPipe Model ports. This feature would be a handy addition to improve the usability of the model

**How do we want to solve it? Describe the solution you'd like**

As far as i'm aware, there should be two possible solutions:

Option 1: 
- Implementation in direct in the PlugFlowPipe model with the integration of the boolean show_T and the corresponding state calculations for the ports

Option 2:
- Implementation in the PartialTwoPortVector Model to enable the usage directly from the BaseClass

**Describe alternatives you've considered**

As far as i know there are now alternatives available in this pipe model

**Additional context**

@bramvdh91 @FelixBue @mwetter were would you like to see the implementation (if there are no general objections on that enhancement)

bramvdh91:
@MichaMans I think this would be a real valuable addition to the model, since it would make temperatures more accessible in a place where you expect them instead of needing to do all administration with separate temperature sensors. 
For me option 1 sounds more logical; in the vectorized port you just have a repetition of the temperature of the mixing volume if I'm not mistaken.

PR #1094, opened 2019-01-31
#1092 added the showT bool and state calculation

MichaMans:
This solves #1092 

Please comment here or use the issue for in depth discussions if this change needs revisions (especially if an implementation according to Option 1 is preferred)

### DIFF
--- a/IBPSA/Fluid/Interfaces/PartialTwoPortVector.mo
+++ b/IBPSA/Fluid/Interfaces/PartialTwoPortVector.mo
@@ -42,7 +42,7 @@ partial model PartialTwoPortVector "Partial component with two ports, one of whi
                           noEvent(actualStream(port_a.Xi_outflow))) if
          show_T "Medium properties in port_a";
 
-  Medium.ThermodynamicState stas_b[nPorts]=
+  Medium.ThermodynamicState sta_b[nPorts]=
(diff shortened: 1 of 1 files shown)


IBPSA/Fluid/Interfaces/PartialTwoPortVector.mo
```diff
@@ -30,6 +30,23 @@ partial model PartialTwoPortVector "Partial component with two ports, one of whi
...
+    "= true, if actual temperature at port is computed"
+    annotation(Dialog(tab="Advanced",group="Diagnostics"));
+
+  Medium.ThermodynamicState sta_a=
+      Medium.setState_phX(port_a.p,
+                          noEvent(actualStream(port_a.h_outflow)),
+                          noEvent(actualStream(port_a.Xi_outflow))) if
+         show_T "Medium properties in port_a";
+
+  Medium.ThermodynamicState stas_b[nPorts]=
```
mwetter: This should be `sta_b` not `stas_b` (even though it is a vector).

MichaMans: @mwetter I revised the naming according to your suggestion. The naming wasn't really clear for me but with the result filtering, its a good hint that ```sta_b``` will be more user friendly.

PR #1096, opened 2019-02-02
Issue842 multiyear

### LINKED ISSUE #842: ReaderTMY3 does not work for data going over into the next year

mwetter:
This closes #842 

Todo:
- [x] The test `BoundaryConditions/WeatherData/BaseClasses/Examples/ConvertTime.mo` fails as the model triggers an assertion when advancing to the next year.

@AnaConstantin: Can you please check the logic in `BoundaryConditions/WeatherData/BaseClasses/ConvertTime.mo`. I don't think the above test should fail.


mwetter:
- [ ] The `getTimeSpanTMY3` function should be changed so that it returns for end time the last time stamp of the weather data plus the time increment, because this is the time span for a weather data.
Then, `ConvertTime` can first check if end time minus start time is a whole year (31536000 seconds) and if so, use the `mod` function. If it is not a whole year, then it should use an assertion if `time > weaDatEndTime`.
This way, we can localize the decision of how long the weather data are, and the `ConvertTime` block simply needs to convert the time and not estimate (with `weaDatAveInc`) how the weather data may be. The parameter `weaDatAveInc` can then be removed as it will only be used inside `getTimeSpanTMY3`.

AnaConstantin:
@mwetter I think I get it. I'll go over the documentation and code comments to make sure endTime = lastTimeStamp + AvgIncrement is highlighted enough. I do not think this is the most intuitive way to do express endTime, but it's consistent with the rest of the code. Just one question: do you want to check only for a year, or also for multiples of a year, e.g. weather data stretching over two years?

mwetter:
@AnaConstantin : We should allow for multiple years of weather data. I believe all the information is available to do the check in the `initial equation` section.

mwetter:
@AnaConstantin @mlauster @Mathadon 
This is ready for review. I added more validation cases and updated the documentation.
Note that models that use the weather data reader such as the `ThermalZones.ReducedOrder` have a small difference at some time steps. I believe this is due to a change in numerics. I see similar changes in our larger models in the `Buildings` library. 
Please review and approve/require changes for this pull request so that we can close it.

Mathadon:
[image1.jpeg]
Working on it, but first things first! Greetings from the modelica conference!

AnaConstantin:
@mwetter thank you for the changes. To me it looks very good. My only question is if the assert has to be outside of the when loop.

mwetter:
@AnaConstantin : Why do you think it should be inside the `when` clause? The conditions `(time - weaDatEndTim) < shiftSolarRad` and `when modTim > pre(tNext) then` are different. If you propose to move the assertion inside the `when` clause, then we should probably use the same conditions and test to make sure it is correct. Did you look into it?

AnaConstantin:
@mwetter you are right. I didn‘t think it through. This when clause is better than the older versions, since it gets activated only when tNext has to be changed. So the condition in the assert is totally separate. Next time I‘ll take your advice and verify my assumptions before stating them. Sorry about that.

IBPSA/BoundaryConditions/WeatherData/BaseClasses/ConvertTime.mo
```diff
@@ -12,28 +16,57 @@ block ConvertTime
...
-  constant Modelica.SIunits.Time year=31536000 "Number of seconds in a year";
-  discrete Modelica.SIunits.Time tStart "Start time of period";
+  constant Modelica.SIunits.Time shiftSolarRad=1800 "Number of seconds for the shift for solar radiation calculation";
+  parameter Modelica.SIunits.Time lenWea = weaDatEndTim-weaDatStaTim "Length of weather data";
+
+  parameter Boolean canRepeatWeatherFile = abs(mod(lenWea, 365*24*3600)) < 1E-2
+    "true if the weather file can be repeated, since it has the lenth of a year or a multiple of it";
 
+  Modelica.SIunits.Time tNext "Start time of next period";
+  function getNextTime
```
Mathadon:
```suggestion
  function getNextTime "Function that computes the next time when a switch needs to happen"
```

IBPSA/BoundaryConditions/WeatherData/BaseClasses/ConvertTime.mo
```diff
@@ -12,28 +16,57 @@ block ConvertTime
...
-    tStart = integer(modTim/year)*year;
+  when modTim > pre(tNext) then
+    // simulation time stamp went over the end time of the weather file
+    //(last time stamp of the weather file + average increment)
+    tNext = if canRepeatWeatherFile then getNextTime(modTim, lenWea) else integer(Modelica.Constants.inf);
   end when;
-  calTim = modTim - tStart;
+  calTim = if canRepeatWeatherFile then modTim - (tNext - lenWea) else modTim;
+  assert(canRepeatWeatherFile or (time - weaDatEndTim) < shiftSolarRad,
+    "Insufficient weather data provided for the desired simulation period.",
```
Mathadon:
```suggestion
    "In " + getInstanceName() + ": Insufficient weather data provided for the desired simulation period.",
```

IBPSA/BoundaryConditions/WeatherData/BaseClasses/getLatitudeTMY3.mo
```diff
@@ -14,7 +14,7 @@ algorithm
     IBPSA.BoundaryConditions.WeatherData.BaseClasses.getHeaderElementTMY3(
       filNam=filNam,
       start="#LOCATION",
-      name=  "longitude",
+      name = "latitude",
```
Mathadon: looks scary, but apparently this is just for error reporting!

IBPSA/BoundaryConditions/WeatherData/ReaderTMY3.mo
```diff
@@ -1452,9 +1460,28 @@ and allows the following configurations:
   </td>
 </tr>
 </table>
+<!-- ============================================== -->
+<h4>Length of weather data and simulation period</h4>
+<p>
+If weather data span a year, which is the default for TMY3 data, or multiple years,
+then this model can be used for simulations that span multiple years. The simulation
+start time need to be set to the clock time of the respective start time. For example,
```
Mathadon:
```suggestion
start time needs to be set to the clock time of the respective start time. For example,
```

IBPSA/BoundaryConditions/WeatherData/ReaderTMY3.mo
```diff
@@ -1452,9 +1460,28 @@ and allows the following configurations:
...
+to start at January 2 at 10am, set start time to <code>t=(24+10)*3600</code> seconds.
```
Mathadon: I don't think this is sufficiently specific/clear. This assumes that the timezone that you define in the TMY3 file corresponds to the time zone of the date that you mention in the documentation?

mwetter: @Mathadon : Not sure what you mean by not being clear. Can you add/propose better wording? Note that the weather files for Chicago and San Francisco both start at `0`. Hence I am not sure what you mean.

Mathadon:
```suggestion
to start at January 2 at 10am, set start time to <code>t=(24+10)*3600</code> seconds.
For this computation, the used date and time (here January 2, 10 am) must be expressed in the same time zone
as the one that is used to define the TMY3 file. This is usually the local (winter) time zone.
The parameter `timZon` represents the TMY3 file time zone, expressed in seconds compared to UTC.
```
@mwetter does this formulation clarify what I mean?

mwetter:
@Mathadon @AnaConstantin 
Thanks for the feedback, and the clearer wording from Filip. I will merge it after the tests pass.

PR #1100, opened 2019-02-14
fix bug in CalendarTime.mo for non-zero offset.

### LINKED ISSUE #1099: CalendarTime shows bug with non-zero offset
When `CalendarTime` is used with non-zero offset, an unexpected behaviour appears: the next day starts at hour 1 instead of 0 and the minutes goes to a higher value than 60:

Examples: The simulation of `Example/CalendarTime.mo` with `calendarTime2016.offset = 3600` gives this:

image

damienpicard:
This closes #1099 .

### DIFF
--- a/IBPSA/Utilities/Time/CalendarTime.mo
+++ b/IBPSA/Utilities/Time/CalendarTime.mo
@@ -259,7 +259,7 @@ equation
 <li>
 February 14, 2019, by Damien Picard:<br/>
 Fix bug when non-zero offset by substracting the offset from hourSampleStart and daySampleStart (see <a href=\"https://github.com/ibpsa/modelica-ibpsa/issues/1099\">#1099</a>).
-<li>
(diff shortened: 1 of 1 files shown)


IBPSA/Utilities/Time/CalendarTime.mo
```diff
@@ -257,6 +257,10 @@ equation
   Documentation(revisions="<html>
 <ul>
 <li>
+February 14, 2019, by Damien Picard:<br/>
+Fix bug when non-zero offset by substracting the offset from hourSampleStart and daySampleStart (see <a href=\"https://github.com/ibpsa/modelica-ibpsa/issues/1099\">#1099</a>).
+<li>
```
Mathadon:
```suggestion
</li>
```

mwetter:
This is ready to merge after the CI tests pass.

Issue #1109, opened 2019-04-04
condensing cooling coil model


Mathadon:
We are working on a cooling coil model that supports condensation, without using discretisations.

thorade:
Would that be a moving boundary model, with 3 segments, or a different approach?

Mathadon:
No, rather a simple model with average specific enthalpy between inlet and outlet and the e-NTU method. PR soon!

Mathadon:
#1113 contains a first version of the model

Many things still need to happen, but the main issue right now is to choose a parameterization. At the Aachen meeting I showed that there was an issue when cross-validating the cooling coil model. This is caused by the fact that the current parameterization of the heating coil model implicitly assumes that the nominal conditions (Q_flow_nominal, m_flow_nominal, etc) are at fully dry conditions. Based on these conditions the UA value of the heat exchanger is computed inconsistently, which causes the observed errors. The question is now how to include the degree of condensation at nominal conditions into the model parameters. Options include parameters for:
- fraction of latent or sensible heat flow rate at nominal conditions
- absolute value of latent or sensible heat flow rate at nominal conditions
- wet bulb inlet temperature at nominal conditions
- absolute humidity or relative humidity of inlet temperature at nominal conditions

Depending on the choice, we may have to do (many) inverse computations of the normal equations to compute UA. If wrong parameters values are supplied, the resulting inverse computations may give infeasible results. From that point of view option 3) or 4) look most dummy-proof, however 1) and 2) are easier to use. @mwetter @icupeiro  Any suggestions here?

icupeiro:
@Mathadon the manufacturer information we are using provides absolute values of latent and sensible heat flow rate, so I'd stick to this

Mathadon:
@icupeiro indeed that makes sense. The question is also how to specify the 'nominal conditions'. For a dry coil this is just the nominal flow rate and temperature difference but for a wet coil the model also requires some indication of the model humidity. This can be provided either using:
- inlet absolute humidity
- inlet relative humidity
- inlet wet bulb temperature

Right now I'm using absolute humidity. The disadvantage is that this value is not that easy to know and it's hard to provide a good default value. Therefore it's perhaps better to use relative humidity, or allow both?

icupeiro:
@Mathadon standard EN 1397:2015 establishes test conditions of 19 degC air wet bulb temperature, so I would stick to that as the default parameter.

Mathadon:
[...]

PR #1112, opened 2019-04-07
Issue1109 coolingcoil

Mathadon:
this merges base classes from buildings for #1109:
i.e. I (manually) merged `IBPSA.Fluid.HeatExchangers.DryCoilEffectivenessNTU`, its depending baseclasses and the examples of the involved models, unit tests and reference results.

Based on these models I can make a second pull request for the new models, which will result in a clear git diff and revision history for the changes that I'll make.


mwetter:
@Mathadon : I saw you modified `epsilon_C.mo`. This has new code
```modelica
  // Gain that goes to zero as CMin_flow gets below deltaCMin
  // This is needed to allow zero flow
  gai := IBPSA.Utilities.Math.Functions.spliceFunction(
           pos=1,
           neg=0,
           x=CMin_flow-deltaCMin,
           deltax=deltaCMin/2);
  if (gai == 0) then
    NTU := 0;
    eps := 1; // around zero flow, eps=Q/(CMin*dT) should be one
  else
    NTU :=gai*UA/CMinNZ_flow;
    eps := gai*IBPSA.Fluid.HeatExchangers.BaseClasses.epsilon_ntuZ(
                  NTU=NTU,
                  Z=Z,
                  flowRegime=flowRegime);
  end if;
```
and `gaiNTU` has been removed (or implemented differently). What is the rational for sneaking in this change? In particular, the revision notes in https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Fluid/HeatExchangers/BaseClasses/epsilon_C.mo state that the `if` construct has been removed because it can cause a discontinuous derivative, and now it is in again and the function declares `smoothOrder=1`.

Did you merge an old version of the Buildings library?

And if you need the new outputs, did you add them to a regression test?

Mathadon:
@mwetter I did indeed merge a wrong version of buildings. I.e. commit f6911ca5cefc4d4c13623842c7693b4113b4a71d. I'll update the PR.

mwetter:
@Mathadon : This is ready to merge after all tests pass.

PR #1115, opened 2019-04-09
Issue1114 json writer

### LINKED ISSUE #1114: utility for writing simulation result to json format
We have `IBPSA.Utilities.IO.Files.CSVWriter` for writing time series data. However, we do not have a tool for writing output (e.g. json format) for a single variable value, e.g. for outputting the total energy use of a building during the whole year at `terminal()`.

I propose to make a block with the same interface as `CSVWriter` that produces such a result file. The resulting files can then be read using https://github.com/modelica-3rdparty/ExternData .

Mathadon:
closes #1114 
@mwetter can you review this?


Mathadon:
The produced files have the following format:

```json
{
"key1" : 2.0000000000e+00,
"key2" : 2.0000000000e+00,
"key3" : 1.0000000000e+00
}
```

IBPSA/Resources/C-Sources/writeJson.c
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+
+void writeJson(const char* fileName, const char** varNames, const int numNames,  const double* varVals, const int numVals){
+	if (numNames!=numVals){
+		ModelicaFormatError("In writeJson.c: The supplied vector of names and values do not have equal lengths: %d and %d", numNames, numVals);
+	}
+
+	FILE *fOut = fopen(fileName, "w");
```
mwetter: fopen returns `NULL` if not successful and you need to guard against that.

IBPSA/Resources/C-Sources/writeJson.c
```diff
@@ -0,0 +1,34 @@
...
+	
+	writeJsonLine(fOut, "{\n", fileName);
+
+	int i;
+	for (i = 0; i < numNames; ++i){
+		if (strlen(varNames[i])>100){
+			ModelicaFormatError("The variable name \"%s\" is too long. It must be shorter than 100 characters.", varNames[i]);
+		}
+		char buffer[200];
```
mwetter: Fixed lenght buffer can lead to buffer overflow which is a security hazard. Uses formatted file writing directly or use `malloc` to allocate memory based on `strlen`

mwetter:
Todo
- [ ] Change inequality check to avoid "The expression "time > jsonWriterCustom.customTime" will generate a state event and not a time event." and also update the info section by replacing "exceeds" with "achieves".
- [ ] For JModelica, the terminal() function-like operator is not supported, and is currently evaluated to false and hence the files are not written. As a work-around, use an ExternalObject, store the values in C in each call, and when the destructor is called, write the values to disk.

IBPSA/Resources/C-Sources/writeJson.c
Mathadon: I'll use `printf` instead. I tried to catch an overflow using the length check. That's not good enough?

mwetter: No, this is not good enough: Instead of 
```
if (strlen(varNames[i])>100){
			ModelicaFormatError("The variable name \"%s\" is too long. It must be shorter than 100 characters.", varNames[i]);
```
and having the user fix it, simply use  `fprintf`, or else use `strlen()` to figure out how long the string is, and `malloc` to assign memory.

Mathadon: Sure, no problem. Just trying to understand why? It's mostly from a usability point of view then for the user?

Mathadon:
@mwetter I fixed some similar issues/error checks in the CSV writer code

Mathadon:
@mwetter The model is compliant with the Modelica specification. Can't we just add to the documentation that `Output.Terminal` is not supported in JModelica? Working around this is quite some work and adds obsolete overhead and code to an otherwise clean implementation.

mwetter:
We should make sure it works with JModelica; it probably is only an hour of work to set it up. Otherwise BOPTEST won't be able to use it, and neither our Spawn users.

Mathadon:
Let's do it for Spawn!

Mathadon:
@mwetter I have adjusted the code. I've tried to avoid code duplication and hence reused to existing functionality to check duplicate file names. That. however required me to introduce new fields in the existing struct. I.e. some fields are used for CSV writers and some others for JSON writers. 

I also partly fixed a bug where the duplicate file writer check returns incorrect output. Right now we produce:
```
FileWriter TerminalOutputs.json writes to file TerminalOutputs.json which is already used by another FileWriter.
```
I.e. the first `TerminalOutputs.json` should be the instance. name instead of the file name. This still has to be fixed. The remaining code is ready if unit. tests pass.

Mathadon:
see bcea8b4 for a revised implementation of the error message, now: 
```
FileWriter JsonWriter.jsonWriterOneVar writes to file TerminalOutputs.json which is already used by FileWriter JsonWriter.jsonWriterTerminal.
```

Mathadon:
@mwetter this is ready to be reviewed

mwetter:
@Mathadon 
I did the following revisions
- Removed fmu and json files
- Fixed code to comply with C90 (to avoid warnings from JModelica).
- Removed buf[255] with a safe call to asprintf
- Added header files

I think we should rename the files from `JsonWriter.mo` to `JSONWriter.mo` because we also capitalized `CSVWriter`, and both JSON and CSV are both abbreviations, so we should be consistent. Can you please do this rename and then merge when all tests pass.

IBPSA/Resources/C-Sources/fileWriterFree.c
```diff
@@ -51,10 +51,11 @@ void fileWriterFree(void* ptrFileWriter){
   /* If this FileWriter writes in a CombiTimeTable format, prepend the required header
   now that we know how many lines have been written. */
   if (ID->isCombiTimeTable){
-    char buf[255];
-    sprintf(buf,"#1\ndouble csv(%i,%i)\n",ID->numRows,ID->numColumns);
+    char* buf;
+    asprintf(&buf,"#1\ndouble csv(%i,%i)\n",ID->numRows,ID->numColumns);
```
Mathadon: Trying to figure out why unit tests fail. I find online that `asprintf` is a GNU extension. Not. sure what that means,  but is it supported by the docker/travis?

mwetter: I was wondering that too, but for jmodelica the compiler is in the docker and hence this should work on travis as it does locally where I used the same image.
I will change this back and see if this is the reason, but I doubt it is.

Mathadon: Perhaps we should add something to Buildingspy where the Dymola error message is included when the Simulation fails?

mwetter: @Mathadon : `asprintf` is not the problem as the test https://travis-ci.org/ibpsa/modelica-ibpsa/builds/520953878 also fails. I probably have to look into it next week when back from travel.  Branch`issue1114_jsonWriter_testing` is set up to run just this test.

Mathadon:
@mwetter I think the problem is caused by a case error, see 
0508383

Mathadon:
@mwetter this can be merged as far as I am concerned

Issue #1059, opened 2018-12-17
Signal overwrite/send block that does not require a socket connection


dhblum:
This issue is to add signal overwrite and send blocks that do not require a socket connection, as described in https://github.com/ibpsa/project1-boptest/issues/15.  This is an alternative solution implementation to https://github.com/ibpsa/modelica-ibpsa/issues/926.

dhblum:
The private parameter names should be changed to boptestOverwrite and boptestRead.

mwetter:
@dhblum Should the overwrite block be merged to the master or is it not yet stable?

dhblum:
@mwetter The block is not yet stable.  To do:
- [ ] Add enumeration parameter for KPI keywords
- [ ] Add string parameter for variable description

JavierArroyoBastida:
@dhblum and @mwetter, 
The enumeration of the KPI signals has been merged in https://github.com/ibpsa/modelica-ibpsa/pull/1116. So far we are using entries for power and energy because we wanted the emulator developer to decide where to integrate: in the model or in the KPI calculator in python. However, I find that misleading and prone to errors in the tagging and the KPI calculation processes. 
For instance the computation of cost and emissions requires power signals since prices and emission factors can vary over time whereas with an energy signal the information of when the energy was consumed is lost.
What do you think? should I leave the energy signals out and compute everything from powers?

dhblum:
@mwetter I've added documentation and reference results for this branch, though the unittests still seem to be failing on travis and I can't figure out why.  Can you please take a look at let me know what seems to be the issue?

See https://travis-ci.org/ibpsa/modelica-ibpsa/jobs/554825134

mwetter:
@dhblum 
the problem is that `package.order` is incorrect:
```
# Check for package.order to be up to date
(cd  ../../../../IBPSA; \
	  python -c 'import buildingspy.development.refactor as r; \
	  r.write_package_order(".", recursive=True)'; \
	  git diff --exit-code . )
diff --git a/IBPSA/Utilities/IO/SignalExchange/Examples/BaseClasses/package.order b/IBPSA/Utilities/IO/SignalExchange/Examples/BaseClasses/package.order
index 380159e..48dc8ed 100644
--- a/IBPSA/Utilities/IO/SignalExchange/Examples/BaseClasses/package.order
+++ b/IBPSA/Utilities/IO/SignalExchange/Examples/BaseClasses/package.order
[...]

PR #1119, opened 2019-04-11
issue1059_controlIONoSocket - Removed energy signals

JavierArroyoBastida:
follows https://github.com/ibpsa/modelica-ibpsa/pull/1116 according to https://github.com/ibpsa/project1-boptest/issues/49


mwetter:
@dhblum : Can you please save this package so that class is saved as its own file. This is needed so it integrates for example with the merge script that all libraries use.

dhblum:
Ok will do.  I've added to the to-do list in https://github.com/ibpsa/modelica-ibpsa/issues/1059.

PR #1123, opened 2019-04-16
Changed computeCSen to avoid volume to be a structural parameter

### LINKED ISSUE #1122: Change ConservationEquation to make fluidVolume not a structural parameter
The `ConservationEquation` makes the `fluidVolume` and hence any volume `V` in the mixing volumes a structural parameter. This can be changed by using a different criteria for `computeCSen`.

This is for https://github.com/lbl-srg/modelica-buildings/issues/1411

mwetter:
This closes #1122


mwetter:
@Mathadon : `Simplified2` fails: https://travis-ci.org/ibpsa/modelica-ibpsa/jobs/520812817
With CVode rather than Lsodar it works. I will change the solver.

Mathadon:
@mwetter Perhaps we should also report this convergence problem to Dassault?

Issue #1136, opened 2019-04-29
replace getline() function used in getTimeSpan.c to ensure platform independency


JayHuLBL:
The `getline()` function used in getTimeSpan.c will break JModelica test, as it's not supported on Windows. 

This issue intends to replace the `getline()` function with an alternative to avoid platform dependency.

JayHuLBL:
Close this issue as it has been merged in PR #1138.

PR #1137, opened 2019-04-30
replaced getline with fscanf #1136

JayHuLBL:
The alternative of `getline()` function include function `fscanf()` and `fgets()`. However, both of them requires to specify the object with length, to store the result:
```
int fscanf ( FILE * stream, const char * format, ... ); ...
The additional arguments should point to already allocated objects ...
```
```
char * fgets ( char * str, int num, FILE * stream );
```

`fscanf()` is used here since it is relatively safe to specify the maximum length of the first section (before first empty space) of each line.

### DIFF
--- a/IBPSA/Resources/C-Sources/getTimeSpan.c
+++ b/IBPSA/Resources/C-Sources/getTimeSpan.c
@@ -26,12 +26,15 @@
 char *concat(const char *s1, const char *s2) {
   const size_t len1 = strlen(s1);
   const size_t len2 = strlen(s2);
-  char *result = malloc(len1 + len2 + 1);
-  if (result == NULL){
+  size_t len = len1 + len2 + 1;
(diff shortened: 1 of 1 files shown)


mwetter:
@JayHuLBL This is not save against buffer overflow. Look into using `fgets` and reallocate memory as needed.

JayHuLBL:
@mwetter It's ready for review.

JayHuLBL:
@mwetter It has been updated to avoid memory leakage. 

The model `IBPSA.BoundaryConditions.WeatherData.BaseClasses.Examples.GetTimeSpanTMY3` can run with JModelica without problem.

mwetter:
@JayHuLBL Please also make sure that the failing tests work again.

JayHuLBL:
Close this PR as it will be addressed in #1138.

PR #1152, opened 2019-06-04
Issue1107 long term geothermal

### LINKED ISSUE #1107: Long-term temperature changes of geothermal bore fields
There is an issue with the integration of the finite line source solution in `IBPSA.Fluid.Geothermal.Borefields.BaseClasses.HeatTransfer.ThermalResponseFactors.finiteLineSource`. The upper bound of integration (in theory infinite) is set to 100.0. This is a correct upper limit for typical boreholes with diameters of the order of 6 inches.

However, in the implementation of the thermal response we decided to evaluate the finite line source analytical solution at a radius `radius = 0.0005*length`. For shorter boreholes (approx. 25 m or less), the radius becomes small and the integrand function is still non-negligible at a value of 100.0. The figure below shows the logarithm (log10) of the integrand.

FiniteLineSource_Integrand_BoreholeLength

This causes issues with the calculation of the thermal response, and becomes progressively worse with decreasing borehole length, as shown below for a 5 meter borehole (the thermal response should be strictly positive).

GFunction_1borehole_5meters

A conservative estimate for the upper bound of integration is `10/radius`, or `20000/length` in our case. This corrects the thermal response, see below.

GFunction_1borehole_5meters_fixed

MassimoCimmino:
This closes #1107.


mwetter:
@MassimoCimmino : Thanks for the changes. If the latest tests pass, this is ready to merge. I would however prefer to rename
```IBPSA.Fluid.Geothermal.Borefields.BaseClasses.HeatTransfer.ThermalResponseFactors.Examples``` to ```... Validation```. If you agree I can make this change.

Note my comment https://github.com/ibpsa/modelica-ibpsa/issues/1107#issuecomment-499984932. I did not find a better work-around until this issue is fixed in JModelica.

MassimoCimmino:
@mwetter : I agree it should be `IBPSA.[...].ThermalResponseFactors.Validation`.

I was a bit confused with the JModelica issue. Thanks for pointing this out.

PR #1161, opened 2019-06-13
Reverted assignment of borefield geometry

### LINKED ISSUE #1107: Long-term temperature changes of geothermal bore fields

mwetter:
This closes #1107, which now works with JModelica revision 12903.


MassimoCimmino:
This does run (in Dymola) on my side. Waiting on the tests.

mwetter:
@MassimoCimmino Is this OK to merge?

PR #1169, opened 2019-07-11
Issue1168 refactor alpha

### LINKED ISSUE #1168: Refactor heat transfer coefficient alpha in aixlib models
In some AixLib models the heat transfer coefficient `alpha` becomes `hConv` or `hRad` according to [https://github.com/RWTH-EBC/AixLib/wiki/Variable-Naming](url). This should also be done in `IBPSA.ThermalZones.ReducedOrder` to avoid problems at the next IBPSA/ AixLib merge.

katharinabr:
This will close #1168. I renamed `alpha` to `hRad` or `hCon` in the reduced order models. To prevent problems with other model libraries, I increased the version number and added a conversion script.


PMehrfeld:
@mwetter : We are not sure, why every second CI job fails. It seems not to be a Modelica error.  
Besides this CI issue, the PR is ready for review.

IBPSA/Resources/Scripts/Dymola/ConvertIBPSA_from_4.0.0_to_4.0.1.mos
```diff
@@ -0,0 +1,790 @@
...
+    "hConExtWall_const");
+convertElement(
+    "IBPSA.ThermalZones.ReducedOrder.Examples.SimpleRoomFourElements",
+    "thermalZoneFourElements.alphaExtWallConst",
+    "thermalZoneFourElements.hConExtWall_const");
+convertElement(
+    "IBPSA.ThermalZones.ReducedOrder.Examples.SimpleRoomOneElement",
+    "thermalZoneOneElement.alphaExtWallConst",
+    "thermalZoneOneElement.hConExtWall_const");
+convertElement(
```
mwetter: It looks like you misunderstand conversion scripts: Entries such as
```
convertElement(
    "IBPSA.ThermalZones.ReducedOrder.Examples.SimpleRoomOneElement",
    "thermalZoneOneElement.alphaExtWallConst",
    "thermalZoneOneElement.hConExtWall_const");
```
are not correct (or at least not needed). Because the _instance_ `thermalZoneOneElement` is an instance of the class `IBPSA.ThermalZones.ReducedOrder.RC.OneElement`, the correct entry should be
```
convertElement(
    "IBPSA.ThermalZones.ReducedOrder.RC.OneElement",
    "alphaExtWallConst",
    "hConExtWall_const");
```
This is sufficient, and will cause Dymola to also update the parameter assignment in `IBPSA.ThermalZones.ReducedOrder.Examples.SimpleRoomOneElement`. See also the Dymola manual.

Once you revised the conversion script, please convert an old example to the new library to make sure the script is correct, as it is easy to make mistakes in this file.

IBPSA/package.mo
```diff
@@ -483,12 +483,15 @@ The library has the following <i>User's Guides</i>:
 end UsersGuide;
 
 annotation (
-version="3.0.0",
-versionDate="2018-09-28",
+version="4.0.1",
```
mwetter: This should be version 4.0.0.

IBPSA/Resources/Scripts/Dymola/ConvertIBPSA_from_4.0.0_to_4.0.1.mos
```diff
@@ -0,0 +1,790 @@
+// Conversion script for IBPSA library
+//  from: Version 4.0.0
+//  to:   Version 4.0.1
+clear
```
mwetter: Because the current version of the `IBPSA` library is 3.0.0, you should call the new version `4.0.0`, and put your conversions to the existing file `ConvertIBPSA_from_3.0_to_4.0.mos`

IBPSA/Resources/Scripts/Dymola/ConvertIBPSA_from_4.0.0_to_4.0.1.mos
katharinabr: We have assumed that we should use dymola's automatic generation of conversion scripts. So we prevent overlooking parameters. Originally I wrote a shorter conversion script according to the Dymola manual. After Dymola wrote commands like
```
convertElement(
    "IBPSA.ThermalZones.ReducedOrder.Examples.SimpleRoomOneElement",
    "thermalZoneOneElement.alphaExtWallConst",
    "thermalZoneOneElement.hConExtWall_const");
```
I thought they might be necessary after all. But I get your point and deleted all entries I consider superfluous. Thinking about this: Would you consider the entries about the example and validation models, e.g.
```
convertElement(
    "IBPSA.ThermalZones.ReducedOrder.Examples.SimpleRoomFourElements",
    "alphaWall",
    "hConWall");
```
necessary? I don't think anyone's gonna use these models in another library. But I'm not sure if I'll forget use cases where it would be necessary to execute these commands.
I successfully tested the conversion on a few examples.

mwetter: I don't think it is important to convert parameters in `Examples` or `Validation`, but if you have already done the scripts, we can include them.

PR #1170, opened 2019-07-11
Issue1059 control io no socket

dhblum:
For https://github.com/ibpsa/modelica-ibpsa/issues/1059.


dhblum:
FYI @JavierArroyoBastida, I've made this PR and asked for a review by @mwetter .

dhblum:
@mwetter Thank you for completing your review.  I agree with all of your suggestions, except made one correction to revision documentation, which I've fixed in https://github.com/ibpsa/modelica-ibpsa/pull/1170/commits/efd933f25bc77300f23c449d7351b082f0c4cd75.

The change in parameter name to lower case is noted.  It will affect downstream tools, but should be fixed easily.  I've already added it as a to-do in the BOTEST repo at https://github.com/ibpsa/project1-boptest/issues/49.

mwetter:
Thanks @dhblum . 
@PMehrfeld , this is ready for review and merging.

mwetter:
@PMehrfeld : can you please review and merge if you agree with the implementation.

PR #1180, opened 2019-07-16
Refrigerants

### LINKED ISSUE #1093: New refrigerant package
## What is the problem?

Currently, there are no refrigerant models within the IBPSA library. Therefore, we want to add different refrigerants (R134a and probably R410A, R32, R744 and R290).

## Why do we want to solve it?

The refrigerants can be used, for example, for detailed heat pumps and refrigerating machines simulation models

## How do we want to solve it?

We will use the approach presented by Sangi et al. (see paper “A Medium Model for the Refrigerant Propane for Fast and Accurate Dynamic Simulations”; [http://www.ebc.eonerc.rwth-aachen.de/go/id/dncb/file/227214](url)). Sangi et al. calculated functions for the density, specific enthalpy and specific entropy for the three different regimes supercooled, superheated and two-phase. Focusing on high accuracy, we need to have six regions, that need to be implemented. These are not necessarily thermodyanmic regions, but they allow good fitting results within our model development. The regression functions were polynomial and fitted to the inverse energy equation of state.

## Who do we think will benefit from our solution?

We think that everyone will benefit from the models who needs refrigerants without using external media libraries. Moreover, the presented approach allows a fast and accurate simulation and, therefore, contributes to everyone who will focus the simulation speed.

cvering:
This closes #1093 

Hey alltogether, we merged our media models in the IBPSA structure. Therefore, we deleted Massimo Cimminos media model of R410A to avoid naming conflicts. Proving our approach, we added Validation Tests. At the same time we will go for a journal publication to show the functionalities in detail.


mwetter:
@cvering : Can you please make a pull request against the branch `issue1093_RefrigerantMerge` which I just created so we can revise before putting it on the master.
Before you make the pull request, please merge the latest master to your branch.
Also, please make sure your pull request only has files that indeed changes (currently there are hundreds of files that have only changes in formatting that should be removed).

mwetter:
Pull request will be closed and @cvering will open a new PR when he merged the updated master (which is now using MSL 4) to the refrigerant development branch.

Issue #1181, opened 2019-07-16
Update old annotations


LisaRivalin:
Some models use old annotations that need to be corrected. See for example below.
```
Buildings/Examples/VAVCO2/VAVSystemCTControl.mo
...
       Text(
        extent=[-140,234; -96,192],
        string="PAtm"),
```
The correct syntax is
```
Text(
          extent={{-44,48},{-26,36}},
          textString="aaa")
```

see https://github.com/lbl-srg/modelica-buildings/issues/1429

mwetter:
This is now on the master

PR #1182, opened 2019-07-17
Issue1181 update annotations

LisaRivalin:
Some models used old annotations causing graphical discrepancies (#1181 ). This has been addressed as follow:

- Changed "extent" coordinate convention from extent=[x1,x2;y1,y2] to extent={{x1,x2},{y1,y2}}
- When inside "annotation", changed string to textString


mwetter:
@Mathadon : Can you please provide a review so we have a non-LBL reviewer, and then merge.
This eliminates the dialog box in Dymola that asks whether old annotations should be converted.

mwetter:
@Mathadon : Can you please provide a review so we have a non-LBL reviewer, and then merge.

PR #1183, opened 2019-07-17
Issue1168 refactor alpha

### LINKED ISSUE #1168: Refactor heat transfer coefficient alpha in aixlib models

mwetter:
This closes #1168

This is non-backwards compatible because of the name in the public variable (`alpha` was changed to `h`).

According to https://github.com/ibpsa/modelica-ibpsa/wiki/Workflow-for-code-changes#non-backward-compatible-changes,
@Mathadon , @PMehrfeld , @nytschgeusen please approve or provide feedback, the latest by August 1 so that we can merge it.

@katharinabr : Thanks for the changes. Will this also require updates to TEASER?


PMehrfeld:
I approve this pull request.  
Only a two questions @mwetter:

- Is the `non-backward compatible` flag necessary. Since @katharinabr delivered a conversion script.
- In current master, the library is at v3.0.0, but the `convertClass` command for the model `IBPSA.Obsolete.Fluid.Sources.FixedBoundary` is triggered with `from(version="3.0.0"`. So this is actually too late, right? Normally each conversion script needs an increase in the minor or patch number. At least if the master branch should be correctly usable after every merge. Or do you only append the conversion script until the next major version such that every release of the library works stable but not the commits in the master inbetween? In this case, be aware that the version number in the `package.mo` was raised from `3.0.0` to `4.0.0` (see [this line](https://github.com/ibpsa/modelica-ibpsa/blob/7686116ff990a7f4d4e7bafe428ccade691c3c55/IBPSA/package.mo#L486)).

PMehrfeld:
Regarding TEASER: [This issue](https://github.com/RWTH-EBC/TEASER/issues/590) covers the refactoring. But thanks a lot for mentioning it!

mwetter:
@PMehrfeld : We change the version number and date only when we make an official release. I changed it back in cc1a25b3. (The master is always stable, but not backward compatible.) Even though we have a conversion script, we still want to flag it as `non-backward compatible`. There is no clear definition what non-backward compatible means in Modelica, but please see https://github.com/ibpsa/modelica-ibpsa/wiki/Workflow-for-code-changes#non-backward-compatible-changes for an (probably incomplete) interpretation.

@Mathadon, @nytschgeusen : Do you approve the changes?

PR #1189, opened 2019-07-26
Issue1188 refactor damper

### LINKED ISSUE #1188: Refactor damper models
This issue is to integrate into `modelica-ibpsa` the developments related to issue [#1298](https://github.com/lbl-srg/modelica-buildings/issues/1298) in `modelica-buildings`.
See the presentation at [Aachen](https://github.com/ibpsa/project1/blob/master/meetings/2019-04-03-expert-meeting-aachen/wp_1_1/ibpsa-project-1-lbl-actuators.html) and the [updated version](https://antoinegautier.github.io/plots/Actuators.html) describing the latest implementation.

AntoineGautier:
This closes #1188.

- `VAVBoxExponential` is retired. The option for the fixed resistance has been integrated into `Exponential`. (A conversion script is attached.)
- The exposed parameter `dp_nominal` in `Exponential` and `PressureIndependent` has been replaced by `dpExp_nominal` and `dpDam_nominal` (modification also automated by the above script). The motivation is to restrict the use of `dp_nominal` to the nominal pressure drop at the model boundaries (which is used by `PartialDamperExponential` in the homotopy initialization cf. `dp/dp_nominal_pos`). In the master branch, it is indeed the case for `Exponential`, it might be the case for `VAVBoxExponential` if `dp_nominalIncludesDamper==True` and it is never the case for `PressureIndependent`. Introducing the new parameters ensures consistency (and appears to be the best way to merge `VAVBoxExponential` into `Exponential` properly).
- The option to linearize the characteristics of the damper plus fixed resistance has been added.
- New limiting values for parameters `k0` and `k1` are used in `PartialDamperExponential` motivated by [https://antoinegautier.github.io/plots/Actuators.html#/6](https://antoinegautier.github.io/plots/Actuators.html#/6).
- `PressureIndependent`: 
    * the algorithm implemented to deal with small derivatives of `m_flow` / `dp` while ensuring a bounded error (approximately) is illustrated here [https://antoinegautier.github.io/plots/Actuators.html#/4](https://antoinegautier.github.io/plots/Actuators.html#/4). It relies on  1) an intermediary `dp_lim` variable limiting the linear domain where `m_flow = y_actual * m_flow_nominal + cregul *...`, 2) an extended domain for the upper spline interpolation and  3) a leakage flow supposed to be mainly turbulent (to limit the domain where the 2 flow functions are not inverse from each other). The resulting flow characteristics is also part of the updated example see `Buildings/Fluid/Actuators/Dampers/Examples/Damper.mo` case `preIndCha`. The impact on the test case described in #1298 is presented in [https://antoinegautier.github.io/plots/Actuators.html#/2/1](https://antoinegautier.github.io/plots/Actuators.html#/2/1).
    * the steep transition of the opening from 1 to 0 when `y=0` and `dp` transitions from negative to positive values is handled through a threshold value (2%) of the input signal below which the opening is forced to zero.

Note:

- Some models using `PressureIndependent` might fail to simulate without changing some parameters values to be consistent with the exponential damper characteristics that is now used by the model to compute the opening (typically models in `Buildings/Examples/VAVReheat` had a too high nominal pressure drop with a too low nominal velocity: those parameters have been changed manually, the script mentioned above does not handle that modification).

Conversion script:

[VAVBoxExponentialRefactor.py.txt](https://github.com/ibpsa/modelica-ibpsa/files/3437798/VAVBoxExponentialRefactor.py.txt)


mwetter:
- [ ] This needs another revision due to an inflection point in m_flow = f(y)

AntoineGautier:
To clarify the motivation for renaming the parameters for nominal pressure drop:

* In the master branch:

  * `Exponential`: `dp_nominal` is the pressure drop between the 2 fluid ports. This is consistent with the super class `PartialTwoPort` which defines `dp = port_a.p - port_b.p`. This super class is inherited from by all damper models.

  * `VAVBoxExponential`:  `dp_nominal` is either the pressure drop of the fixed resistance only (if `not dp_nominalIncludesDamper`) or the total pressure drop between the 2 fluid ports (if `dp_nominalIncludesDamper`).
  This class extends `PartialDamperExponential` which uses `dp_nominal` as the total pressure drop between the 2 fluid ports in the call to `homotopy(..., simplified=m_flow_nominal_pos*dp/dp_nominal_pos)`.

  * `PressureIndependent`: `dp_nominal` is the pressure drop of the damper only (which is the exact opposite of `VAVBoxExponential` with `dp_nominalIncludesDamper=false`), `dpFixed_nominal` is the pressure drop of the fixed resistance.

* I propose:

  * `Exponential` (merge of `Exponential` and `VAVBoxExponential`): new parameter `dpExp_nominal` is the pressure drop of the fixed resistance or between the 2 fluid ports depending on `dp_nominalIncludesDamper` value.
  * `PressureIndependent`: new parameter `dpDam_nominal` is the pressure drop of the damper only, `dpFixed_nominal` is the pressure drop of the fixed resistance.
  * For all damper models: `dp_nominal` is the pressure drop between the 2 fluid ports, which is consistent with the expression of `dp` in `PartialTwoPort` and the use of `dp_nominal_pos` in `PartialDamperExponential`.

Mathadon:
I would try to keep consistency with other similar models, e.g. `IDEAS.Fluid.Actuators.Valves.TwoWayLinear`. I.e. we should then have a parameter like `dpValve_nominal` -> `dpDam_nominal` sounds good, and `dpFixed_nominal` -> the same name can be used here? I'd try to do an implementation that does not have `dp_nominalIncludesDamper` since I find it a confusing option and there doesn't seem to be a reason why one of the three dampers should have it and the others and regular valves do not have it.

Mathadon:
That looks like a very low default. I'm more used to ~50Pa, see e.g. page 6 of https://www.trox.nl/downloads/1512ded69ad60d13/tvz.pdf?type=product_info . However, that's a VAV of course, not a damper.

Damper:
https://www.trox.be/nl/downloads/676c1818c3d9435e/5_10_3_vfr.pdf?type=product_info 
page 2, line 8 = fully opened. Lists 500 m3/h for 20 Pa and 125 mm. For more reasonable flow rates the pressure would of course be lower too.

AntoineGautier:
10 to 40 Pa seems the typical range for balancing ducted dampers see:
[http://www.swegon.lt/wp-content/uploads/2012/03/REACT.pdf](http://www.swegon.lt/wp-content/uploads/2012/03/REACT.pdf)
ASHRAE Guideline 46 INFORMATIVE ANNEX C shows pressure drop around 10 Pa for relief, outdoor and return air damper. The guideline also mentions typical range of face velocity for relief (5 to 8 m/s), outdoor (2 to 8 m/s) and return air damper (4 to 8 m/s).
Adopting 10 Pa as default in the library leads to a nominal face velocity of 6.1 m/s which seems reasonable (note that 1 m/s is the current default value).

Remark: ASHRAE Dampers and Air Flow Control p 44 shows that the application has a major effect on the pressure drop: depending on the configuration the same damper with the same face velocity can have a total pressure drop ranging from 6.5 Pa to 42.5 Pa.

Mathadon:
good to know! do you have a link to `ASHRAE Dampers and Air Flow Control`?

AntoineGautier:
Out of github because I am not sure about licensing issues on this one
https://uservice.ir/upload/download/Catalog/Standard/Dampers%20and%20airflow%20control%20ASHRAE.pdf

AntoineGautier:
Close as it is now addressed in branch [issue1188_dampersRefactor](https://github.com/ibpsa/modelica-ibpsa/tree/issue1188_dampersRefactor). 

(Important parts of the discussion have been added to the corresponding issue.)

PR #1193, opened 2019-08-20
Issue1192 time documentation

### LINKED ISSUE #1192: time convention in Modelica
Once more, I'd like to open an issue about the meaning of the variable `time` in our simulations, since it keeps confusing me and I have the feeling we can document it better. The way I understand it now, we can summarize the convention as follows:

`time` is the index variable of the weather file, which by convention starts with data for midnight January 1st _**in the local time zone**_ at `time=0`. `timZon` from the weather file shifts this `time` such that the solar position is correct in the solar irradiation computations. And that's pretty much all there's to it.

I think this should be clearly mentioned in the documentation of `ReaderTMY3` and `CalendarTime`. 

With respect to `CalendarTime` I think we actually have a bug on our hands since the unix time stamp is defined as `It is the number of seconds that have elapsed since 00:00:00 Thursday, 1 January 1970,[2] Coordinated Universal Time (UTC), minus leap seconds` (Wikipedia) whereas we use the _**local**_ time instead of UTC to compute the hour, day, etc in `CalendarTime`. Using a time stamp without specifying the time zone thus seems incorrect, although it is more intuitive for someone who hasn't had to worry about time zones before. I propose nonetheless to revise the implementation and to force the user to provide a time zone when using unix time stamps.

If all agree then I will propose an implementation.

Mathadon:
this closes #1192


mwetter:
@Mathadon : Please merge if you agree with my revisions to the documentation. I removed
```
This option should thus not be combined with
<a href=\"modelica://IBPSA.BoundaryConditions.WeatherData.ReaderTMY3\">
IBPSA.BoundaryConditions.WeatherData.ReaderTMY3</a>.
```
because I think the two models can be used together because the data reader does not use the output of the calendar time.

I also removed the "expert option" qualifier as I think GMT suffices (and it is not clear what one needs to be an expert in).

Mathadon:
I don't think they should be used together since the `zerTim=UnixTimeStampGMT` assumes that `time` is in a different time zone for `CalendarTime` than `ReaderTMY3`. E.g. for US based locations when the sun is at its highest point (noon) the calendar time will indicate that it's ~ 4 AM?

Mathadon:
Both models use the built-in variable `time` so what I mean is that a model contains an instance of both `CalendarTime` and `ReaderTMY3`.

Mathadon:
@mwetter see 
2733edf

Issue #1206, opened 2019-09-13
Restructuring Media.Examples.BaseClasses to avoid duplicated codes for Water, Air, and Steam


YangyangFu:
Steam model is to be added to Modelica Buildings library. Base testing examples `Media.Examples.BaseClasses` need to be restructured in order to avoid duplicated codes.

YangyangFu:
@mwetter Dr. Wetter, can you assign this me, and make a new dev branch so that i can make a pull request?

mwetter:
@YangyangFu : Please make a pull request against the branch `issue1206_media_steam` when you implemented the media.

mwetter:
The latest code is at `issue1206_media_steam` and needs to be reviewed.

PR #1208, opened 2019-09-17
Issue1206 media steam

YangyangFu:
Update steam model, which is able to model phase changing between liquid and vapor.

This is for issue #1206.


mwetter:
The code is at https://github.com/ibpsa/modelica-ibpsa/tree/issue1206_media_steam for further review and revisions.
I will close this PR as it is against the master rather than the development branch.

PR #1211, opened 2019-09-26
Added test for null pointer

### LINKED ISSUE #1210: C-Sources/cryptographicsHash.c does not guard against failure of malloc
C-Sources/cryptographicsHash.c calls `malloc` without testing the value of pointer. A test of the form 
```c
if (!ptr)
  ModelicaError(...);
```
needs to be added.

mwetter:
This closes #1210


mwetter:
@MassimoCimmino : Are you OK with this change?

MassimoCimmino:
Sorry for the slow response. I agree with the changes proposed in the issue. I have tested the geothermal model and everything works fine. I am unable to comment on the implementation as I am not that well versed in C.

Issue #1209, opened 2019-09-19
Enhance ROM (ThermalZone.RC) for moisture balance


KremerMartin:
**What do I want to do?**
I want to implement the possibility to take into account moisture production and moisture balance in thermal zones. 
For this reason I would like to modify the air volume model in the `ThermalZones.ReduceOrder.RC.OneElement.mo`.

**How I want to solve it**

- I will implement a boolean parameter `use_moisture_balance` that will be _false_ per default.

- The `volAir` will be conditioned with this parameter. If it is _false_ the existing `volAir` will be used.

- A second air volume `volMoistAir` will be implemented (type: `Fluid.MixingVolumes.MixingVolumeMoistAir`) and conditioned by the same parameter (if _true_ use `volMoistAir`)

- An additional input connector `mWat_flow` and output connector `X_w` both conditioned with `use_moisture_balance` will be implemented and conditionally connected to the `volMoistAir`

In the Aixlib this implementation has been performed. To avoid Merge-Conflicts with the IBPSA, I would like to implement the same changes into the IBPSA. Changes can be seen in issue [695](https://github.com/RWTH-EBC/AixLib/issues/695) of the Aixlib.

Are there any concerns about this changes?

KremerMartin:
@mwetter Could you assign me to this issue and give me rights to push a new branch where I can solve this issue, if you approve the change?

mwetter:
@KremerMartin : I made the branch `issue1209_thermalZoneMoistAir` for this development.

KremerMartin:
@mwetter Thanks a lot. Can you give me permission to push to this branch? Until now I have no access.

mwetter:
@KremerMartin : you should still be able to make a pull request against the branch.

KremerMartin:
This issue has been solved by PR #1213. The branch is now ready to be merged into the master.

mwetter:
@KremerMartin : Thanks for the PR. I am working on reviewing and revising it.

mwetter:
@KremerMartin : I revised the code and added the following:

- Corrected the implementation of the latent gain. The earlier version only added moisture to the volume, but this would have only changed the mass fraction in the volume and not the internal energy. Hence, it would have adiabatically cooled the air. 
[...]

PR #1213, opened 2019-10-07

KremerMartin:
This should solve the issue #1209 . As the default value for the boolean variable `use_moisture_balance` is false, there is no influence on the results of the unit tests.


KremerMartin:
@mwetter As I see it, the check fails for dymola due to a missing docker account. Is there anything I can do about it?

Mathadon:
@KremerMartin we have to recreate the pull request to have the required write rights on the repo. There is nothing you can do to resolve that. :)

PR #1215, opened 2019-10-10

KremerMartin:
@mwetter Thanks for your revision. Here is the update of the `VAir` parameter. It might interest you, that the room volume is declared the same for all test cases of the VDI6007.


KremerMartin:
For #1209

KremerMartin:
Sorry, I missed the change for the second test case. I will add that.

PR #1216, opened 2019-10-10

KremerMartin:
For #1209


Mathadon:
reopening for unit tests

PR #1218, opened 2019-10-14

### LINKED ISSUE #1209: Enhance ROM (ThermalZone.RC) for moisture balance

mwetter:
This closes #1209.


mwetter:
@Mathadon : Can you please do a last review and then merge.

IBPSA/ThermalZones/ReducedOrder/RC/OneElement.mo
```diff
@@ -129,8 +139,25 @@ model OneElement "Thermal Zone with one element for exterior walls"
     final C_start=C_start,
     final C_nominal=C_nominal,
     final mSenFac=mSenFac,
-    final use_C_flow=false) if VAir > 0 "Indoor air volume"
-    annotation (Placement(transformation(extent={{38,-10},{18,10}})));
+    final use_C_flow=false) if VAir > 0 and not use_moisture_balance
+    "Indoor air volume"
+    annotation (Placement(transformation(extent={{42,-26},{22,-6}})));
+  Fluid.MixingVolumes.MixingVolumeMoistAir volMoiAir(
```
Mathadon: This makes me wonder why we have a separate model for `MixingVolume` with and without moisture..?

IBPSA/ThermalZones/ReducedOrder/RC/OneElement.mo
```diff
@@ -233,12 +262,29 @@ protected
...
     annotation (Placement(transformation(extent={{-186,118},{-174,130}})));
 
+  Modelica.Blocks.Math.Gain mWat_flow(
+    final k(unit="kg/J") = 1/h_fg,
+    u(final unit="W"),
+    y(final unit="kg/s")) if
+       use_moisture_balance and ATot >0 "Water flow rate due to latent heat gain"
+    annotation (Placement(transformation(extent={{-200,-100},{-180,-80}})));
+
+  Buildings.HeatTransfer.Sources.PrescribedHeatFlow conQLat_flow if
```
Mathadon: needs IBPSA counterpart?

Mathadon:
Indeed, unit tests fail. In case this is of interest, we can add 
```
after_script:
  - cat IBPSA/failed-simulator-dymola.log
```
to `.travis.yml`, which would allow to debug the CI output ` Simulation failed for 'RunScript("Resources/Scripts/Dymola/ThermalZones/ReducedOrder/Validation/RoomWithLatentGain.mos");'` etc more easily.

Mathadon:
Btw, I'd be in favor of adding `Buildings.HeatTransfer.Sources.PrescribedHeatFlow` to IBPSA since it has better algebraic-loop-ish properties.

IBPSA/ThermalZones/ReducedOrder/RC/OneElement.mo
mwetter: I wonder too.

IBPSA/ThermalZones/ReducedOrder/RC/OneElement.mo
mwetter: Yes. Let's let travis fail, then I can add the `after_script` to see if it does the right thing, and then I can fix it.

mwetter:
> Btw, I'd be in favor of adding `Buildings.HeatTransfer.Sources.PrescribedHeatFlow` to IBPSA since it has better algebraic-loop-ish properties. In which case it would make sense to replace its body by
> 
> ```
> extends Modelica.Thermal.HeatTransfer.Sources.PrescribedHeatFlow(final alpha=0)
> ```
> 
> to be a bit more didactic than the current implementation. Although it would require more translation overhead..

We could add `Buildings.HeatTransfer.Sources` to `IBPSA` if there is a consensus.

PR #1222, opened 2019-10-19
disabled homotopy for #1221

### LINKED ISSUE #1221: Unbounded limiter in MSL 3.2.3 causes negative phi
Since MSL 3.2.3 the model `Modelica.Blocks.Nonlinear.Limiter` has the parameter `homotopyType`, which sets a homotopy option. By default `linear` is used, which neglects the min/max bounds. When updating IDEAS to Dymola 2020 this caused the PI controller output bounds to be ignored at initialization, leading to a PI outlet control signal of -10, which then fed into a valve and caused the resulting algebraic loop not to converge due to the negative control signal. I propose to disable homotopy by default in `IBPSA.Controls.Continuous.LimPID` to avoid this.

Mathadon:
closes #1221


Mathadon:
@mwetter Only now I realize that only the JModelica unit tests are failing. Any chance that they are not yet using the latest version of MSL that includes the homotopy option?

mwetter:
@Mathadon : It looks like JModelica is not yet using the latest MSL that has the homotopy option.

Mathadon:
Not sure how to best approach that? It would make sense to use the same MSL version for the testing of the library. Alternatively we can duplicate the implementation of the model that was changed in the latest MSL release.

mwetter:
@Mathadon : The current JModelica trunk uses MSL 3.2.2, as does OMEdit, but the new limiter is from 3.2.3.
I think best would be to make the new limiter a protected class in LimPID and then use this until JModelica and OpenModelica updates to 3.2.3.

Mathadon:
@mwetter done. I did not exactly copy the implementation since this would require me to also copy `Modelica.Blocks.Types.HomotopyType`, which does not exist in MSL 3.2.2. Instead I stripped the `homotopyType` option from the implementation of MSL 3.2.3 and hardcoded the code for option `NoHomotopy`.

PR #1224, opened 2019-10-19
Issue1198 hydraulic check valve

### LINKED ISSUE #1198: Discuss CheckValve implementation
This issue discusses the implementation of a CheckValve Model.

Mathadon:
closes #1198


Mathadon:
@Kristoff6 I have revised the documentation, changed line endings, added a valve with series restriction in the example and added unit test and reference results. I also revised the check valve saturation such that the k value is maximum at 10% of dp_nominal. In case you have comments?

Mathadon:
@mwetter this is almost ready to be reviewed. The only thing that I am not sure about is how to parameterize the valve. I.e. we should clearly specify what values should be used for `dp_nominal` and `m_flow_nominal`. My intuition is that `dp_nominal` should be the pressure at which the valve is fully opened, i.e. when `k` saturates. This is however a relatively low pressure probably and chances are that users will set the pressure much higher. So I'm not sure how to document this best. Any suggestions here?

IBPSA/Fluid/FixedResistances/CheckValve.mo
```diff
@@ -0,0 +1,120 @@
+within IBPSA.Fluid.FixedResistances;
+model CheckValve "Check valve that avoids flow reversal"
+  extends IBPSA.Fluid.BaseClasses.PartialResistance(
+    dp(nominal=6000),
+    final dp_nominal=dpValve_nominal + dpFixed_nominal,
+    final m_flow_turbulent=deltaM*abs(m_flow_nominal),
+    final from_dp=true,
+    final linearized=false,
+    allowFlowReversal=false);
```
Mathadon:
```suggestion
    allowFlowReversal=true);
```

allowsFlowReversal should be `true` since the flow does become slightly negative.

Mathadon:
@mwetter can you check my question? thanks!

mwetter:
@Mathadon . `dpValve_nominal` (or `Kv`) should be for fully open valve. This would be consistent with for example https://ch.buildup.group/product/c0ffee81-eceb-7503-fc58-3d8bfd7956a2/view#-485251046 which specifies a Kvs value.

Mathadon:
@Mathadon to check comment Michael, remove `inline=True`, add typical pressure drop to documentation, put equations in equation section.

Mathadon:
This second source gives similar Kvs values:
https://imihydronic.blob.core.windows.net/resources/Documents/Catalogues/English_International/PDF_low/TA-NRV_F_EN_low.pdf 
Schermafbeelding 2019-11-09 om 09 29 52

DN50 at 1m/s is 7m3/h or 2 kg/s, which converts into a nominal pressure drop (Kv=36.6) of 0.036 bar or 3.6 kPa. I'll add that as a typical value in the documentation.

Mathadon:
See the flow characteristic:

image


It has the behavior that was mentioned during the call that the extrapolated curve extends onto a fixed pressure drop rather than the origin. To be honest this is more by chance than by design. I.e. the used HeaviSide function causes that behavior. Near the origin the pressure drop suddenly decreases, which is good since the curve has to go through the origin. 
In practice, as for instance seen in the data sheet that I linked, check valves do not necessarily operate based on a spring. So I wouldn't even agree that the behavior is really a requirement. So for me the used flow curve is good.

I have revised the implementation by adding a parameter `dpValve_closed`, which allows the user to specify when the valve starts to close, i.e. when the flow coefficient starts to change. Default is `dpValve_closed=dpValve_nominal/100`. @Kristoff6 is that ok for you?

Mathadon:
Strange, I get:

```
Differentiating (if noEvent(abs(dp) > dp_turbulent) then noEvent((if dp > 0 then 1 else (if dp < 0 then -1 else 0)))*k*sqrt(abs(dp)) else (1.40625+(0.15625*dpNormSq-0.5625) *dpNormSq)*m_flow_turbulent*dpNorm)
```
when prescribing the flow rate. For other models, Dymola does not generate the warning and it inverses `IBPSA.Fluid.BaseClasses.FlowModels.basicFlowFunction_dp` when needed such that no algebraic loop is formed. For the `CheckValve` Dymola can't use the inverse since `k` depends on `dp`. A warning is generated, which can be avoided by adding `smooth(2,.)` in `IBPSA.Fluid.BaseClasses.FlowModels.basicFlowFunction_dp`. Let's add this for now and report this as a bug? See also #1240 .

AntoineGautier:
I followed up our discussion on the call by reading some documentation from manufacturers e.g. _Crane Valves and Fittings Catalog 60 §Proper Sizing of Check and Foot Valves_. It appears that:

- the minimum pressure drop for **full opening** does not depend primarily on the size but rather on the design of the valve,
- typical values are 0.25 psi (1.7 kPa) for horizontal in-line ball, 0.5 psi (3.4 kPa) for swing valves and 2 psi (13.8 kPa) for lift valves,
- the valve should be sized to be fully open in standard conditions,
- this usually leads to a **smaller size** than the pipe e.g. for 10 m3/h nominal flow rate, the required pipe size is DN65 and considering 3.4 kPa to size the valve leads to Kvs=53.9 so indeed between DN50 and DN65 from the data above. However it is hard to believe that HVAC contractors would select smaller valves and use reducing flanges. So most likely the check valves are often oversized i.e. only partially open at nominal flow rate.

Eventually it would make sense to me in the model:

- to specify the default value for `dpValve_closed` e.g. 3.4 kPa,
- and use `dpValve_nominal=dpValve_closed`.
- (By the way I am confused about what `dpValve_closed` stands for exactly. If it is indeed the minimum pressure drop for full opening as I understand it should rather be named `dpValve_open` or `dpValve_closing` no?)

Mathadon:
@AntoineGautier thanks for looking this up! That's some useful information. I agree that `dpValve_closing` would be a more appropriate name. 
I'm surprised by the high pressure drop before the valve is fully open, but it seems consistent with what you say, that the valve should have a relatively small diameter in order to be fully open. If that's true then I propose to set `dpValve_nominal=4000` and `dpValve_closing=dpValve_nominal/2`. Do you have the original source for this information? Then I add it in the documentation.

Mathadon:
@mwetter this is ready for review

mwetter:
@Mathadon : Please go ahead and merge it if you agree with my revision, mainly in the info section of the model.

PR #1242, opened 2019-11-09
added phi=max(y_actual,0) for all two way valves

### LINKED ISSUE #1223: bound inputs of valves
linked to #1221 : shouldn't we bound the inputs of valves to avoid non-physical situations? We have an assert now but it's set at `-0.2`, which is still quite large. I think it would be a better solution to guard the inputs using a `phi=max(l, ...)` expression?

Mathadon:
removed assert that checks phi > -0.2
closes #1223


Mathadon:
I have revised the implementation such that `phi` never becomes negative, instead of `y_actual`. This has the advantage that the bound is not active within the normal operating region of the valve.

PR #1245, opened 2019-11-13
Issue1227 three way valve

### LINKED ISSUE #1227: ThreeWayTable Valve
The library contains a two way valve that has a table based flow characteristic (TwoWayTable). However, a three way valve with table-based characteristics does not exist. I tried to model a three way table-based valve in the same way as the model ThreeWayLinear or ThreeWayEqualPercentageLinear: I tried to redeclare the partial two way valve (res1 and res3) and use the TwoWayTable. However, this approach does not work because the parameter l in the partial valve (res1 and res3) is final as well as the parameter l in the model TwoWayTable.

I think it would be a good approach to release the "final" modifier in the PartialThreeWayValve because this model is a base class. The final modifier could be added in the child classes if needed.
Are there any suggestions how to solve this issue?

Best regards

mwetter:
This closes #1227.

@alexanderAKU: It would be nice if you could provide the table based three-way valve to the IBPSA library, assuming it is similar to `TwoWayTable`.

### DIFF
--- a/IBPSA/Fluid/Actuators/BaseClasses/PartialThreeWayValve.mo
+++ b/IBPSA/Fluid/Actuators/BaseClasses/PartialThreeWayValve.mo
@@ -9,7 +9,6 @@ partial model PartialThreeWayValve "Partial three way valve"
           deltaM=deltaM,
           dp(start=dpValve_nominal/2),
           from_dp=from_dp,
-          final l=l[1],
           final linearized=linearized[1],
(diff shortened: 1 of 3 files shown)


alexanderAKU:
@mwetter shall I open a new issue and branch for the model + example or could I use the existing ones?

mwetter:
@alexanderAKU : Better use a new branch. I will merge this to the master.

PR #1271, opened 2019-12-24
Removed defaultComponentName for #1270

### LINKED ISSUE #1270: Media package defines defaultComponentName which is not needed
Some media models declare
```
defaultComponentName="gas"
```
There is no point in this declaration as it is a package and hence cannot be dragged into the schematic view of a model. Therefore, this can be removed.

mwetter:
This closes #1270


mwetter:
@Mathadon : Can you please merge if you agree with the change. Thanks.

Mathadon:
There's no revision history not but I guess it does not matter that much in this case.

PR #1300, opened 2020-02-28
Issue1297 scale coeff1

### LINKED ISSUE #1297: Scale coeff1 with damper nominal pressure in PressureIndependent
PR [#1267](https://github.com/ibpsa/modelica-ibpsa/pull/1267) changed what `dp_nominal` stands for. As described in issue [#1188](https://github.com/ibpsa/modelica-ibpsa/issues/1188), it was previously the pressure drop of the damper alone whereas it is now the pressure drop of the damper plus the optional fixed resistance. Since `coeff1` is computed from `dp_nominal`, the resulting change in the order of magnitude can cause some convergence issues. This issue is to assign:
```
coeff1 = l2 / dpDamper_nominal * m_flow_nominal
```

AntoineGautier:
This closes #1297.
I also added 0 as a default value for the fixed resistance of the damper model in `MixingBox`.
This is ready to merge.


mwetter:
These changes look good to me.
@Mathadon : Can you please review so we have a non-LBL reviewer.

PR #1312, opened 2020-03-07
Issue1310 rpipe

### LINKED ISSUE #1310: Inaccuracy calculation thermal resistance plug flow pipe model
Currently the thermal resistance is calculated as follows:

```modelica
parameter Real R=1/(kIns*2*Modelica.Constants.pi/Modelica.Math.log((dh/2 + dIns)/(dh/2)))
```

which incorrectly considers the pipe thickness. The insulation starts at radius dh/2+thickness instead of dh/2. The calculation should therefore be as follows according to me:
```modelica
parameter Real R=1/(kIns*2*Modelica.Constants.pi/Modelica.Math.log((dh/2 + thickness + dIns)/(dh/2 + thickness)))
```

Mathadon:
closes #1310 
@mwetter can you review this?


Mathadon:
@mwetter I did not think about updating reference results. Thanks for fixing that for us!

PR #1313, opened 2020-03-08
Issue1289 pipe data

### LINKED ISSUE #1289: refactor data entry for Validation.PlugFlowPipes
The examples in `IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes` define large data records with a `CombiTimeTable` in the form
```modelica
Modelica.Blocks.Sources.CombiTimeTable combiTimeTable(table=pipeDataToBeRead.data)
```
In OPTIMICA, this leads to a `modelDescription.xml` file that is around 230 MB large because each data entry becomes a parameter that the user can change.  That in turns slows down the compilation and leads to very slow loading of the FMU. This issue is to correct this.

mwetter:
This closes #1289


Mathadon:
@mwetter I have reviews the code. All looks good.

bramvdh91:
Couldn't check if the results remain the same, but as far as I can tell from the code this is okay.

mwetter:
@bramvdh91 : Thanks for the review. The results remain the same.

PR #1337, opened 2020-04-06
added flow direction indicators to PartialFourPort for #1336

### LINKED ISSUE #1336: flow direction indicators in PartialFourPort
To be added.

Mathadon:
closes #1336


mwetter:
@Mathadon These graphical elements clash with input and output signals. 
image
Maybe it is better to add icons on the line between the ports so it is out of the way of hopefully all other graphical elements.

Mathadon:
@mwetter fair enough! I'll make them a bit larger then to ensure that they are visible.

Mathadon:
Not working:
image

Mathadon:
Also not that great: 
image

Mathadon:
Compromise?

image

mwetter:
I prefer https://github.com/ibpsa/modelica-ibpsa/pull/1337#issuecomment-610369687 as everything is inside the icon. This detail need not be prominent in my view. Your version "Compromise?"  has icons where we often have input and output connectors, so this won't work. See the branch `issue1336_iconFourport_v2` for a cleaned up version.

Mathadon:
> I prefer [#1337 (comment)](https://github.com/ibpsa/modelica-ibpsa/pull/1337#issuecomment-610369687) as everything is inside the icon. This detail need not be prominent in my view. Your version "Compromise?" has icons where we often have input and output connectors, so this won't work. See the branch `issue1336_iconFourport_v2` for a cleaned up version.

Isn't it better to have icons sometimes overlap with the blue arrows rather than the arrows always being underneath the other bits of the icon?

mwetter:
I think the arrows are visible enough if they are below the line, but think it is confusing for users if they look like a RealInput hanging off the side of the icon.

PMehrfeld:
Looks like a good graphical enhancement.  
One comment / suggestion from my side would be, to also add "1" and "2" near the ports or arrows. If not even "a1", "b1", "a2", "b2".  
Often people have to check which side is 1 and which is 2.  
This is just a minor point, I would like to put into this discussion. If these extra Strings lead to unclear icon, we should let it be.

Mathadon:
> Looks like a good graphical enhancement.
> One comment / suggestion from my side would be, to also add "1" and "2" near the ports or arrows. If not even "a1", "b1", "a2", "b2".
> Often people have to check which side is 1 and which is 2.
> This is just a minor point, I would like to put into this discussion. If these extra Strings lead to unclear icon, we should let it be.

I think it would overload the icon a bit. The extra arrows do so too, but they're optional so it's not as bad then. People who use `allowFlowReversal` should care enough not to be bothered about it.

mwetter:
Shall we go with https://github.com/ibpsa/modelica-ibpsa/pull/1337#issuecomment-610369687 (which @Mathadon labelled as "not working")? I don't like 
https://github.com/ibpsa/modelica-ibpsa/pull/1337#issuecomment-610370756 as this space is used for actual connectors in some models (see IBPSA.Fluid.HeatExchangers.DryCoilEffectivenessNTU) and it makes me want to connect some connectors to the arrows that hang off the icon as if they were input or output connectors.

Mathadon:
How about this? 
image

mwetter:
I suggest to align the graphic with the grid and simply show the arrow at the outlet as it is clear that it applies to inlet and outlet. The icon from 421906741 is

image

Mathadon:
ok!

PR #1338, opened 2020-04-06
Issue1188 damper opening

AntoineGautier:
This is the PR 2 out of 3 for the issue #1188. 
This one addresses the computation of the damper opening in `PressureIndependent`.
There is a related issue in MBL [#1873](https://github.com/lbl-srg/modelica-buildings/issues/1873) to update all the VAV models using `PressureIndependent` so that the implemented control logic be consistent.
@Mathadon Would you please consider having a first look at the implementation and the validation model `IBPSA.Fluid.Actuators.Dampers.Validation.PressureIndependent`? But note that I would like to address (at least partially) [#1873](https://github.com/lbl-srg/modelica-buildings/issues/1873) so that the implementation be tested thoroughly before merging.


Mathadon:
@AntoineGautier I'll look into it but might take a while since it's a lot of code.

IBPSA/Fluid/Actuators/BaseClasses/exponentialDamper_inv.mo
```diff
@@ -0,0 +1,57 @@
...
+      y1d=invSplDer[i - 1],
+      y2d=invSplDer[i]),
+    0,
+    1,
+    1E-3);
+annotation (smoothOrder=1,
+Documentation(info="<html>
+<p>
+This function provides an approximate inverse of the exponential damper characteristics.
+</p><p>
```
Mathadon, line 37:
```suggestion
</p>
<p>
```

IBPSA/Fluid/Actuators/BaseClasses/exponentialDamper_inv.mo
```diff
@@ -0,0 +1,57 @@
...
+The function is used by the model
+<a href=\"modelica://IBPSA.Fluid.Actuators.Dampers.PressureIndependent\">
+IBPSA.Fluid.Actuators.Dampers.PressureIndependent</a>.
+</p><p>
```
Mathadon, line 41:
```suggestion
</p>
<p>
```

IBPSA/Fluid/Actuators/BaseClasses/exponentialDamper_inv.mo
```diff
@@ -0,0 +1,57 @@
...
+      y2=ySupSpl[i],
+      y1d=invSplDer[i - 1],
+      y2d=invSplDer[i]),
+    0,
+    1,
+    1E-3);
+annotation (smoothOrder=1,
+Documentation(info="<html>
+<p>
+This function provides an approximate inverse of the exponential damper characteristics.
```
Mathadon, line 36: I would add a sentence like 'This function thus computes the valve opening that corresponds to some loss coefficient.'

IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo
```diff
@@ -3,15 +3,39 @@ model PressureIndependent
...
+  parameter Real y_min = 2E-2
+    "Minimum value of control signal before zeroing of the opening";
+  parameter Integer sizeSupSplBnd = 5
+    "Number of support points on each quadratic domain for spline interpolation";
+  parameter Integer sizeSupSpl = 2 * sizeSupSplBnd + 3
+    "Total number of support points for spline interpolation";
+  parameter Real[sizeSupSpl] ySupSpl_raw = cat(
+    1,
+    linspace(1, yU, sizeSupSplBnd),
+    {yU-1/3*(yU-yL), (yU+yL)/2, yU-2/3*(yU-yL)},
```
Mathadon, line 25: can we document that rationale for this implementation somewhere?

IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo
```diff
@@ -125,8 +155,22 @@ equation
...
+    dp=dp,
+    m_flow_small=1E-2*abs(m_flow_nominal),
+    dp_small=1E-4*dp_nominal_pos);
+  kDam = if dpFixed_nominal > Modelica.Constants.eps then
+    sqrt(1 / (1 / k^2 - 1 / kFixed^2)) else k;
+  // Use of regStep might no longer be needed when the leakage flow modeling is updated.
+  y_actual_smooth = IBPSA.Utilities.Math.Functions.regStep(
+    x=y_internal - y_min,
+    y1=IBPSA.Fluid.Actuators.BaseClasses.exponentialDamper_inv(
+      kThetaSqRt=sqrt(2*rho)*A/kDam, kSupSpl=kSupSpl, ySupSpl=ySupSpl, invSplDer=invSplDer),
```
Mathadon, line 170: I'm confused by this implementation, specifically the first argument: that input requires something with units of 'k' (judging from the name), but you assign something with units rho*A/k, which seems wrong?

IBPSA/Fluid/BaseClasses/FlowModels/basicFlowFunction_inv.mo
```diff
@@ -0,0 +1,43 @@
+within IBPSA.Fluid.BaseClasses.FlowModels;
```
Mathadon, line 1: I think that we can add this function in the inverse annotation of `basicFlowFunction_m_flow` and `basicFlowFunction_dp`. In that sense, a more consistent naming would be `basicFlowFunction_k`.
Vice versa, `basicFlowFunction_m_flow` and `basicFlowFunction_dp` can be added in the inverse annotation of this function. This would however require the model to be refactored to have `m_flow_turbulent` as an input instead of `m_flow_small`. For I consistency reasons I believe that it is worth considering this change.

IBPSA/Resources/Scripts/Dymola/Fluid/Actuators/Dampers/Validation/PressureIndependent.mos
```diff
@@ -0,0 +1,22 @@
...
+ color = true);
+createPlot(id=1, y={"damPreInd.y", "damPreInd.y_actual"},
+ position={73, 9, 598, 390},
+ autoscale = true,
+ autoerase = true,
+ autoreplot = true,
+ grid = true,
+ color = true,
+ subPlot=2);
+plotExpression(apply(PressureIndependent[end].senRelPre.p_rel/PressureIndependent[end].dp_nominal),
```
Mathadon, line 20: Does it work to use `[end]` in these scripts? Might want to reconsider this formulation if it was not intentional.

IBPSA/Fluid/Actuators/BaseClasses/exponentialDamper_inv.mo
AntoineGautier: Addressed

IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo
AntoineGautier: Addressed

IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo
AntoineGautier: `exponentialDamper_inv` uses the square root of the loss coefficient (similar to `exponentialDamper`), `kThetaSqRt = sqrt(dp / (1/2 * rho * v^2))`. The first argument is the inline computation of this value from the flow coefficient `k = m_flow / sqrt(dp)`.

IBPSA/Fluid/BaseClasses/FlowModels/basicFlowFunction_inv.mo
AntoineGautier: I wonder about the renaming because `basicFlowFunction_m_flow` outputs `dp`, not `m_flow`. So the name of the existing flow functions refers to the input variable, not to the output variable. With this logic the function should be renamed `basicFlowFunction_m_flow_dp`.

About the inverse, please note that currently the function only represents the turbulent regime. This is because it is the regime where the actual opening matters for any control sequence that would use it (also because the condition to transition to laminar flow is debatable for dampers and valves as it does not take into account the variation of the free area). 

Representing the laminar regime would lead to something like
```
k := smooth(2, if m_flow > m_flow_turbulent then
    mPos_flow / dpSqrt else
    0.866025403784439 * sqrt(-mPos_flow * (1/6 * mPos_flow^4 -
    mPos_flow^2 * m_flow_turbulent^2 - 0.5 * m_flow_turbulent^4)) /
    dpSqrt / sqrt(m_flow_turbulent^3));
```
Please let me know if this is worth considering.

IBPSA/Resources/Scripts/Dymola/Fluid/Actuators/Dampers/Validation/PressureIndependent.mos
AntoineGautier: It works but it also works without so I removed it.

IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo
AntoineGautier: @mwetter You are right that for this model it seems better suited to use the actual value of the loss coefficient in `exponentialDamper_inv` and the square value of the flow coefficient in `basicFlowFunction_inv`. However 

- an equation for the actual value of the flow coefficient would still need to be included as the variable is declared in the base class (although it might not require the same smoothness),
- this would introduce some ambiguity on the quantity of the I/O variables between these new  functions and the original functions that they inverse.

AntoineGautier:
@Mathadon Thanks a lot for your feedback. I addressed your comments with my last commits. Still an open discussion regarding the inverse annotation.
Also please consider that I want to wait for the impact study on the VAV models in the Buildings library before merging this PR.

IBPSA/Fluid/BaseClasses/FlowModels/basicFlowFunction_inv.mo
Mathadon: Indeed, there are some trade-offs to be considered here. @mwetter what are your thoughts on this?

mwetter: The transitioning to laminar -- which is really in our implementation simply a regularization that ensures that the derivative is bounded near zero -- is needed only for numerical reasons. If for this model you can guarantee that the derivative is bounded then we won't need this transition as laminar flow through a damper is not likely happening.

I think clearest would be to call this function `basicFlowFunction_m_flow_dp` as this denotes its input arguments.

AntoineGautier:
@Mathadon The impact of this branch developments on the CPU time for simulation has been evaluated on the VAV models of MBL [here](https://github.com/lbl-srg/modelica-buildings/pull/1918#issuecomment-622605572). It appears marginal. 
In addition I refactored the two inverse functions to simplify the computation scheme (without noticing any significant improvement on the computation time though). The drawback is that their I/O variables now differ from the original functions. Since I think they might not be used outside of this model, I decided to include them directly in the protected section.

If you agree with these last changes, then this is ready to merge when all tests pass.

PR #1360, opened 2020-05-02
Issue1359 weather data interpolation

### LINKED ISSUE #1359: Weather Data Sub hourly interpolation
The IBPSA library default option for weather data sub hourly interpolation is the spline approach. While conducting the analysis for the weather bestest <a href="https://github.com/ibpsa/modelica-ibpsa/issues/1314">#1314</a>, we noticed that when in the weather file two consecutive hourly values were identical, the spline interpolation would lead to an error with respect to the exact interpolation. Therefore this simplified yearly analysis was carried out to compare the Spline, the Fritisch-Butland and the Steffen inteporlation models. The results are shown in the charts below for the weatherdata carried by the weatherbus in terms of % of hourly steps for which there are almost identical successive values, and the yearlt % error with respect to the mean of those two identical values, in the table instead are reported the absolute yearly errors. I also attach a .zip file with the sample .csv output files and the Python script I used to generate the charts.
Spline
Fritsch-Butland
Steffen

image

But taking a look at the results it could be a good option to change the default interpolation method from Spline to the Fritich-Butland.

[WeatherInterpolation.zip](https://github.com/ibpsa/modelica-ibpsa/files/4565370/WeatherInterpolation.zip)

mwetter:
This close #1359 
See issue for discussion.


mwetter:
This won't be merged. See discussion at https://github.com/ibpsa/modelica-ibpsa/issues/1359

PR #1368, opened 2020-06-01
Issue1365 pid reverse action

### LINKED ISSUE #1365: reverse action in LimPID controller
Currently, the LimPID controller uses the wrong convention for reverse action. They use the following where `u_m` is the measurement signal and `u_s` is the set point signal:

> If the parameter `reverseAction=false` (the default), then `u_m < u_s` increases the controller output, otherwise the controller output is decreased. Thus,
>  - for a heating coil with a two-way valve, set `reverseAction = false`,
>  - for a cooling coils with a two-way valve, set `reverseAction = true`. 
> This was done on the wrong understanding that reverse action relates to the control error (`e = u_s - u_m`), and not to the measurement signal. Because the measurement signal is negative in `e = u_s - u_m`, the convention of `reverseAction` is backwards to what the HVAC (and other industries?) are using.
> 

We need to fix this, in a way that allows an update with a conversion script so that existing models don't break and keep operating correctly, e.g., we need to use a new name for this parameter. We could change this to

> Through the parameter <code>reverseActing</code>, the controller can be configured to
> be reverse or direct acting.
> The standard form is reverse acting, which is the default configuration.
> For a reverse acting controller, for a constant set point,
> an increase in measurement signal <code>u_m</code> decreases the control output signal <code>y</code>
> (Montgomery and McDowall, 2008).
> Thus,
> </p>
> <ul>
>   <li>
>   for a heating coil with a two-way valve, leave <code>reverseActing = true</code>, but
>   </li>
>   <li>
>   for a cooling coil with a two-way valve, set <code>reverseActing = false</code>.
>   </li>
> </ul>

The code that will need to be changed is
```modelica
  final parameter Real revAct = if reverseAction then -1 else 1
    "Switch for sign for reverse action";
```
There also need to be a conversion script.

This revision also changes the default configuration from PID to PI.

mwetter:
This closes #1365

This changes `reverseAction` to `reverseActing`, and corrects the meaning of this parameter to comply with the convention used in the HVAC industry.

This also changes the default from PID to PI.

This changes is for all three PID controllers in `IBPSA.Controls.Continuous`.

Todo: 
- [ ] Add conversion scripts

### DIFF
--- a/IBPSA/Controls/Continuous/Examples/LimPID.mo
+++ b/IBPSA/Controls/Continuous/Examples/LimPID.mo
@@ -14,7 +14,7 @@ model LimPID "Test model for PID controller with optional reverse action"
           annotation (Placement(transformation(extent={{-20,30},{0,50}})));
   IBPSA.Controls.Continuous.LimPID limPIDRev(
     controllerType=Modelica.Blocks.Types.SimpleController.PID,
-    reverseAction=true,
+    reverseActing=false,
     Ti=1,
     Td=1,
     yMax=1,
--- a/IBPSA/Controls/Continuous/Examples/LimPIDWithReset.mo
+++ b/IBPSA/Controls/Continuous/Examples/LimPIDWithReset.mo
@@ -130,7 +130,6 @@ First implementation.
       final reset=reset,
       yMax=1,
       yMin=0,
-      controllerType=Modelica.Blocks.Types.SimpleController.PI,
       Ti=1,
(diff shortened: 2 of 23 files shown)


mwetter:
@Mathadon @PMehrfeld @nytschgeusen : Can you please review and if no changes requested approve this backward incompatible change. This is the change we discussed on the last coordination call.
@dhblum also reviewed it on the LBL side.

Issue #1389, opened 2020-09-01
Refactored Steam Medium 


khinkelman:
Hi @mwetter, I have refactored the `Steam` medium and it is ready for review. The changes are on my forked branch [issue1206_media_steam](https://github.com/khinkelman/modelica-ibpsa/tree/issue1206_media_steam). If you could please create a branch for this update, I (or you) can create a PR. Please let me know what you need from me.

As a high-level summary, here are the new models developed on this branch:
- `IBPSA.Media.Steam` -- New refactored steam medium model for water in the vapor-phase (IF97 region 2) only
- Example models for `IBPSA.Media.Steam`
- `IBPSA.Media.WaterHighTemperature` -- Extends `Media.Water` with a higher T_max and a function to calculate saturation temperature with respect to pressure. This water model can be used directly in high-pressure situations. 
- `IBPSA.Media.Interfaces.PartialPureSubstanceWithSat` -- Partial medium model to allow non-instantiated models to implement saturation state functions. 

Thanks!

mwetter:
@khinkelman : I just created the branch https://github.com/ibpsa/modelica-ibpsa/tree/issue1389_mediaSteam
Can you please make a PR against that branch.
Thanks.

PR #1391, opened 2020-09-02
Steam Medium Refactoring

khinkelman:
This PR is for the refactored steam medium and associated changes listed in [Issue 1389](https://github.com/ibpsa/modelica-ibpsa/issues/1389).

@mwetter, please let me know what comments/revisions/questions you have. Thanks!


mwetter:
@khinkelman : Thanks for the pull request. Before I merge it, can you please address the following:

- [ ] Move `IBPSA.Media.WaterHighTemperature` to `IBPSA.Specialized.Water.HighTemperature` as most users won't need to use this model. You should be able to use https://simulationresearch.lbl.gov/modelica/buildingspy/development.html#buildingspy.development.refactor.move_class to move the class (I often use it for moving models, but I think it works also for packages.)
- [ ] Your code has `IBPSA/Media/Steam.mo` and `IBPSA/Media/Steam/...` I think this is a mistake and only one is needed.
- [ ] I am not sure why the change at https://github.com/khinkelman/modelica-ibpsa/blob/e53a3e88ea9831176bed09b4664759ac96840384/IBPSA/Media/Examples/BaseClasses/PartialProperties.mo#L67 is needed, but in general, I don't think `function` should have a `parameter`. I am not certain if this is allowed in Modelica, but I generally simply use a `protected` variable for intermediate results.
- [ ] Running `jm_ipython.sh jmodelica.py IBPSA.Media.Examples.SteamProperties` gives a lot of warnings of the form 
> Warning at line 186, column 5, in file '/home/mwetter/proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA/Media/Steam.mo':
  The component state is declared multiple times and can not be verified to be identical to other declaration(s) with the same name.

  These should be avoided by removing the lines that start with a "+" as in
```
   redeclare function extends setState_phX
     "Return thermodynamic state as function of p, h and composition X or Xi"
-    extends Modelica.Icons.Function;
-    input AbsolutePressure p "Pressure";
-    input SpecificEnthalpy h "Specific enthalpy";
-    input MassFraction X[:]=reference_X "Mass fractions";
-    output ThermodynamicState state "Thermodynamic state record";
   algorithm
```
Because the function `extends setState_phX` it inherits all these elements and hence the lines should be deleted to avoid declaring it twice.

khinkelman:
Hi @mwetter, thank you for all of our insightful comments. I have moved `IBPSA.Media.WaterHighTemperature` to `IBPSA.Specialized.Water.HighTemperature` as requested, and deleted the `IBPSA/Media/Steam/...` mistake. 

For the changes to [IBPSA.Media.Examples.BaseClasses.PartialProperties](https://github.com/khinkelman/modelica-ibpsa/blob/e53a3e88ea9831176bed09b4664759ac96840384/IBPSA/Media/Examples/BaseClasses/PartialProperties.mo#L67), I thought that a relative error calculation would be helpful, particularly with water vapor that experiences larger pressure changes with incremental temperature changes. The absolute error of 1E-8 that is allowed in the function `checkState` cannot be met with the steam medium model due to the different forward/backward formulas for temperature/enthalpy calculations. Please advise on how you would like to proceed. Options in my mind: (1) We can keep the absolute and relative error functions as existing in this PR. (2) We can lower the acceptable error for the `checkState` function to meet the steam needs (note, `WaterProperties` and `AirProperties` also use this base class). (3) I can copy the code from `PartialProperties` into the `SteamProperties` to use a different absolute error criteria. (4) We can remove `checkState` from `protected` and make it `replaceable`, so that I can modify it for steam.

I ran the Steam example models with JModelica and uncovered another error when running `Examples.SteamDerivativeCheck`:
image

This `g2` function calculates the dimensionless Gibbs function and derivatives and outputs a record including the following variables:
image
I moved this function to the top level Steam medium model and added a smoothOrder=2 annotation. With this, JModelica couldn't compile the model. (It stalled for 1+ hour before I aborted the process.) 
Before proceeding, I would like to get your recommendations if you don't mind.
I expect the best solution would be to split this `g2` function into 8 separate functions (pi, tau, g, gpi, etc.), such that each could then be given the smoothOrder annotation. This would obviously add more lines of code to the top-level `Media.Steam` model if we inline them all in that model. Do you think this is the best approach? If not, do you have any other suggestions? Thanks again for your help!

mwetter:
@khinkelman 

Can you please merge ibpsa:issue1389_mediaSteam and resolve the conflicts that show in this PR.

Regarding checkState, I suggest you add as inputs the allowed error, for example
```modelica
   input Boolean errAbs=1E-8 "Absolute error used to in the check of the state calculations";
```
or option (3) would also be good.


Regarding the compilation problem please see my commit
https://github.com/khinkelman/modelica-ibpsa/commit/50b131a7f7118137bc607f36886e3a59290b8655
This works for JModelica and OPTIMICA.

khinkelman:
@mwetter, all requested changes are complete. Please let me know if there are any further revisions needed. Thanks again!

PR #1395, opened 2020-09-14
Issue1389 media steam

### LINKED ISSUE #1389: Refactored Steam Medium

mwetter:
This closes #1389


IBPSA/Media/Steam.mo
```diff
@@ -0,0 +1,855 @@
...
+    hl := Modelica.Media.Water.IF97_Utilities.BaseIF97.Regions.hl_p(sat.psat);
+    annotation (Inline=true);
+  end enthalpyOfSaturatedLiquid_sat;
+
+  redeclare replaceable function extends enthalpyOfSaturatedVapor_sat
+    "Return enthalpy of saturated vapor"
+  algorithm
+    hv := Modelica.Media.Water.IF97_Utilities.BaseIF97.Regions.hv_p(sat.psat);
+    annotation (Inline=true);
+  end enthalpyOfSaturatedVapor_sat;
```
mwetter, line 357: See above comment about whether the functions ending in `_sat` are really needed.

IBPSA/Media/Steam.mo
```diff
@@ -0,0 +1,855 @@
...
+           = 1
+</pre>
+</html>"));
+  end setSmoothState;
+
+  redeclare replaceable function extends saturationState_p
+    "Return saturation property record from pressure"
+  algorithm
+    sat.psat := p;
+    sat.Tsat := saturationTemperature_p(p);
```
mwetter, line 564: Better use the function name `saturationTemperature` that takes as an input `p` and returns `T`. Then it has the same API as `Modelica.Media.Interfaces.PartialRealCondensingGases.saturationTemperature`

IBPSA/Media/Examples/SteamSaturatedProperties.mo
```diff
@@ -0,0 +1,88 @@
...
+  parameter Modelica.SIunits.Temperature TMax = 647.096
+    "Maximum temperature for the simulation";
+  parameter Modelica.SIunits.Pressure pMin = 620
+    "Minimum pressure for the simulation";
+  parameter Modelica.SIunits.Pressure pMax = 20000000
+    "Maximum pressure for the simulation";
+  Modelica.SIunits.Pressure pSat "Saturation pressure";
+  Medium.Temperature TSat  "Saturation temperature";
+  Modelica.SIunits.Conversions.NonSIunits.Temperature_degC TSat_degC
+    "Celsius saturation temperature";
```
mwetter, line 19: I don't see a need to introduce `TSat_degC`. Better is to change the line above to
```
Medium.Temperature TSat(displayUnit="degC")  "Saturation temperature";
```

mwetter (commented):
@khinkelman : Please see the comments below, and let me know if it is not clear. The main changes/questions are why to introduce functions with endings `_sat`. This seem to overspecify the properties if they return values at the saturation line, and the input list is different from other media functions. I therefore think we should change them so they use the same input.

The info section in `Media.Steam` says that the pressure is assumed constant, but the media computes `d = density_pT(p,T);` which contradicts this statement.

I wonder if you made performance tests already. Evaluating `h` requires evaluating the function `g2` which seems quite expensive.

Simulating
```
jm_ipython.sh jmodelica.py IBPSA.Media.Examples.SteamTemperatureEnthalpyInversion
```
allows you to generate the html diagnostics, which then shows the function calls. Maybe this helps in seeing what is all involved in this seemingly small example and may help simplifying the implementation.

khinkelman:
Hi @mwetter, I've revised the steam medium per our discussion and want to get your input before proceeding further with testing. Please see the revised medium model [here](https://github.com/khinkelman/modelica-ibpsa/blob/issue1206_media_steam_rev2/IBPSA/Media/Steam.mo). Do you think we should continue moving forward with this approach? A few highlights to note:

- The enthalpy(p,T) and inverse temperature_ph functions are the linear approximations discussed
- The IF97 functions are still used for functions called less frequently
- I changed the base package to `Modelica.Media.Interfaces.PartialCondensingGases`. This allows the `Steam` medium to extend an existing partial medium from the Modelica library, but does not comprehensively provide all saturation state functions for model development. I think this is okay since we can design the models to only need `saturationPressure` and `enthalpyOfVaporization` functions to perform the necessary calculations.  

Testing highlights so far:
- Simulating Examples.SteamTemperatureEnthalpyInversion with JModelica shows a much smaller set of function calls.
- Performance testing with the 1st generation district heating component models have been successful. Using this medium with the full system model is still in progress.

mwetter:
@khinkelman : The equations look good in terms of the complexity. It looks however a bit strange to use `PartialCondensingGases` because this steam model is not condensing but I currently see no drawback in using this base class. So I think we should proceed with it. Worst case, if there are issues, we could base it on `PartialMedium` and then reimplement the functions you need from `PartialCondensingGases` but I don't expect any problems as-is.

khinkelman:
Thanks @mwetter, 
Please correct me if I am wrong, but I believe if we were to base the medium on `PartialMedium`, then we would not be able to call the saturation functions (e.g. `enthalpyOfVaporization`) in the component models (as I am doing currently) and have the actual `Steam` medium only declared at the top-level system model (with partial medium models declared through all base models). Is this true?  

Based on your comment, I may suggest we go ahead and make `PartialMedium` the base to minimize future redesign of 1st gen district heating models. We could then calculate the saturation properties only at the top level and propagate them as parameters down through the model layers as needed. This won't be a problem since we are treating pressure as constant and thus, these property values are also fixed at initiation. An additional benefit of this approach is that the non-instantiable models would all declare the mediums as as `PartialMedium` . Rather than currently, I am declaring some medium packages as (for example) `PartialCondensingGas` so that we can use the `enthalpyOfVaporization` function in lower-level models. What do you think?

mwetter:
@khinkelman : Correct, if the base class is `PartialMedium`, then it would be a syntax error to call `Medium.enthalpyOfVaporization`. The model check should fail in this case because the function is not known.

Your suggestions sounds good. However, if you don't like to send properties through parameters, you could simply compute these parameters in the model by calling a function such as
```modelica
protected
  parameter Medium.Density rho_default=Medium.density(sta_default)
    "Density, used to compute fluid volume"; // or you can call any other function using e.g., Medium.p_default.
  parameter Medium.ThermodynamicState sta_default=
     Medium.setState_pTX(T=Medium.T_default, p=Medium.p_default, X=Medium.X_default)
    "Default medium state";
```.

**Comment by khinkelman** [comment]
Thanks @mwetter, great suggestion. I'll move forward with `PartialMedium` as the base.

File: IBPSA/Media/Specialized/Water/HighTemperature.mo [hunk 4]
```diff
@@ -0,0 +1,117 @@
+within IBPSA.Media.Specialized.Water;
+package HighTemperature
+  "Package with model for liquid water with constant density and an extended 
+  maxiumum temperature threshold"
+   extends IBPSA.Media.Water(
+     T_max = 273.15+600);
+
+  replaceable function saturationTemperature_p
+    "Return saturation temperature (K) from a given pressure (Pa)"
```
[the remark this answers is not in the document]
mwetter: @khinkelman : I will rename it to `IBPSA.Media.Specialized.Water.HighTemperature.saturationTemperature`, and change the argument `p_sat` should be `p` so it has the same conventions as
`IBPSA.Media.Air.saturationPressure`

mwetter:
Todo
- [ ] The function `IBPSA.Media.Specialized.Water.HighTemperature.saturationTemperature` is not used anywhere. Should this be deleted? If it stays, it also needs a unit test.
- [ ] Functions that are specialized and you think are only needed in this media should be made protected. For example, `temperature_h` and `temperature_s` in `Media.Steam` are probably not directly called by any other model. Hence it is better to make them protected. This will make it easier to refactor or remove them without breaking backward compatibility if needed.

khinkelman:
Thanks @mwetter for your insightful commits and comments. I've addressed the todo items on my fork [issue1206_media_steam](https://github.com/khinkelman/modelica-ibpsa/tree/issue1206_media_steam). Could you please pull my recent commits? Let me know if there is anything else you need. 

1.  `IBPSA.Media.Specialized.Water.HighTemperature.saturationTemperature` is redundant when considering the coupling with `IBPSA.Media.Steam`. Thus, I've removed this function.
2.  I've made the suggested functions protected. I also made the top-level functions (e.g. `setState_phX`, which uses `temperature_h`) replaceable to allow flexibility in the implementation.

khinkelman:
Hi @mwetter, I've deleted `IBPSA.Media.Specialized.Water.HighTemperature` on my fork [issue1206_media_steam](https://github.com/khinkelman/modelica-ibpsa/tree/issue1206_media_steam) and adjusted the Info section of `Media.Steam` accordingly. Please pull my commits and let me know if there is anything else you need. Thanks again for all of your help!

mwetter:
@khinkelman 
The steam implementation looks good and ready to merge from my point of view. I will also work on getting it on the Buildings branch. 

The main changes I made was
- correct the smoothOrder annotation as not all functions are differentiable
- remove the constants "phase" and "region". As they were public, they constrained what other media can be used if a model implements IBPSA.Media.Steam as a replaceable medium.

khinkelman:
Thanks @mwetter, with your most recent commits, the implementation looks good to me.

Issue #1402, opened 2020-09-29
Add weather station for BOPTEST


dhblum:
This issue is suggested to address https://github.com/ibpsa/project1-boptest/issues/234.  In particular, to add a model to the package ``IBPSA.Utilities.IO.SignalExchange`` that implements a number of pre-configured instances of ``IBPSA.Utilities.IO.SignalExchange.Read`` in order to facilitate the reading of weather measurements for BOPTEST from building emulators.  These measurements would represent a controller's access to a weather station at or near the building site.

PR #1403, opened 2020-09-29
Issue1402 wea sta boptest

dhblum:
This is proposed for #1402.

- Adds weather station model (``IBPSA.Utilities.IO.SignalExchange.WeatherStation``)
- Adds example (``IBPSA.Utilities.IO.SignalExchange.Examples.WeatherStation``)


dhblum:
Suggest also a review from @JavierArroyoBastida.

JavierArroyoBastida:
@dhblum the implementation looks good to me in general, I'm also loving the icon :) 
Just have a question, is there a reason for the selection of variables that the weather station is reading? shouldn't we read all variables carried by `weaBus`?

dhblum:
@JavierArroyoBastida That's a good point.  The original intent was that there are more typical variables that are measured with weather stations.  But, we do provide forecasts for all data in the bus right now, so I suppose we should also measure them, and also not decide for control developers what is "typical" at this point.

I will change so that all variables are measured, and also check what's going on with the broken tests.

JavierArroyoBastida:
Sounds good to me. I'll go through it once again when ready.

dhblum:
@JavierArroyoBastida This is ready to review again, also waiting for checks to pass.

dhblum:
I don't understand why the tests are failing now, as they pass locally.  Is there a way to see the ``simulator-dymola.log`` from travis?

Mathadon:
image

dhblum:
Thanks @Mathadon!  I found the error and pushed a fix.  Awaiting test results.

dhblum:
@JavierArroyoBastida Any more comments?

JavierArroyoBastida:
@dhblum no further comments from my side. Nice to see that you fixed the bug. I think this can be merged already.

IBPSA/Utilities/IO/SignalExchange/WeatherStation.mo
```diff
@@ -0,0 +1,316 @@
...
+model WeatherStation
+  "Implements typical weather measurements with signal exchange blocks"
+  IBPSA.BoundaryConditions.WeatherData.Bus weaBus "Weather data"
+    annotation (Placement(
+        transformation(extent={{-110,-10},{-90,10}}), iconTransformation(extent={{-112,
+            -14},{-86,12}})));
+  IBPSA.Utilities.IO.SignalExchange.Read reaWeaTDryBul(description=
+        "Outside drybulb temperature measurement", y(unit="K"))
+    "Outside drybulb temperature measurement"
+    annotation (Placement(transformation(extent={{-40,80},{-20,100}})));
```
mwetter, line 11: @dhblum : Can you please correct the line breaks for these instances, and make the `unit` final.

IBPSA/Utilities/IO/SignalExchange/WeatherStation.mo
```diff
@@ -0,0 +1,316 @@
...
+      color={255,204,51},
+      thickness=0.5), Text(
+      string="%first",
+      index=-1,
+      extent={{-6,3},{-6,3}},
+      horizontalAlignment=TextAlignment.Right));
+  annotation (Documentation(info="<html>
+<p>
+This block enables the reading of weather measurements and their meta-data by an external
+program without the need to explicitly propogate the signal to a top-level model.
```
mwetter, line 270: Shouldn't this be "propagate"? We have this typo in various models.

mwetter (changes requested):
Please see inline comments. Then it is ready to merge.

IBPSA/Utilities/IO/SignalExchange/WeatherStation.mo
dhblum: Fixed in https://github.com/ibpsa/modelica-ibpsa/pull/1403/commits/97e362e1c1789e7ccf2baff623a9715ebe0b04d7.

IBPSA/Utilities/IO/SignalExchange/WeatherStation.mo
dhblum: Resolved in https://github.com/ibpsa/modelica-ibpsa/pull/1403/commits/45bb615e3c8b97590970f4619cdc2138e4442ef1 and https://github.com/ibpsa/modelica-ibpsa/pull/1403/commits/c9e43f476b51e0de0e72863ae28bafd6822bc06c.

dhblum:
@mwetter I have addressed your comments.  Waiting for tests to pass.

mwetter (approved):

PR #1404, opened 2020-09-30
Issue1401 event indicators

### LINKED ISSUE #1401: Change event triggering assertions in Media
The `IBPSA.Media` implementation has various assertions in the `BaseProperties` that lead to zero crossing functions. This issue is to refactor them to avoid these zero crossing functions, and also to make an implementation that is easier to understand (as it won't use the `redeclare replaceable model extends BaseProperties` implementation).

See also https://github.com/lbl-srg/modelica-buildings/issues/2152

mwetter:
This closes #1401.

We could add the same refactoring to the other media, for now I did it for `Air` and `Water` as these are the ones we use most.

### DIFF
--- a/IBPSA/Media/Air.mo
+++ b/IBPSA/Media/Air.mo
@@ -118,6 +118,7 @@ package Air
     X[1] = Xi[1];
     X[2] = 1 - X[1];
 
+    // Assertions to test for bounds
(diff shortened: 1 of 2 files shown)


mwetter:
@Mathadon : Can you please review this PR. Thanks.

Mathadon (approved):

Issue #1396, opened 2020-09-18
converting epw files with a space in filename fails


mwetter:
@EttoreZ reported that if the weather data converter is run with a file that has a space in its name, it fails with an error. This issue is to correct the code.

EttoreZ:
As @mwetter mentioned I found a small bug in the java weather converter, where if there is a space in the filename you get an error. See screenshot below:
![image](https://user-images.githubusercontent.com/46685675/93600240-275ab500-f974-11ea-8f7e-fbe5bef9e812.png)
I will make a change to the java code and make a pull request to address the issue.

EttoreZ:
Putting a space in the filename causes a loss of the part after the space in the string read by the Java script. To avoid this the file with a space in the name should be called between " ". See screenshot below as proof. My suggestion would be to integrate the existing warning message " Name of weather data file should end with \".epw\" with  "Name of weather data file should end with \".epw\" or name contains space in that case write name between "" as "weatherdata .epw"" in this way it will work both in Linux terminal and windows cmd. @mwetter if you agree with the change I will make a pull request.
![image](https://user-images.githubusercontent.com/46685675/93692335-e374c680-fabf-11ea-9a12-0a992adbc021.png).

EttoreZ:
@mwetter I created a branch on my fork to address the issue <a href="https://github.com/EttoreZ/modelica-ibpsa/tree/issue1396_WeatherFileWarning">issue1396_WeatherFileWarning</a>, as soon as you create the branch on the new repository I will make the pull request

mwetter:
@EttoreZ : I created the branch `issue1396_WeatherFileWarning`.

PR #1405, opened 2020-10-06
Issue1396 weather file warning

EttoreZ:
This pull request is to adresses [#1396 ](https://github.com/EttoreZ/modelica-ibpsa/tree/issue1396_WeatherFileWarning) by:

- Updating documention of ReaderTMY3 module
- Adding an additional warning to the Java script  missing : "Name of weather data file should end with ".epw" or if spaces are present in the name write name between "" "


mwetter (commented):
@EttoreZ : Is the `jar` file needed for this PR? If so, please also add the `.java` source code that changed and was used to produce the `.jar` file.

EttoreZ:
@mwetter : The `.jar` file was the only file present in the Resources/bin folder. Inside the `.jar` file there is both the compiled and the source code. I kept the orginal `.jar` file and updated the `.class` and `.java` code inside.

mwetter:
@EttoreZ : This needs to be put in `IBPSA/Resources/src/convertEPW/src` because `IBPSA/Resources/src/convertEPW/src/Makefile` generates the jar file and copies it to `bin`.

PR #1351, opened 2020-04-24
Issue1314 bestest weather

mwetter:
This merges the IBPSA weather data CI testing to the master.

This integrates https://github.com/ibpsa/modelica-ibpsa/pull/1344 because it changes the results.

Note that the results are not stored in git since they take a lot of space:
```
$ python generateResults.py -p
$ du -hs results/JsonResults
16M	results/JsonResults
$ python generateResults.py 
$ du -hs results/JsonResults
6.6M	results/JsonResults
```
Results can be generated using
```
cd Resources/Data/BoundaryConditions/Validation/BESTEST
python3 generateResults.py -p
```

Closes #1314

PR #1408, opened 2020-10-12
Issue1314 bestest weather

EttoreZ:
This is to adresses @PMehrfeld  comment on #1351.  I updated the formatting of the Python script.

- I checked the code with  http://pep8online.com. I corrected the large majority of the errors, there are just a couple which may be an issue with the website and not the code
- I updated the parameter documentation for all the functions
- I updated all the variables names taking out the camel case, apart from the ones that are strictly required to be formatted as such.


PMehrfeld (approved):

PMehrfeld:
Thanks, @EttoreZ, looks good so far :+1:.

The CI failed, but I am pretty sure this is because of the fork in the namespace `EttoreZ`. 
@mwetter: If my assumption is correct and the CI fail is no problem, would you please merge?

PR #1415, opened 2020-11-08
Reformulated R410A for #1414

### LINKED ISSUE #1414: inconsistent use of parameter in functions in IBPSA/Media/Refrigerants/R410A
The functions in `IBPSA/Media/Refrigerants/R410A` use constructs such as
```modelica
  Real A[:] = {-1.721781e2, 2.381558e-1, -4.329207e-4, -6.241072e-7}
    "Coefficients A for Martin-Hou equation of state";
...
  parameter Integer n = size(A, 1);
```
This leads to a compilation error in OpenModelica. Here, the `parameter` keyword can be removed, or `size(A, 1)` can be directly used in the function body.

Also, some functions don't use `n`, in which case it can be removed.

mwetter:
This closes #1414


mwetter:
@MassimoCimmino : Please see https://github.com/ibpsa/modelica-ibpsa/pull/1415/commits/0782dc34961545d6c7161c9858460d0376f82f99 where I moved the constants to the package scope. The relative error is small, but the absolute is sufficiently big to require updating the reference results.

Can you please verify whether
https://github.com/ibpsa/modelica-ibpsa/blob/0782dc34961545d6c7161c9858460d0376f82f99/IBPSA/Media/Refrigerants/R410A/enthalpySatLiq_T.mo#L13
and
https://github.com/ibpsa/modelica-ibpsa/blob/0782dc34961545d6c7161c9858460d0376f82f99/IBPSA/Media/Refrigerants/R410A/enthalpySatVap_T.mo#L13
are correct.
It looks strange to have
```modelica
  final constant Real x0 = 0
```
If `x0=0` is indeed correct, it would be best to either remove it, or add a comment so it is clear that this is not a bug.

MassimoCimmino:
@mwetter This is correct. There was no correlation for the enthalpy of saturated vapor in the reference properties so I produced a polynomial fit like the one for the enthalpy of saturated liquid. The parameter `x0` does not affect the accuracy of the fit and I did not consider fixing it to `x0 = 0.5541498` (it would have been consistent).

I removed it from the code.

mwetter:
@MassimoCimmino : Thanks, if you are happy with the changes, please approve the PR and I can then merge it.

MassimoCimmino (approved):

mwetter:
Using the sum as opposed to a loop should be as efficient, if not more
efficient (depending on how it is handled by the compiler). It also is
easier to read.

If you like it would be great to optimize the other functions as well, and
also use intermediate variables for common expressions such as
Modelica.Math.exp(-k*T/TCri))/(v_abs - b)
in pressureVap_Tv.mo

Best would be to do it on a new branch so we can go ahead merging this one.

PR #1418, opened 2020-11-11
Issue1416 java weather converter

### LINKED ISSUE #1416: Java weather converter
This issue is to update the Java weather converter so that it will handle sub hourly data as intended in the .epw format. In particular 
if the field "minutes" is 00 or 60, then the data is considered hourly and the value reported is the value at the end of the hour. Instead if the "minutes" value is between 00 and 60 the field "hour" and "minutes (1,10) should be interpreted as  10 minutes in the first hour or just 10 minutes, while currently the code currently interprets it as 1 hour and 10 minutes or 70 minutes.
I already corrected the code and as soon as an issue branch is open I will make the pull request.

mwetter:
This closes #1416.

To do:
- [x] recreated binaries.


IBPSA/BoundaryConditions/WeatherData/ReaderTMY3.mo
```diff
@@ -1470,6 +1470,17 @@ For instance, the unit must be
 </ul>
 </li>
 <li>
+<p>
+Hourly and subhourly timestamp are handled in a different way in <code>.epw</code> files.
+From the EnergyPlus Auxiliary Programs Document (v9.3.0, p. 63):
+In hourly data the minute field can be <code>00</code> or <code>60</code>. In this case as mentioned in the previous section, the weather data
+is reported at the hourly value and the minute field has to be ignored, wrinting <code>1,60</code> or <code>1,00</code> is equivalent.
```
mlauster, line 1477: writing instead of wrinting

mlauster, line 1477: And sure about `1,60`? It really looks uncommon. Edit: I just saw the next section, so `1,60` might be equivalent to `02:00`?

mlauster (changes requested):
I didn't check the .js files, since they are auto-generated. The rest is fine, thanks for your work! I just found one typo. :-)

IBPSA/BoundaryConditions/WeatherData/ReaderTMY3.mo
mwetter: Thanks, I corrected it in 6b1e36d74

mwetter: @mlauster : Yes, we are sure and agree that it look inconsistent. @EttoreZ double-checked with the EnergyPlus support and inquired specifically about this inconsistency. This is how EnergyPlus handles the minute field.

mlauster (approved):

PR #1430, opened 2020-12-08
Setup github actions for documentation checks

mwetter:
This PR sets up the documentation tests using github actions.
This gives faster turn-around times than waiting for travis.

The PR also contains fixes the the html formatting that were detected in the tests.


Mathadon:
@mwetter thanks for looking into this! I was also looking into GitHub actions yesterday. An alternative that I found is https://www.appveyor.com.

Issue #1412, opened 2020-11-03
Fluid.Interfaces.Examples.Humidifier_u is unstable with Cvode in Dymola


mwetter:
The model IBPSA.Fluid.Interfaces.Examples.Humidifier_u shows oscillations if simulated with CVODE:

![image](https://user-images.githubusercontent.com/1329080/98037002-cd049f00-1dcf-11eb-8c25-be8367e41d8f.png)

If changed to Radau, these oscillations do not occur.

Moreover, the statistics is for the original formulation:
```
Sizes of linear systems of equations: {3, 3, 3, 3}
Sizes after manipulation of the linear systems: {0, 0, 0, 0}
Sizes of nonlinear systems of equations: {3, 3, 1, 4, 3, 3, 4, 1, 4, 3}
Sizes after manipulation of the nonlinear systems: {1, 1, 0, 1, 1, 1, 1, 0, 1, 1}
Number of numerical Jacobians: 0

Initialization problem
Sizes of nonlinear systems of equations: {3, 1, 3, 1, 4, 1, 3, 1, 3, 1, 4, 1, 4, 1, 3, 1}
Sizes after manipulation of the nonlinear systems: {1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0}
Number of numerical Jacobians: 0
```

In the Buildings library result comparison across tools, we remove the inverse function annotation in `basicFlowFunction_m_flow` and in `basicFlowFunction_dp`. This leads to
```
Sizes of linear systems of equations: {3, 3, 3, 3}
Sizes after manipulation of the linear systems: {0, 0, 0, 0}
Sizes of nonlinear systems of equations: {3, 3, 1, 4, 3, 3, 4, 1, 4, 3}
Sizes after manipulation of the nonlinear systems: {1, 1, 1, 1, 1, 1, 1, 1, 1, 1}
Number of numerical Jacobians: 0

Initialization problem
Sizes of nonlinear systems of equations: {3, 3, 1, 4, 3, 3, 4, 1, 4, 3}
Sizes after manipulation of the nonlinear systems: {1, 1, 1, 1, 1, 1, 1, 1, 1, 1}
Number of numerical Jacobians: 0
```

It turns out that this model then does not simulate with Cvode 1E-6 because the oscillations trigger the assertions in the media, e.g., the model is unstable. Best is to switch for this model to the Radau solver.

Mathadon:
[...]

PR #1433, opened 2020-12-09
As a work-around for #1412, changed solver for Humidifer_u

mwetter:
This is a work-around because addressing #1412 will take longer.

@Mathadon : I suggest to merge this small change, and in parallel I will work on the state selection to address the root cause for #1412. Merging it will make at least this model work fine again.


Mathadon (approved):

Mathadon:
okay @mwetter #1412 indeed seems to be a hard nut to crack

PR #1438, opened 2020-12-15
Changes for kT in simpleDoor model

kldjonge:


kldjonge:
Noticed other issues,  additional commits following

kldjonge:
When dP=0pa, results match:
image
When dP=1pa, this is the results:
image
image

DoorAdap is the doormodel with adjusted stack effect.

IBPSA/Airflow/Multizone/DoorOpen.mo
```diff
@@ -24,8 +24,8 @@ protected
     "Polynomial coefficient for regularized implementation of flow resistance";
 
   parameter Real kVal=CD*AOpe*sqrt(2/rho_default) "Flow coefficient, k = V_flow/ dp^m";
-  parameter Real kT = CD*wOpe*sqrt(2/rho_default)
-    *(Modelica.Constants.g_n*rho_default*hOpe/2/Medium.T_default)^m *hOpe/2
+  parameter Real kT = CD*wOpe*hOpe/2*sqrt(2/rho_default)
+    *(Modelica.Constants.g_n*rho_default*(2*hOpe/9)/Medium.T_default)^m
```
mwetter, line 28: If I see this right, then the `hOpe/2` has just been moved around (no real change here), but the `1/2^m` has been changed to `(2/9)^m`. What is the explanation for this? If it is justified, can you please add it to the info section.

mwetter (commented):
@kldjonge : Please see the question before merging.

Mathadon:
@kldjonge nice analysis! In the middle graph, the Modelica results don't make sense, though? It seems as if the temperature and pressure-driven flow are treated separately since the typical mass flow curve of Modelica can still be recognized and it's point of symmetry is still around dT=0K. Does the model implementation indeed consider the two effects separately?

mwetter:
Regarding https://github.com/ibpsa/modelica-ibpsa/pull/1438#issuecomment-745555414 this is correct. The non-physical behavior is because of the model simplification that separates pressure-driven flow from temperature-driven flow.

IBPSA/Airflow/Multizone/DoorOpen.mo
kldjonge: The CONTAM user guide provides this information. They convert equation 69 to fit the powerlaw equation as defined by 
[1]W. G. Brown and K. R. Solvason, “Natural convection through rectangular openings in partitions—1 Vertical Partitions,” Int. J. Heat Mass Transfer, vol. 5, pp. 895–868, Mar. 1962.

This results in a stack effect of 4/9H between the two flow paths,  or twice 2/9H from the centerline.

image

kldjonge:
Considering the simplifications, for which applications can this model be used? Is it clear for the user for which applications this model can be used reliably? 
e.g. in the 3rd graph, the behaviour will deviate quite fast when temperature differences increase/

IBPSA/Airflow/Multizone/DoorOpen.mo
kldjonge: To be clear, the original implementation assumed a stack column height of hOpe between the two flow paths, which was an overestimation.

PR #1441, opened 2020-12-20
Issue975 butterfly valve

### LINKED ISSUE #975: add butterfly valve model
We currently do not have a model for a butterfly valve. Typically these valves are used for open/closed operation only, however in some cases they are used for control too. In these cases it would be good to have a somewhat accurate valve characteristic.

Some sources for such characteristics:
 - http://cgproducts.johnsoncontrols.com/MET_PDF/347VB2.PDF?x=86 p13 figure 6
 - https://www.engineeringtoolbox.com/butterfly-valves-flow-coefficients-d_224.html
 - CFD study, but which does not seem to provide data for constructing a valve characteristic http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.1005.2457&rep=rep1&type=pdf 
 - https://www.sciencedirect.com/science/article/pii/0019057895000240 fig 8
 - https://pdfs.semanticscholar.org/8fa6/277c7729cd5aba9631961cdd214007bdc882.pdf fig 7 (large diameter)
 - http://www.dcsc.tudelft.nl/~bdeschutter/private_20100705_acc_2010/data/papers/1640.pdf figure 3: analytical model
 - https://link.springer.com/chapter/10.1007/978-3-540-75995-9_149 fig 7: experiment + CFD
 - http://blog.opticontrols.com/archives/727 fig 3: flow characteristic can be computed from gain
 - http://www.samson-regeltechniek.nl/2_Markten/Technische_onderwerpen/Selectie_&_trim/Klepkarakteristiek_7_5_4_1.pdf last page
 - https://repository.tudelft.nl/assets/uuid:727d9a2a-a852-4fd7-99b5-b7f045c7d512/318.pdf figure 5
 - http://iopscience.iop.org/article/10.1088/1755-1315/12/1/012098/pdf large diameter
 - http://www.valmatic.com/pdfs/Cavitation_in_Valves_7-22-08.pdf figure 3
 - http://www.polna.com.pl/en/flow-characteristic.html figure 2
 - http://www.valmatic.com/pdfs/SurgeControlinPumpingSystems3-17-09.pdf figure 10

Some questions before I implement this:
 - I propose to implement the characteristic by extending `TwoWayPolynomial`.
 - We could add the valve diameter as parameter that computes a default value for `dp_nominal`, using another polynomial e.g. based on the data in https://www.engineeringtoolbox.com/butterfly-valves-flow-coefficients-d_224.html , or http://cgproducts.johnsoncontrols.com/MET_PDF/347VB2.PDF?x=86 page 22. 
 - The cleanest way to do this would be to add the option `Diameter` to `CvData`, but this would be an invalid option for the other valve models so I don't think is realistic. Unless an `enumeration` can be extended and redeclared. So perhaps it's better to not add the `diameter` option for now, or to implement it as a separate function.
 - Proposed name: `TwoWayButterfly`

@icupeiro @damienpicard this is for Infrax

Mathadon:
closes #975 

@mwetter can you review this?


IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
```diff
@@ -0,0 +1,30 @@
+within IBPSA.Fluid.Actuators.Valves;
+model TwoWayButterfly
+  "Two way valve with the flow characteristic of a typical butterfly valve"
+  extends IBPSA.Fluid.Actuators.Valves.TwoWayPolynomial(
+    final CvData = IBPSA.Fluid.Types.CvTypes.Kv,
```
mwetter, line 5: I don't see a reason why setting this final. Can you please provide an assignment for `Cv` and if possible for `Av` based on the diameter?

IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
```diff
@@ -0,0 +1,30 @@
...
+    Kv = 0.0472*(DN*1000)^2 -2.0104*(DN*1000),
+    final c={0,1.101898284705380E-01, 2.217227395456580,  - 7.483401207660790, 1.277617623360130E+01, -6.618045307070130});
+
+  parameter Modelica.SIunits.Length DN(displayUnit="mm") "Nominal diameter";
```
mwetter, line 9: This should be `d` for inner diameter. DN is usually based on outer diameter and not a good value to compute the inner diameter because the wall thickness depends on the material and rated pressure. Maybe we could add a table in the info section that shows typical inner diameters for a given DN.

Also, I suggest to not set the `displayUnit` attribute because in `HydaulicDiameter`, it is also not set. Then the parameter window will be consistent for the two models.

IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
```diff
@@ -0,0 +1,30 @@
...
+
+annotation (
+defaultComponentName="valBut",
+Documentation(info="<html>
+<p>
+Two way valve with the flow characteristic of a typical butterfly valve.
```
mwetter, line 15: Can you add a figure that shows the users the opening characteristics?

IBPSA/Fluid/Actuators/Valves/Examples/TwoWayValves.mo
```diff
@@ -64,6 +64,13 @@ model TwoWayValves "Two way valves with different opening characteristics"
     CvData=IBPSA.Fluid.Types.CvTypes.OpPoint)
     "Valve with polynomial opening characteristic"
     annotation (Placement(transformation(extent={{0,-90},{20,-70}})));
+  TwoWayButterfly valBut(
+    redeclare package Medium = Medium,
+    Kv=valPol.Kv,
+    use_inputFilter=false,
+    m_flow_nominal=1,
+    l=0.05) "Butterfly valve"
```
mwetter, line 72: The test fails because the diameter is not set.

mwetter (changes requested):
Please see inline comments.

IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
Mathadon: If the user changes `CvData`, `Kv` will get `fixed=false`, but the model also assigns `Kv = 0.0472*(DN*1000)^2 -2.0104*(DN*1000),`. This results in a warning and I don't see another way of working around that. If the warning is ok then we can remove the `final`.

image

IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
Mathadon: We could use the inner diameter, but the source table that I used lists DN's. Do we keep the same data?

IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
Mathadon: I removed `final`.

Mathadon:
@mwetter see my answers. Changes have been pushed.

IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
mwetter: @Mathadon : I think it is better to use
```modelica
    final CvData = IBPSA.Fluid.Types.CvTypes.Kv,
```
because `Kv` has a binding, which would need to be removed by the user if another option is selected for `CvData`. I will add `final` in my revision.

IBPSA/Fluid/Actuators/Valves/TwoWayButterfly.mo
mwetter: I see. The data fit is based on an "inner diameter" which is set to be the same number as the DN value, even though they are typically not the same. I suggest to then just use a `choices` selecting that allows the user to select a DN value, and assigns the `Kv` value. Please see https://github.com/ibpsa/modelica-ibpsa/pull/1441/commits/50f11fc6554e6f0a27203369c410d35cee4adcf9

mwetter (approved):
@Mathadon : This is ready to merge if you agree with my changes, in particular https://github.com/ibpsa/modelica-ibpsa/pull/1441/commits/50f11fc6554e6f0a27203369c410d35cee4adcf9

Mathadon:
@mwetter thanks for implementing the choices. It would make sense to extrapolate the data quadratically since the DNs that are listed are rather large. The extrapolation should go through the origin (Kv of 0 for diameter of 0). The curve that I implemented had that property. I'll add some more choices for smaller diameters?

mwetter:
@Mathadon : If you think a quadratic extrapolation is valid then it would indeed be good if you can add more choices for smaller diameters.

mwetter:

PR #1442, opened 2021-01-23
Issue1353 simple door

### LINKED ISSUE #1353: add simplified door model
The current door models `Airflow.DoorDiscretized*` are based on an orifice equation model in which the pressure difference drives the flow rate. If used with an HVAC system that models flow friction in the return duct, then each pressure (or mass) in the room is a state that varies very quickly due to very small difference in pressure of the rooms that are connected. This causes very fast transient of the pressure (or mass), driven by small differential pressures, and the equations are coupled to slow dynamics of temperature, and hence stiff. This causes problems for example in
https://github.com/lbl-srg/modelica-buildings/issues/1911#issuecomment-620193181.
It may also be the reason for https://github.com/lbl-srg/modelica-buildings/issues/1895.

This issue is to implement a simplified door model that still allows for bi-directional, temperature difference-driven flow between zone, but it will set the pressure of each zone to be equal. Then, a translator will do index reduction and only use one state for all room pressures (or mass).

mwetter:
This closes #1353 

This merges the simple door model. The validations 
`IBPSA.Airflow.Multizone.Validation.ThreeRoomsContam` (with the new door)
and
`IBPSA.Airflow.Multizone.Validation.ThreeRoomsContamDiscretizedDoor` (with the old door)
now produce almost identical results. The small difference is due to the discretization of one of the door model.

Also, validation tests


mwetter:
See https://github.com/ibpsa/modelica-ibpsa/issues/1353#issuecomment-766932383 for discussion of computing time.

@Mathadon : This is ready for review.

IBPSA/Resources/Scripts/Dymola/Airflow/Multizone/Validation/DoorOpenClosed.mos
```diff
@@ -0,0 +1,3 @@
+simulateModel("IBPSA.Airflow.Multizone.Validation.DoorOpenClosed", method="CVode", stopTime=1, tolerance=1e-06, resultFile="DoorOpenClosed");
+createPlot(id=1, position={15, 15, 799, 543}, y={"doo.port_a1.m_flow", "doo.port_b2.m_flow", "dooOpeClo.port_a1.m_flow", "dooOpeClo.port_b2.m_flow"}, range={0.0, 1.0, -1.0, 3.0}, grid=true, colors={{28,108,200}, {238,46,47}, {0,140,72}, {217,67,180}});
+createPlot(id=1, position={15, 15, 799, 543}, y={"lea.port_a.m_flow", "mNetClo_flow.y"}, range={0.0, 1.0, -0.005, 0.009999999999999998}, grid=true, subPlot=102, colors={{28,108,200}, {238,46,47}});
```
Mathadon, line 3: Is this `subplot=102` intentional? Dymola 2020 does not like it:

image

Mathadon:
@mwetter I made some minor changes and please check the comment regarding the plot command in Dymola. Approved once we sort that out (or accept it as an artifact of an old dymola version if that is the case).

IBPSA/Resources/Scripts/Dymola/Airflow/Multizone/Validation/DoorOpenClosed.mos
mwetter: @Mathadon : The `subPlot=102` is not intentional. I can fix it tomorrow. Thanks for catching it.

Mathadon: Just to be sure: the other scripts have it too.

mwetter:
@Mathadon : do you approve this PR? Please let me know so it can be merged or improved.

Mathadon (approved):

Mathadon:
The plots work now, thx!

PR #1443, opened 2021-01-27
Issue1432 get time span long header

### LINKED ISSUE #1432: Possible Bug in getTimeSpanTMY3 function
There might be a bug in the getTimeSpanTMY3 function, used by the ReaderTMY3 of the library. Under windows, with Dymola Version 2021 (64-bit), 2020-04-17 with Visusal Studio 2017/Visual C++ 2017 Express Edition (15) set as compiler, the function cannot read properly long lines.

It might be an issue with the c - code behind that function as this is an external c function call. Unfortunately I have no idea about c and how to fix the problem. 

Reproducing the error is not that straight forward as I am getting the error, but not all the time.

At least in my last tries I managed to reproduce the error under windows with the Testgettimespan model attached and the reading of the example_weather.mos. I hope this is as simple as possible to hopefully reproduce the error.
[dymola.zip](https://github.com/ibpsa/modelica-ibpsa/files/5660508/dymola.zip)

mwetter:
This closes #1432 

The implementation is completely refactored because the old one had complexity that was not needed. In particular, there is no need to read the whole line and manage its memory requirement which is not known a priori.


mwetter:
@MichaMans : Can you please test if this works on your system. For me it works on Linux (Dymola and OCT) and Windows (Dymola)

Mathadon:
@mwetter I have reviewed the code. The main concern that I have is whether the user gets a clear error message when the file is not correctly formatted?

MichaMans (commented):
@mwetter Thank you very much for this fix!

I tested it with Dymola on a Windows machine and it worked as expected. As I am not really into c programming, I can not review your actual code but what I tested is:
- checking out commit 3276b8b953976ac10001976dc6f6fcf167554783 with the example which does not work as well on my machine
- checked the latest version of this branch which works as expected even with much longer headers

So for me the code works and I think the bug is solved with the rework on the getTimeSpan function you did.

mwetter:
@Mathadon : please see https://github.com/ibpsa/modelica-ibpsa/pull/1443/commits/776cda6dd24afa1617149f38871b18ad76619630 which gives a clearer message.

Mathadon:
Great, thanks!

PR #1444, opened 2021-02-03
Better error message for out of bound eps

bravache:
Originally this change was only to fix the calculation in the error message, but looking further into it, I also figured that the `abs` is redundant in the calculation of `eps_nominal`.

Reasoning: Without the `abs`, you could rewrite the `eps_nominal` equation like so:
```
    eps_nominal = (T_b2_nominal - T_a2_nominal) / (T_a1_nominal - T_a2_nominal)
```

The factor `(T_b2_nominal - T_a2_nominal) / (T_a1_nominal - T_a2_nominal)` is always positive when the user selected nominal values are right: side 2 is heating up (b2 >= a2) if and only if side 1 comes in hotter (a1 >= a2). If the values are wrong, then the following `assert` statement would catch it (`eps_nominal < 0`).

### DIFF
--- a/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
+++ b/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
@@ -114,7 +114,7 @@ initial equation
   assert(eps_nominal > 0 and eps_nominal < 1,
     "eps_nominal out of bounds, eps_nominal = " + String(eps_nominal) +
     "\n  To achieve the required heat transfer rate at epsilon=0.8, set |T_a1_nominal-T_a2_nominal| = "
-     + String(abs(Q_flow_nominal/0.8*CMin_flow_nominal)) +
+     + String(abs(Q_flow_nominal/0.8/CMin_flow_nominal)) +
     "\n  or increase flow rates. The current parameters result in " +
     "\n  CMin_flow_nominal = " + String(CMin_flow_nominal) +
     "\n  CMax_flow_nominal = " + String(CMax_flow_nominal));
--- a/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
+++ b/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
@@ -103,8 +103,7 @@ initial equation
   Q_flow_nominal = m1_flow_nominal*cp1_nominal*(T_a1_nominal - T_b1_nominal);
   if use_Q_flow_nominal then
     Q_flow_nominal = -m2_flow_nominal*cp2_nominal*(T_a2_nominal - T_b2_nominal);
-    eps_nominal = abs(Q_flow_nominal/((T_a1_nominal - T_a2_nominal)*
-      CMin_flow_nominal));
+    eps_nominal = Q_flow_nominal/((T_a1_nominal - T_a2_nominal)*CMin_flow_nominal);
   else
     T_a1_nominal = Medium1.T_default;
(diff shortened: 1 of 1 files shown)


AntoineGautier (changes requested):
The modified model needs a revision note.

It is also a non-backward compatible change.
And I don't see an easy conversion script to automate the refactoring.
The issue is that `Q_flow_nominal` is only loosely defined by its description string `Nominal heat transfer`. In the code it is clear that it is the heat transfer from 1 to 2.
https://github.com/ibpsa/modelica-ibpsa/blob/d1ee2f415d443f60437bf609840afbfa2dfcb829/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo#L97-L103 

However, the model is fully reversible and more often than not users configure it with side 2 being the source and side 1 the load (don't ask me why).
So the change proposed here will trigger this kind of error (as in `Buildings.Examples.VAVReheat.ASHRAE2006`) in number.
```
Error: The following error was detected at time: 0
  eps_nominal out of bounds, eps_nominal = -0.504117
  To achieve the required heat transfer rate at epsilon=0.8, set T_a1_nominal-T_a2_nominal = 41.2746
  or increase flow rates. The current parameters result in 
  CMin_flow_nominal = 5474.46
  CMax_flow_nominal = 17739
  Failed condition: cor.terHea.eps_nominal > 0 and cor.terHea.eps_nominal < 1
```
If we agree on it, we should at least modify the description string of `Q_flow_nominal`, perhaps catch the 2nd Law violation with a specific and more explicit assert statement, and think if there is a conversion script that can be developed for Dymola users at least.

mwetter:
I don't quite see what the benefit is of avoiding the `abs()` value and having a bit a shorter expression but at the cost of a backward incompatible change that affects many models. This equation is only evaluated once in the initial equation section (but not during the simulation) and therefore there would be no change in computing time. Is there any other more important benefit that I miss?

bravache:
> I don't quite see what the benefit is of avoiding the `abs()` value and having a bit a shorter expression but at the cost of a backward incompatible change that affects many models. This equation is only evaluated once in the initial equation section (but not during the simulation) and therefore there would be no change in computing time. Is there any other more important benefit that I miss?

@mwetter I originally did not catch that it would not be backward compatible, but Antoine showed me that the convention on the direction of `Q_nominal_flow` was not enforced, and therefore can have the wrong sign. Given that, I guess it's a small change that we could drop, although we should still fix the error message.
If I were to argue on why we should go forward with this breaking change:
- It would enforce the same convention for all heat exchanger, that `Q_flow_nominal` is the heat from port `*1` to `*2`, or in other words, that it is negative when `*1` is heating.
- It would prevent `T_b1_nominal` and `T_b2_nominal` from taking the wrong value. Right now, if a user input `Q_flow_nominal` with the wrong sign, the calculated outlet temperature of both port will be wrong, since the `deltaT` will be inversed. This is not critical since these values are apparently not used outside of that sizing calculation, but this could lead to confusion.
- Future code might rely on an assumed convention of `Q_flow_nominal` or rely on the nominal outlet temperatures, and this tech debt would come in the way.

bravache:
@mwetter @AntoineGautier Following our meeting conversation, I have reverted the changes in this branch to only include the fix to the error message. I am opening another PR for raising a DeprecationWarning and updating `Q_flow_nominal` description.

mwetter:
@Mathadon : Can you please review and merge if you agree (wrong equation in message of assertion).

Mathadon (approved):

Mathadon:
I cannot merge since the PR has to be approved by @AntoineGautier

AntoineGautier (approved):

PR #1455, opened 2021-03-25
Refactor OccupancySchedule

### LINKED ISSUE #1454: Wrong result with OccupancySchedule
This integrates the change from https://github.com/lbl-srg/modelica-buildings/issues/2369 in the IBPSA library for the model `IBPSA.Controls.SetPoints.OccupancySchedule" so it works with all versions of OPTIMICA.

It is also a cleaner implementation as the same function is used in the initialization and during the events that update the model output.

mwetter:
This closes #1454


mwetter:
@MichaMans : Thanks for the review, I will merge when the tests pass. Yes, it is the same implementation as in `Buildings`, with some minor improvements after I merged it upstream to `IBPSA`.

PR #1456, opened 2021-03-26
Added missing comments for #515

### LINKED ISSUE #515: missing comments in model instances in Airflow.Multizone package
Many models in `Annex60.Airflow.Multizone.Examples` and in `Annex60.Airflow.Multizone.Validation` have instances that have no comment. These missing comments should be added.

mwetter:
This closes #515.

Missing comments have been added to all instances in `Examples` and `Validation`.
Some instance names have been updated to follow the naming convention,
and some documentation strings have been changed for consistency.

In a few models, I also rerouted connections to make the model cleaner.


MassimoCimmino (commented):
This seems out of scope for #515, but I noticed some of the Example and Validation cases use Air Medium models different from `IBPSA.Media.Air`. Is this intentional? If there is a physical reason why `Modelica.Media.Air` could not be used, it might be useful to document it in the examples.

Here are the cases :
- `IBPSA.Airflow.Multizone.Examples.ChimneyShaftNoVolume` uses `Modelica.Media.Air.SimpleAir`;
- `IBPSA.Airflow.Multizone.Examples.ClosedDoors` uses `IBPSA.Media.Specialized.Air.PerfectGas`;
- `IBPSA.Airflow.Multizone.Examples.OneOpenDoor` uses `Modelica.Media.Air.SimpleAir`;

If this is as intended, then the PR is ready to merge.

mwetter:
@MassimoCimmino : Thanks for the review. These are indeed the correct media for these examples because they heat up air to drive a flow, and `IBPSA.Media.Air` assumes `d/dStp = p/pStp;` to decouple the energy and pressure simulation. So that medium won't work for the way these examples are configured.

PR #1457, opened 2021-03-27
Utilize Modelica compliant string allocation

tbeu:
Particularly, avoid the resource leak of the allocated string.


mwetter:
@tbeu : Thanks for the pull request. I will merge through it https://github.com/ibpsa/modelica-ibpsa/pull/1458

PR #1484, opened 2021-05-22
Updated script so it works with Dymola 2022 which by default loads MSL 4.0.0

### LINKED ISSUE #1483: Merge changes for BESTEST with Dymola 2022 to IBPSA
This issue is to merge changes from the Buildings library to the IBPSA library for running the weather data BESTEST.

The issue is caused because of the switch to Dymola 2022 in Buildings, which in `Resources/Data/BoundaryConditions/Validation/BESTEST/generateResults.py` loads first MSL 4.0.0 and then tries to convert the library which fails if run from an automated script (because by default, the conversion is cancelled).
The work-around is the add a statement to unload MSL 4.0.0, which then causes the script to automatically load MSL 3.2.3 as specified in the `uses` statement. Unloading is done using
```python
s.addPreProcessingStatement("eraseClasses({\"Modelica\"});")
```

With this change, the boolean variable `CLEAN_MAT` has also been removed as this is in my opinion not needed and it caused complexity during debugging.

mwetter:
This closes #1483

### DIFF
--- a/IBPSA/Resources/Data/BoundaryConditions/Validation/BESTEST/generateResults.py
+++ b/IBPSA/Resources/Data/BoundaryConditions/Validation/BESTEST/generateResults.py
@@ -27,8 +27,6 @@ import git
 
 # Check if it just implements post-process (from .mat files to Json files)
 POST_PROCESS_ONLY = False
-# Erase old .mat files
-CLEAN_MAT = False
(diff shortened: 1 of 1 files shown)


EttoreZ (approved):
I agree that the CLEAN_MAT is probably not necessary and the other changes are consistent. I tested the script on my machine and runs without problems.

PR #1495, opened 2021-06-08
Issue1389 media steam

khinkelman:
Hi @mwetter, here is the revised steam medium model. The main changes since last time:

1. I merged in the IBPSA/master branch to be up-to-date.
2. The applicable pressure and temperature ranges were reduced to improve accuracy (as previously discussed).
3. Previously, we had been using `p_default` throughout the media model because we had assumed pressure would be constant in applications of this model. However through further testing, this was not the case (e.g., with the pressure reducing valve). To avoid confusion by users and accurately calculate thermodynamic states across the range of possible pressure values, this media model now uses the variable `state.p` in function calls. In applications of the steam medium, we are using (for example) `rhoStd(T_default, p_default)` where possible to decouple the pressure and density calculations throughout simulations. 

If you have any questions or need further revisions from me, please let me know. If acceptable to you, I feel this version of the model can be merged with the IBPSA/Buildings master branches based on the testing completed to date.

Thanks!


mwetter:
@khinkelman : Thanks, I will review over the next few days. I merged it and will further process through https://github.com/ibpsa/modelica-ibpsa/pull/1395

PR #1500, opened 2021-06-11
1436 ma airflowv2

Mathadon:
this PR merges all changes from @kldjonge and merged the master.


Mathadon (commented):
@kldjonge I have reviewed the files up to IBPSA/Airflow/Multizone/Validation/OneWayFlow.mo 

I made a lot of changes myself related to code formatting and documentation. Apart from that there are other issues that have to be resolved. I tried to make a clear distinction between the two by explicitly mentioning what I solved already. Can you have a look at those? Preferably also take these comments with you to fix the models that I did not yet review.

One notable change: I refactored some of the base classes such that they are better readable.

Mathadon:
Pff. there were 35 comments but most of them seem to have been discarded since I made changes to the file that was involved. So much for giving feedback it seems.

IBPSA/Airflow/Multizone/BaseClasses/flowElementData.mo
```diff
@@ -0,0 +1,66 @@
...
+  for j in 1:size(xd, 1) - 1 loop
+    if u > xd[j] then
+      i := j;
+    end if;
+  end for;
+
+  // Extrapolate or interpolate the data
+
+  if i == (size(xd, 1) - 1) then
+    z:=yd[end-1]+(u-xd[end-1])*(yd[end]-yd[end-1])/(xd[end]-xd[end-1]);//Interpolate linearly between last and second-to-last point.
```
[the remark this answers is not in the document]
kldjonge: I based this on CONTAM and  from the documentation it is not very clear how it is handles there but adding the  lineair interpolation at the start does makes sense. I will add it.

kldjonge:
> @kldjonge I have reviewed the files up to IBPSA/Airflow/Multizone/Validation/OneWayFlow.mo
> 
> I made a lot of changes myself related to code formatting and documentation and documented that too as a way of giving feedback. Apart from that there are other issues that have to be resolved. I tried to make a clear distinction between the two by explicitly mentioning what I solved already. Can you have a look at those? Preferably also take these comments with you to fix the models that I did not yet review.
> 

@Mathadon, thanks for the review and especially for checking all the HTML changes. I wasn't aware that the Dymola editor changed all that. I pushed the commits with the changes.

Mathadon:
@kldjonge There are some formatting errors in the mos script. See here for more details:

image
 

Can you have a look at the html formatting problems that are reported here? https://github.com/ibpsa/modelica-ibpsa/pull/1500/checks?check_run_id=2820296563 

I'll fix the unit test formatting problems mentioned here https://github.com/ibpsa/modelica-ibpsa/pull/1500/checks?check_run_id=2820296562
which should fix the errors mentioned here https://github.com/ibpsa/modelica-ibpsa/pull/1500/checks?check_run_id=2820296112 and here https://github.com/ibpsa/modelica-ibpsa/pull/1500/checks?check_run_id=2820296555

Mathadon:
Some unexpected result differences that I accepted:

image

image

Mathadon:
@mwetter not all errors in the first batch of tests are reported at once. Is there any chance to change that in the future such that all errors can be fixed in one run?

IBPSA/Fluid/Sources/Examples/Outside_CpData.mo
```diff
@@ -0,0 +1,81 @@
...
+      points={{-60,10},{-50,10},{-50,-29.8},{-6,-29.8}},
+      color={255,204,51},
+      thickness=0.5,
+      smooth=Smooth.None));
+  connect(weaDat.weaBus, east.weaBus) annotation (Line(
+      points={{-60,10},{-50,10},{-50,-10},{30,-10},{30,8.2},{40,8.2}},
+      color={255,204,51},
+      thickness=0.5,
+      smooth=Smooth.None));
+  annotation (__Dymola_Commands(file="modelica://IBPSA/Resources/Scripts/Dymola/Fluid/Sources/Examples/Outside_CpLowRise.mos"
```
Mathadon, line 51: @kldjonge this script still needs an update

IBPSA/Fluid/Sources/Examples/Outside_CpData.mo
```diff
@@ -0,0 +1,81 @@
...
+Weather data are used for San Francisco, for a period of a week
+where the wind blows primarily from North-West.
+The plot shows that the wind pressure on the north- and west-facing
+facade is positive,
+whereas it is negative for the south- and east-facing facades.
+</p>
+</html>", revisions="<html>
+<ul>
+<li>
+December 22, 2014 by Michael Wetter:<br/>
```
Mathadon, line 66: @kldjonge I think you still need to update the documentation and revision history?

IBPSA/Fluid/Sources/Outside_CpData.mo
```diff
@@ -0,0 +1,119 @@
+﻿within IBPSA.Fluid.Sources;
+model Outside_CpData
+  "Boundary that takes weather data as an input and computes wind pressure for low-rise buildings"
```
Mathadon, line 3: It would be good to adjust this comment (and the one in Outside_CpLowRise) such that the difference between the two models becomes clear.

IBPSA/Fluid/Sources/Outside_CpData.mo
```diff
@@ -0,0 +1,119 @@
...
+
+
+  annotation (defaultComponentName="out",
+    Documentation(info="<html>
+<p>
+This model describes boundary conditions for
+pressure, enthalpy, and species concentration that can be obtained
+from weather data. The model is identical to
+<a href=\"modelica://IBPSA.Fluid.Sources.Outside\">
+IBPSA.Fluid.Sources.Outside</a>,
```
Mathadon, line 62: Explain the difference to Outside_CpLowRise?

Mathadon:
@kldjonge I added some comments for the files that I did not review yet. Were all the comments from before resolved?

kldjonge:
>  I added some comments for the files that I did not review yet. Were all the comments from before resolved?

All comments were resolved. I also resolved the new comments.

Mathadon:
@mwetter this is ready for review! I cannot generate correct reference results since I'm using an older version of Dymola.

IBPSA/Airflow/Multizone/Examples/TrickleVent.mo
```diff
@@ -0,0 +1,119 @@
...
+          {0,20},{0,-10}},          color={191,0,0}));
+  connect(east.ports[1], tabdat_M.port_a)
+    annotation (Line(points={{-40,-30},{-30,-30}}, color={0,127,255}));
+  connect(tabdat_M.port_b,room. ports[1])
+    annotation (Line(points={{-10,-30},{8,-30},{8,-20}},  color={0,127,255}));
+  connect(tabdat_V.port_a,room. ports[2]) annotation (Line(points={{30,-30},{14,
+          -30},{14,-20},{12,-20}}, color={0,127,255}));
+  connect(tabdat_V.port_b, west.ports[1])
+    annotation (Line(points={{50,-30},{60,-30}}, color={0,127,255}));
+  annotation (__Dymola_Commands(file="modelica://IBPSA/Resources/Scripts/Dymola/Airflow/Multizone/Examples/OneEffectiveAirLeakageArea.mos"
```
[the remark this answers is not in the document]
kldjonge: I'll add it immediately.

mwetter:
Todo
- [ ] Make sure regression tests work
- [ ] Test integration with Buildings library

kldjonge:
https://github.com/ibpsa/modelica-ibpsa/pull/1500/commits/f38428d0fc091976f5d04545f09f1937bc5fa8e9
Its not entirely clear, was it not consistent then before? 

It would indeed make sense to be consistent within the package, including the fluid package. So using k for m_flow and C for V_flow then? Or do you propose something else?

mwetter:
@kldjonge : Orifice is using 
image
Let me change this to using \dot V = C dp^m, and the other models as \dot m = k dp^m. It turns out that C or k is not used in the old implementation of `Airflow.Multizone` as a user parameter, so this change won't cause any backward incompatibility issues, and then we have the same notation as in `IBPSA.Fluid`. In comparison, CONTAM uses C for the power law for volume and for mass flow rate (p. 52 in https://nvlpubs.nist.gov/nistpubs/TechnicalNotes/NIST.TN.1887.pdf) but I rather use different symbols to distinguish the different coefficients.

kldjonge:
> 

I agree. Differentiating between k and C is better. I think it is clear now.

mwetter:
@kldjonge : Can you please make the documentation of the incidence angle `incAng` in `Fluid.Sources.Outside_CpData` clearer. I think it is relative to north, rather than the surface, because the surface azimuth is also a parameter. It would be good to have a figure that explains these angles. Currently we have the documentation
azi: Surface azimuth (South:0, West:pi/2) [rad]
but it is not clear how the `incAng` are defined.

Also, could you please add a small validation that makes sure the angles are right. The current example is symmetric relative to the North-South axis, so it won't pick up an error in the sign of the angle.

I hope to get back to this model on Thursday or Friday

kldjonge:
> 

@mwetter , I updated the documentation of and added an additional example with an asymmetrical Cp profile. I hope it is clearer now.

mwetter:
@kldjonge : Thanks for the figure. I don't understand how the angles relate to the model, and what the numbers in the figure means as there are boxes "Wind Direction", "Incidence Angle" and "Surface Azimuth". Should the number be equal to the variable variable `winSurAng`? This must be made much clearer. I added the definition of `winDir` and `azi` as it is used in the IBPSA library. It is still not clear what the angles in `incDir` mean, and what the role of `azi` is. Can you try to explain this in the info section of the model `Outside_CpData`.

I also added a validation model `IBPSA.Fluid.Sources.Validation.Outside_CpData_Directions` where I wanted to have 4 surfaces, one in each direction. The value for `Cp` is such that it is 1 if the wind is for the direction of the surface (normal to the surface) and 0 otherwise. However, the results are wrong. I would expect that all trajectories for `pWin` look similar, but shifted by t=0.25. There is something wrong with the model or my understanding of what these parameter mean. Can you please clarify that by improving the info section of the model.

Also, I tried to understand what the model `IBPSA.Fluid.Sources.Examples.Outside_CpData_Angles` is about, but its info section is just a copy of `IBPSA.Fluid.Sources.Examples.Outside_CpData` which is not applicable. Can you please finish that example and also follow the naming convention for the instance names.

kldjonge:
@mwetter,

The model and (internal) angles work exactly the same as the existing Outside_CpLowrise. Here also only Cp,0 is a parameter which is the value IF the wind if normal to the wall.

With regards to this model, only the equation for Cp,act is changed. But the user is still asked to input Cp values relative to the normal of the surface (with the normal being 0). In that sense, some of the refactoring you proposed give the same variables a different name for Outside_CpLowrise and Outside_CpData

Because of the Cp values being relative to the surface normal, the new examples doesn't make sense. To have 1  when the wind is normal to the surface the Cp data for each surface stays the same: {1,0,0,1}. I changed it.

mwetter:
@kldjonge : Thanks for the clarification. That makes sense now. I further revised the models. Also, there is no need to require the Cp value for 360 deg, and then enforcing it to be the same as the one for 0 deg. I updated the model so that the last value is no longer needed, but rather added automatically in a protected parameter.
I also added a new figure that shows an example for how to specify the profile for the wind pressure coefficient. I removed the old figure that is no longer needed.
When all tests run, this is ready for a last review/approval.

mwetter (approved):

mwetter:
@kldjonge @Mathadon : From my end, this is ready to be merged into the master. It also passed all tests in the Buildings library and has been merged to the master.

Do you want to do a last review as I made a lot of changes as part of my review?

kldjonge:
@mwetter, I kept track off the changes you made. I’m fine with all of them. 

Thank you for the effort Michael!

Mathadon:
@mwetter thanks for the finishing touches. No further review from my side is required!

PR #1503, opened 2021-07-12
Replace PlugFlowPipe outlet from vectorized to single port

### LINKED ISSUE #1494: Resolve PlugFlowPipe compatibility with PartialTwoPortInterface
Linked issue (with further discussion): https://github.com/lbl-srg/modelica-buildings/issues/2449

## Issue
The `IBPSA.Fluid.FixedResistances.PlugFlowPipe` model extends from `PartialTwoPortVector`, which uses a vectorized outlet port. This inheritance is at odds with the other pipe models in the `FixedResistances` packages, and prevents this model from being used to redeclare `PartialTwoPortInterface` objects. 

## Context

The model `IBPSA.Fluid.FixedResistances.PlugFlowPipe` is a wrapper for the `IBPSA.Fluid.FixedResistances.BaseClasses.PlugFlowCore` that incorporates a mixing volume `vol` and a vectorized outlet port `ports_b[]`.
file:///home/user/Pictures/Screenshot%20from%202021-06-02%2013-30-21.png

The mixing volume is used to model the heat capacity of the pipe walls and the outlet port is vectorized to prevent the need to add a splitter or mixing volume after the model when the outlet is branching out, therefore making the integration of the model more computationally efficient.

## Proposed solution

The solution proposed is to deprecate the `PlugFlowPipe` model as it currently exist, and promote the `BaseClasses.PlugFlowCore` in its place. The mixing volume would either be integrated into the `PlugFlowCore` model, but connected to a non-vectorized port (similar to the implementation of `Buildings.Fluid.FixedResistances.Pipe` in MBL) or removed entirely (if we can validate that its impact on outputs are negligible in most use cases).

This solution would simplify the use of the plug flow model, and make its integration more explicit and clear (i.e. the model only represent a pipe, and not a pipe + mixing volume). This solution also remove the redundancy of having two models with similar usage, name and behavior (`PlugFlowPipe` and `BaseClasses.PlugFlowCore`).

bravache:
Closes #1494 

This PR addresses the problem of compatibility of `IBPSA.Fluid.FixedResistances.PlugFlowPipe` with other pipe classes, by changing its outlet port from a vectorized port to a single port.

The PR brings ***non-backward compatible*** changes, for which a conversion script is provided for Dymola users.

## Validation results
The following example and validation models uses the `PlugFlowPipe` class and were updated to reflect that change:

- `IBPSA.Fluid.FixedResistances.Examples.PlugFlowPipe`
- `IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.FlowReversal`
- `IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.PlugFlowAIT`
- `IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.PlugFlowULg`
- `IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.TransportWaterAir`

For most of the models, the only change required was to remove the parameter `nPorts` from the parameterization of `PlugFlowPipe` and update the outlet connection. Additionally, for `PlugFlowAIT`, two splitters had to be introduced to replace the connections out of `pip1` and `pip5`.
The results of all example/validation models remained unchanged, with the exception of `PlugFlowAIT` which only shows an error in the initial step of the simulation, but no error for the remainder of the runtime.
image

On a side note, the pressure boundary condition for the `FlowReversal` example was updated to 5bar, to avoid negative pressure at the inlet when the flow is reversed. This does not change the expected results, but make the example more robust.

## Backward compatibility
This change is non-backward compatible. A conversion script was written to redirect previous objects to the obsolete class.


bravache:
@AntoineGautier Please take a look and let me know if I've missed something. Thanks!

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
```diff
@@ -274,8 +282,6 @@ The thermal capacity of the pipe wall is implemented as a mixing volume
 of the fluid in the pipe, of which the thermal capacity
 is equal to that of the pipe wall material.
 In addition, this mixing volume allows the hydraulic separation of subsequent pipes.
-Thanks to the vectorized implementation of the (design) outlet port,
-splits and junctions of pipes can be handled in a numerically efficient way.
 <br/>
 This mixing volume is not present in the
 <a href=\"modelica://IBPSA.Fluid.FixedResistances.BaseClasses.PlugFlowCore\">PlugFlowCore</a> model,
```
AntoineGautier, line 287: I would delete that sentence as we agreed that the mixing volume was necessary to represent the effect of the pipe wall capacitance (not only for decoupling the DAE between multiple instances). 
Instead I would add something like:

> Note that in order to model a branched network it is recommended to use `IBPSA.Fluid.FixedResistances.Junction` at each junction and to configure that component with a state (see for instance
<a href=\"modelica://IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.PlugFlowAIT\">
IBPSA.Fluid.FixedResistances.Validation.PlugFlowPipes.PlugFlowAIT</a>). 
This will avoid the numerical Jacobian that is otherwise created when the inlet ports of two instances of the plug flow model are connected together.

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
```diff
@@ -1,7 +1,7 @@
 within IBPSA.Fluid.FixedResistances;
 model PlugFlowPipe
   "Pipe model using spatialDistribution for temperature delay"
-  extends IBPSA.Fluid.Interfaces.PartialTwoPortVector;
```
AntoineGautier, line 4: `IBPSA.Fluid.Interfaces.PartialTwoPortVector` may be retired into Obsolete as well as it was only used by the plug flow model.
To be validated with the IBPSA team.

AntoineGautier (commented):
@bravache Please see my inline comments.
Also, and as proposed in the issue page, I would merge `PlugFlowCore` and `PlugFlowPipe` into one, making the fluid volume conditional to an advanced Boolean parameter which would be true by default. The two models are too similar in my opinion to justify keeping them both.

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
bravache: I might keep it like that since it predates the `PlugFlowPipe` model. Also, in the same package, there is also the `PartialFourPortParallel` which is not in used.

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
bravache: I've reworded the whole documentation to incorporate part of the `PlugFlowCore` doc. Let me know if that works.

bravache:
I've moved the `PlugFlowCore` to Obsolete, and expanded its content directly in `PlugFlowPipe`. I've made the mixing volume conditional on the value of `have_PipCap`, and was also able to reuse the former `PlugFlowCore` validation results to verify the results of the model when that Boolean is false.

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
bravache: I filed a ticket so that it can be discussed before implemented here: #1504

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
```diff
@@ -157,19 +198,38 @@ initial equation
...
+    annotation (Line(points={{-30,0},{-20,0}}, color={0,127,255}));
+  connect(heaLos_b.port_a,del. port_b)
+    annotation (Line(points={{40,0},{30,0}}, color={0,127,255}));
+  connect(heaLos_b.heatPort, heatPort) annotation (Line(points={{50,10},{50,86},
+          {0,86},{0,100}}, color={191,0,0}));
+  connect(heaLos_a.heatPort, heatPort) annotation (Line(points={{-70,10},{-70,
+          86},{0,86},{0,100}}, color={191,0,0}));
+  connect(heaLos_a.port_b, port_a)
+    annotation (Line(points={{-80,0},{-100,0}}, color={0,127,255}));
+  if have_PipCap then
```
AntoineGautier, line 225: Remove that statement. If `vol` is conditionally removed then the following connect clauses are also automatically removed at translation.
Only keep the `else` branch as `if not have_pipCap then`. 
Or even better, use a conditional instance of `IBPSA.Fluid.FixedResistances.LosslessPipe` instead to model that case, and remove all conditional statements from the equation section. This makes the model structure more visible.

AntoineGautier (changes requested):
I just have one minor change request and a suggestion.
Otherwise the changes look good to me.

IBPSA/Fluid/FixedResistances/Examples/PlugFlowPipeNoMix.mo
```diff
@@ -0,0 +1,26 @@
+within IBPSA.Fluid.FixedResistances.Examples;
+model PlugFlowPipeNoMix
```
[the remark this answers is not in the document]
bravache: Done. Although I find merging two variation of the same examples a bit awkward since it requires to duplicate part or all of the other classes that are connected to it. I know that they are work around it (maybe creating a base class, or by vectorizing the other components), but it does create example models that are not as simple to read.

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
bravache: Done!

bravache:
@Mathadon @nytschgeusen @cvering @mwetter 
This PR is ready to merge. Antoine had a couple of comments left, but those were addressed and he is now on break, so I think it is reasonable to move forward.

Let me know if you have any comments. Otherwise, this will be merged on 09/03/2021.

## Content

The main change introduce is changing the vectorized outlet port of `PlugFlowPipe` to be a single outlet port, which is consistent with other pipe model in the same package.

Additionally, the `Fluid.FixedResistances.BaseClasses.PlugFlowCore` and `Fluid.FixedResistances.PlugFlowPipe` are combined into a single model (previously the latter was a wrapper around the former), and the outlet mixing volume is made conditional on a boolean parameter `have_pipCap`.

The previous classes were moved to `Obsolete` and a conversion script was written to update `PlugFlowCore` and `PlugFlowPipe` instances to these obsolete classes.

The existing unit tests stored output had to be updated to reflect the change in translation statistics, but the timeseries results remained unchanged.

Mathadon:
@bravache thanks for the summary! I have forwarded this to a colleague who uses this model more actively than I do.

mwetter:
Michael to merge after last review.

mwetter:
Todo
- [ ] Consider removing constant `computeFlowResistance` in the new pipe model. This would require a refactoring in the Buildings library for the distributed pipe model, which would in turn allow eliminating states.

mwetter:
@Mathadon : This is ready to merge from my point of view, but I think it would be good to have another approvals as during the clean up, I changed in `PlugFlowPipe` most instances to be protected, and instead exposed and/or introduced new variables `QEnv_flow` (for heat exchange with environment), `dp`, `m_flow` and `v`. 
`QEnv_flow = heatPort.Q_flow` was introduced because in Buildings (and maybe soon in IBPSA) there is a discretized version of this model, in which `QEnv_flow = sum(heatPort[:].Q_flow)`. By introducing this variable, both models have the same high level variables.

Mathadon:
@mwetter I'll try to review this tomorrow!

IBPSA/Fluid/FixedResistances/BaseClasses/PlugFlowTransportDelay.mo
```diff
@@ -48,8 +49,7 @@ initial equation
   t0 = time;
 
 equation
-  u = m_flow/(rho*(dh^2)/4*Modelica.Constants.pi)/length;
-
+  u = m_flow * conUM;
```
Mathadon, line 52: Great!

IBPSA/Fluid/FixedResistances/Examples/PlugFlowPipe.mo
```diff
@@ -30,40 +29,90 @@ model PlugFlowPipe "Simple example of plug flow pipe"
...
+    annotation (Placement(transformation(extent={{40,10},{60,30}})));
   IBPSA.Fluid.Sensors.TemperatureTwoPort senTemIn(
     redeclare package Medium = Medium,
     m_flow_nominal=1,
     T_start=323.15) "Temperature sensor"
-    annotation (Placement(transformation(extent={{-30,-10},{-10,10}})));
+    annotation (Placement(transformation(extent={{-30,10},{-10,30}})));
+  Sensors.TemperatureTwoPort senTemInNoMix(
+    redeclare package Medium = Medium,
+    m_flow_nominal=1,
```
Mathadon, line 55: this could use `tau=0`?

IBPSA/Fluid/FixedResistances/Examples/PlugFlowPipe.mo
```diff
@@ -30,40 +29,90 @@ model PlugFlowPipe "Simple example of plug flow pipe"
...
+    thickness=0.0032,
+    initDelay=true,
+    m_flow_start=1,
+    rhoPip=8000,
+    T_start_in=323.15,
+    T_start_out=323.15) "Pipe"
+    annotation (Placement(transformation(extent={{0,-30},{20,-10}})));
+  Sensors.TemperatureTwoPort senTemOutNoMix(
+    redeclare package Medium = Medium,
+    m_flow_nominal=1,
```
Mathadon, line 76: tau=0?

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
```diff
@@ -151,25 +128,132 @@ protected
...
+    final m_flow_nominal=m_flow_nominal,
+    final V=if rho_default > 500 then VEqu else VEqu/1000,
+    final nPorts=2,
+    final T_start=T_start_out,
+    final energyDynamics=Modelica.Fluid.Types.Dynamics.FixedInitial,
+    final mSenFac=if rho_default > 500 then 1 else 10) if have_pipCap
+    "Control volume connected to port_b. Represents equivalent pipe wall thermal capacity."
+    annotation (Placement(transformation(extent={{70,20},{90,40}})));
+
+  LosslessPipe noMixPip(
```
Mathadon, line 148: why not use a simple connection instead?

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowAIT.mo
```diff
@@ -261,6 +255,36 @@ model PlugFlowAIT
         rotation=180,
         origin={46,56})));
 
+  IBPSA.Fluid.FixedResistances.Junction splPip1(
+    redeclare package Medium = Medium,
+    m_flow_nominal={pip1.m_flow_nominal,pip5.m_flow_nominal,pip4.m_flow_nominal},
+    from_dp=true,
+    linearized=true,
```
Mathadon, line 262: this has no purpose if `dp_nominal=0`?

IBPSA/Obsolete/Fluid/FixedResistances/BaseClasses/Validation.mo
```diff
@@ -0,0 +1,153 @@
+within IBPSA.Obsolete.Fluid.FixedResistances.BaseClasses;
+package Validation "Collection of validation models"
+  extends Modelica.Icons.ExamplesPackage;
+
+  model PlugFlowCore "Simple example of plug flow pipe core"
```
Mathadon, line 5: It would be better to split this package in multiple .mo files?

Mathadon (changes requested):
My largest concern is that the model IBPSA.Fluid.FixedResistances.PlugFlowPipe is not yet symmetric: a mixing volume consists at just one of the two pipe outlets/inlets. I propose to use two mixing volumes, each with half the mass.

IBPSA/Fluid/FixedResistances/Examples/PlugFlowPipe.mo
Mathadon: Same for all other examples to be honest: the filter slows down the model and skews the results.

IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo
bravache: I believe I've written it like that to conserve a symmetrical syntax between the case with the mixing volume (`have_pipCap`) and the case without a mixing volume (`not have_pipCap`).
Also, in the graphical editor of Dymola, you get an annotation in the hover tooltip when a component is optional (screenshot below), but not when a connect statement is optional, which makes that structure a bit more explicit.
conditionalPip

That being said, if there is a computational or syntax issue with that format, I can switch back to an optional connect.

IBPSA/Fluid/FixedResistances/Examples/PlugFlowPipe.mo
bravache: Added for all temperature sensors.

IBPSA/Fluid/FixedResistances/Validation/PlugFlowPipes/PlugFlowAIT.mo
bravache: Indeed! Good catch. Removed for both junctions in this model.

IBPSA/Obsolete/Fluid/FixedResistances/BaseClasses/Validation.mo
bravache: Good catch! Updated

bravache (commented):
Thanks Filip, I've reviewed your inline comments.
With regard with the symmetry of the model, I was reluctant to introduce it because when I implement it naively, models that have two `PlugFlowPipe` in series become structurally singular because of the two connected mixing volumes.
I am discussing this with Michael and will update you if we come up to a solution.

Edit: Actually you can disregard that message, the structural singularity was a different issue from my rushed implementation.

mwetter:
@bravache : Above, there is the declaration:
```
Sensors.TemperatureTwoPort senTemInNoMix(
    redeclare package Medium = Medium,
    m_flow_nominal=1,
```
This looks wrong. All models use `m_flow_nominal = 1` kg/s, but the source uses 3 kg/s. This should be made consistent.

bravache:
@mwetter 
Following our conversation and the test that showed that the computational impact was low in a sufficiently detailed model, I've implemented the symmetry in the `PlugFlowPipe` class. I've also added an advanced boolean flag `have_symmetry` which allows disabling the inlet mixing volume.
In addition, I've made the model extends from `PartialTwoPortInterface` rather than `PartialTwoPort`, since the former is more constraining, and adds `show_T` as a parameter.

PR #1509, opened 2021-07-24
added public output for XiSat in TWetBul_TDryBulXi.mo and TWetBul_TDr…

### LINKED ISSUE #1508: move XiSat to public in IBPSA.Utilities.Psychrometrics.TWetBul_TDryBulXi
The variable `XiSat` is protected in `IBPSA.Utilities.Psychrometrics.TWetBul_TDryBulXi`. This variable is accessed in `IDEAS.Fluid.HeatExchangers.IndirectEvaporativeHex`. Consequently OpenModelica complains about accessing a protected variable (see https://github.com/open-ideas/IDEAS/issues/1225). Since  `IndirectEvaporativeHex` requires both the wet bulb temperature and `XiSat`, manually computing `XiSat` instead of accessing the variable in `TWetBul_TDryBulXi` would cause code duplication, which I'd like to avoid. Therefore I propose to make this variable public in `TWetBul_TDryBulXi` and in `TWetBul_TDryBulPhi`.

Mathadon:
…yBulPhi.mo

closes #1508

### DIFF
--- a/IBPSA/Utilities/Psychrometrics/TWetBul_TDryBulPhi.mo
+++ b/IBPSA/Utilities/Psychrometrics/TWetBul_TDryBulPhi.mo
@@ -30,14 +30,17 @@ block TWetBul_TDryBulPhi
     final unit="K",
     min=0) "Wet bulb temperature"
     annotation (Placement(transformation(extent={{100,-10},{120,10}})));
-
+  Modelica.Blocks.Interfaces.RealOutput XiSat(start=0.01,
+                                      nominal=0.01)
+    "Water vapor mass fraction at saturation"
+    annotation (Placement(transformation(extent={{100,-50},{120,-30}})));
 protected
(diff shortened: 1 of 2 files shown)


mwetter (changes requested):
This implementation is not (yet) good: Both blocks only compute `XiSat` if the parameter `approximateWetBulb = true`. Computing `XiSat` introduces a nonlinear equation if `approximateWetBulb = false`, which is why by default this is `true`.

The issue is that whenever a user selects `approximateWetBulb = true` the new output is `0`, so users don't get a correct output for this model configuration.

I am not sure yet if other models would benefit from this change. If not, would it be better if IDEAS were to extend this model in a protected model, set `final approximateWetBulb = false` in the `extends` statement and add this output?

(Also, if the change were in IBPSA, the unit tests should be changed to add this output.)

Mathadon:
Ok, makes sense! I'll fix it in IDEAS.

PR #1537, opened 2021-10-15
Move plugflowpipe to baseclasses

bravache:
The main change of this PR is to move the content of `IBPSA/Fluid/FixedResistances/PlugFlowPipe.mo` to `IBPSA/Fluid/FixedResistances/BaseClasses/PlugFlowPipe.mo` and have the internal pipe resistance element `res` replaceable.
The exposed class has the same behavior as previously, and the baseclass can be reused to replace the pipe resistance with a `LosslessPipe`.


mwetter:
@Baptiste Ravache ***@***.***> : do you also want to add the
discretized plug flow pipe or is there a reason to omit it?

PR #1558, opened 2021-11-25
Included ModelicaAllocateString.h in cryptographicsHash.c

mwetter:
This integrates https://github.com/lbl-srg/modelica-buildings/pull/2784
It avoids a wrong memory access in OpenModelica

### DIFF
--- a/IBPSA/Resources/C-Sources/cryptographicsHash.c
+++ b/IBPSA/Resources/C-Sources/cryptographicsHash.c
@@ -24,6 +24,7 @@ A million repetitions of "a"
 /* for uint32_t */
 #include <stdint.h>
 #include "cryptographicsHash.h"
+#include "ModelicaUtilities.h"
 
 
 #define rol(value, bits) (((value) << (bits)) | ((value) >> (32 - (bits))))


Mathadon:
Interesting fix.. 

In https://github.com/open-ideas/IDEAS/issues/1225#issuecomment-884737144 we had a case where we had to remove `ModelicaUtilities.h` to fix a bug, and here it should be added. See also this related post by @casella : https://github.com/open-ideas/IDEAS/issues/1225#issuecomment-885588542 

I'm ok with merging this but there may be an underlying problem that may need to be looked into to avoid similar problems in the future with OM.

casella:
Wait, see discussion [here](https://github.com/lbl-srg/modelica-buildings/pull/2784)

Mathadon:
Happy to proceed with this after the feedback from @casella here: https://github.com/lbl-srg/modelica-buildings/pull/2784#issuecomment-979414713 !

casella:
> Happy to proceed with this after the feedback from @casella here: [lbl-srg/modelica-buildings#2784 (comment)](https://github.com/lbl-srg/modelica-buildings/pull/2784#issuecomment-979414713) ! This can be merged.

@Mathadon, you need to explicitly post a positive review to allow merging in, I guess. GitHub is not smart enough to read between the lines of your comment 😄

casella:
BTW, the fix worked fine in Buildings, see [report](https://libraries.openmodelica.org/branches/history/master/2021-11-25%2011:41:34..2021-11-25%2015:40:12.html).

Mathadon (approved):

Mathadon:
@casella Fair point, I guess that I got used to @mwetter using his admin privileges 😅

PR #1565, opened 2021-12-08
Updated BuildingsPy to BuildingsPy@v3.0.0

mwetter:
This is the version of BuildingsPy that likely is going to be released as 3.0.0
*Update*: Changed to BuildingsPy v3.0.0


mwetter:
@Mathadon : can you please review so we can merge this PR.

Mathadon (approved):

PR #1583, opened 2022-02-07
Corrected wrong port name

mwetter:
This corrects the name of a port. The old implementation used a port name that does not exist.


mwetter:
@Mathadon : Can you please review (only 1 file) and merge. This breaks the OpenModelica CI testing. In the future, we will catch this earlier once https://github.com/ibpsa/modelica-ibpsa/pull/1581 is completed and merged

Mathadon (approved):

PR #1589, opened 2022-03-04
Changed tab for massDynamics and added assertion

### LINKED ISSUE #1542: Validity range Media.Air
# Problem
At our institute (@martinkremer, @larissakuehn) we came accross the issue that the simplifications of `IBPSA.Media.Air` might lead to significant temperature-related and energy-related deviations in comparison to `Modelica.Media.Air.MoistAir`.  
For this purpose I created an example library containing two examples:
1. `CompareMediaInSimpleCircuits`: Focus on temperature development (constant negative Q_flow as boundary condition).  
  Leads at simulation stop time to a temperature difference of `27.07 °C - 14.59 °C = 12.48 K`
1. `CompareMediaInSimpleCircuitsControlled`: Focus on energy consumption (controlled Q_flow to reach ).  
  Leads at simulation stop time to a relative energy-consumption difference of `(46885.7-30507.3) J / ((46885.7+30507.3) J / 2) = 42.3 %`

Both values are extremly significant deviations. 
Of course we hope to not have overseen anything while modeling this minimal example library.

The volume starts with a temperature of 50 °C and a low relative humidity and ends with a temperature of around 20 °C and a higher relative humidity (no water added).

I added the library including a ReadMe.txt:
[CompareMediaMoistAir.zip](https://github.com/ibpsa/modelica-ibpsa/files/7442163/CompareMediaMoistAir.zip)

# Solution / Discussion
In order to create faster simulations with the IBPSA media compared to MSL media, the simplifications make sense. However, eventually a validity range (temperature, moisture/humidity) should be set and asserted. Or at least a strong indication (bold, red font or something similar) should be added to the documentation of `IBPSA.Media.Air`.

What do you think @mwetter?  
We would welcome any discussion /explenation since this topic is probably not new to you.

mwetter:
This closes #1542

### DIFF
--- a/IBPSA/Fluid/HeatExchangers/ActiveBeams/BaseClasses/Convector.mo
+++ b/IBPSA/Fluid/HeatExchangers/ActiveBeams/BaseClasses/Convector.mo
@@ -24,8 +24,8 @@ model Convector "Heat exchanger for the water stream"
     "Type of energy balance: dynamic (3 initialization options) or steady state"
     annotation(Evaluate=true, Dialog(tab = "Dynamics", group="Equations"));
   parameter Modelica.Fluid.Types.Dynamics massDynamics=energyDynamics
-    "Type of mass balance: dynamic (3 initialization options) or steady state"
-    annotation(Evaluate=true, Dialog(tab = "Dynamics", group="Equations"));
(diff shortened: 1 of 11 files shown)


mwetter:
Closed, replaced by https://github.com/ibpsa/modelica-ibpsa/pull/1591

PR #1591, opened 2022-03-08
Issue1542. Remove massDynamics

### LINKED ISSUE #1542: Validity range Media.Air

mwetter:
This closes #1542 

This is based on https://github.com/ibpsa/modelica-ibpsa/pull/1589 and makes massDynamics final, or removes it, except in the MixingVolume.
This is to address https://github.com/ibpsa/modelica-ibpsa/issues/1542#issuecomment-1059329392

### DIFF
--- a/IBPSA/Fluid/Geothermal/Borefields/BaseClasses/Boreholes/BaseClasses/InternalHEXTwoUTube.mo
+++ b/IBPSA/Fluid/Geothermal/Borefields/BaseClasses/Boreholes/BaseClasses/InternalHEXTwoUTube.mo
@@ -19,7 +19,7 @@ model InternalHEXTwoUTube
     final tau4=VTubSeg*rho4_nominal/m4_flow_nominal,
     vol1(
       final energyDynamics=energyDynamics,
-      final massDynamics=massDynamics,
+      final massDynamics=energyDynamics,
       final prescribedHeatFlowRate=false,
       final allowFlowReversal=allowFlowReversal1,
       final m_flow_small=m1_flow_small,
@@ -27,14 +27,14 @@ model InternalHEXTwoUTube
       final mSenFac=mSenFac),
     vol2(
       final energyDynamics=energyDynamics,
-      final massDynamics=massDynamics,
+      final massDynamics=energyDynamics,
       final prescribedHeatFlowRate=false,
       final m_flow_small=m2_flow_small,
       final V=VTubSeg,
       final mSenFac=mSenFac),
     vol3(
       final energyDynamics=energyDynamics,
-      final massDynamics=massDynamics,
+      final massDynamics=energyDynamics,
       final prescribedHeatFlowRate=false,
       final allowFlowReversal=allowFlowReversal3,
       final m_flow_small=m3_flow_small,
@@ -42,7 +42,7 @@ model InternalHEXTwoUTube
       final mSenFac=mSenFac),
     vol4(
       final energyDynamics=energyDynamics,
-      final massDynamics=massDynamics,
+      final massDynamics=energyDynamics,
       final prescribedHeatFlowRate=false,
       final m_flow_small=m4_flow_small,
       final V=VTubSeg,
@@ -253,6 +253,12 @@ International Journal Of Energy Research, 35:312-320, 2011.
 </p>
 </html>", revisions="<html>
 <ul>
+<li>
+March 7, 2022, by Michael Wetter:<br/>
+Removed <code>massDynamics</code>.<br/>
+This is for
+<a href=\"https://github.com/ibpsa/modelica-ibpsa/issues/1542\">#1542</a>.
+</li>
 <li>
 July 10, 2018, by Alex Laferri&egrave;re:<br/>
 Updated documentation following major changes to the IBPSA.Fluid.HeatExchangers.Ground package.
--- a/IBPSA/Fluid/Humidifiers/SprayAirWasher_X.mo
+++ b/IBPSA/Fluid/Humidifiers/SprayAirWasher_X.mo
@@ -26,8 +26,8 @@ model SprayAirWasher_X
 
   // Dynamics
   parameter Modelica.Fluid.Types.Dynamics massDynamics = Modelica.Fluid.Types.Dynamics.SteadyState
-    "Type of mass balance: dynamic (3 initialization options) or steady state, must be steady state if energyDynamics is steady state"
-    annotation(Evaluate=true, Dialog(tab = "Advanced", group="Dynamics"));
+    "Type of mass balance: dynamic (3 initialization options) or steady state"
+    annotation(Evaluate=true, Dialog(tab = "Dynamics", group="Equations"));
 
   // Set maximum to a high value to avoid users mistakenly entering relative humidity.
   Modelica.Blocks.Interfaces.RealInput X_w(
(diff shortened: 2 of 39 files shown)


mwetter:
@Mathadon : Can you please review this before I assign it to others to review as well because it is non-backward compatible. This PR sets `final massDynamics=energyDynamics`, except in the `MixingVolume` and `MixingVolumeMoistAir` (which check for a wrong configuration with an assertion).

I assume you have a good use case the leave the `massDynamics` parameter in the mixing volumes, as this is the only place where it is left, and I don't have any need to keep this parameter accessible by the user.

Please let me know if you see any issues, as I am working on testing and integrating this in Buildings via https://github.com/lbl-srg/modelica-buildings/pull/2911

IBPSA/Fluid/Geothermal/Borefields/BaseClasses/Boreholes/BaseClasses/InternalHEXTwoUTube.mo
```diff
@@ -19,30 +19,30 @@ model InternalHEXTwoUTube
...
-      final massDynamics=massDynamics,
+      final massDynamics=energyDynamics,
       final prescribedHeatFlowRate=false,
       final allowFlowReversal=allowFlowReversal3,
       final m_flow_small=m3_flow_small,
       final V=VTubSeg,
       final mSenFac=mSenFac),
     vol4(
       final energyDynamics=energyDynamics,
-      final massDynamics=massDynamics,
```
Mathadon, line 45: Good catch..

IBPSA/Fluid/Humidifiers/SprayAirWasher_X.mo
```diff
@@ -12,9 +12,7 @@ model SprayAirWasher_X
       final QMax_flow = Modelica.Constants.inf,
       final QMin_flow = -Modelica.Constants.inf,
       final mWatMax_flow = mWatMax_flow,
-      final mWatMin_flow = 0,
-      final energyDynamics = Modelica.Fluid.Types.Dynamics.SteadyState,
```
Mathadon, line 16: Is this removal intentional?

IBPSA/Fluid/Humidifiers/SteamHumidifier_X.mo
```diff
@@ -10,9 +10,7 @@ model SteamHumidifier_X
       final QMax_flow = Modelica.Constants.inf,
       final QMin_flow = -Modelica.Constants.inf,
       final mWatMax_flow = mWatMax_flow,
-      final mWatMin_flow = 0,
-      final energyDynamics = Modelica.Fluid.Types.Dynamics.SteadyState,
```
Mathadon, line 14: Assuming this removal is intentional.

IBPSA/Fluid/Interfaces/TwoPortHeatMassExchanger.mo
```diff
@@ -17,10 +17,7 @@ model TwoPortHeatMassExchanger
   // Dynamics
   parameter Modelica.Fluid.Types.Dynamics energyDynamics=Modelica.Fluid.Types.Dynamics.DynamicFreeInitial
     "Type of energy balance: dynamic (3 initialization options) or steady state"
-    annotation(Evaluate=true, Dialog(tab = "Dynamics", group="Equations"));
-  parameter Modelica.Fluid.Types.Dynamics massDynamics=energyDynamics
```
Mathadon, line 21: I don't see the need for removing massDynamics but in case I ever see a need for a separate parameter in this model then we can still add it back :)

Mathadon (approved):
@mwetter thanks for doing the heavy lifting on this!

IBPSA/Fluid/Humidifiers/SprayAirWasher_X.mo
mwetter: Yes, this is intentional as `energyDynamics` is assigned in the base class now (https://github.com/ibpsa/modelica-ibpsa/pull/1591/files#diff-70ded3e6ed7a3a11eee46a4517919a9762e70e4855632fde06dd590b2258ddcfR35). Previously, it was missing in the base class, and extending models corrected this by assigning it one level higher.

IBPSA/Fluid/Humidifiers/SteamHumidifier_X.mo
mwetter: Yes, this is intentional as `energyDynamics` is assigned in the base class now (https://github.com/ibpsa/modelica-ibpsa/pull/1591/files#diff-70ded3e6ed7a3a11eee46a4517919a9762e70e4855632fde06dd590b2258ddcfR35). Previously, it was missing in the base class, and extending models corrected this by assigning it one level higher.

IBPSA/Fluid/Interfaces/TwoPortHeatMassExchanger.mo
mwetter: Yes, I also feel better add it back if there is a use case. Removing it makes it simpler for the non-expert user to understand.

mwetter:
@nytschgeusen @cvering : Can you please review this backward incompatible change so we can go ahead and merge it if there are no more changes needed.

IBPSA/Fluid/Humidifiers/SteamHumidifier_X.mo
cvering: This change is reasonable. Thanks for all changes.

cvering (approved):

mwetter:
Thanks @cvering and @Mathadon . Also, @nytschgeusen approved by email. I will go ahead and merge it.

nytschgeusen:
This non-backward compatible change is also ok for the BuildingSystems library.

PR #1598, opened 2022-03-23
Added compiler directives for #1596

### LINKED ISSUE #1596: Avoided multiple inclusion of getTimeSpan.c
When compiling some model with Spawn of EnergyPlus, such as `Buildings.ThermalZones.EnergyPlus.Examples.SingleFamilyHouse.ShadeControl` with OMEdit inside a VirtualBox that runs as guest Windows and uses as host Linux, the compilation can sometimes fail due to repeated inclusion of `getTimeSpan.c`.

This should not happen and I don't know if this is related to latency issues in this setup. Just compiling weather data models did not exhibit this issue, and there is nothing special in this regard for the Spawn model.

Regardless of the root cause, this can easily be avoided without side effect by putting the c file into
```c
#ifndef GETTIMESPAN_C_
#define GETTIMESPAN_C_
...
#endif
```
This issue is to add these compiler directives.

mwetter:
This closes #1596

### DIFF
--- a/IBPSA/Resources/C-Sources/getTimeSpan.c
+++ b/IBPSA/Resources/C-Sources/getTimeSpan.c
@@ -2,6 +2,9 @@
  * getTimeSpan.c
  */
 
+#ifndef GETTIMESPAN_C_
+#define GETTIMESPAN_C_
+
 #include <stdio.h>
 #include "ModelicaUtilities.h"
 
@@ -137,3 +140,5 @@ void getTimeSpan(const char * fileName, const char * tabName, double* timeSpan) 
 
   return;
 }
+
+#endif


MassimoCimmino (commented):
@mwetter I don't have much experience with C (if at all). I do not fully understand the issue and the solution.

I can attest that tests pass and that the modifications do not have negative impacts on the library. I can approve the changes if this is a sufficient review.

mwetter:
@MassimoCimmino : Thanks for the feedback.
@Mathadon : Could you please review/approve.

Mathadon (approved):

Mathadon:
@MassimoCimmino you have no idea what you're missing ;)

PR #1622, opened 2022-06-06
Issue1621 negative fan power

### LINKED ISSUE #1621: Negative Fan Power
This issue is to add a non-negative constraint to the flow work computation
_W&#775;<sub>flo</sub> = V&#775; &Delta; p >= 0_
in `IBPSA.Fluid.Movers.BaseClasses.FlowMachineInterface` in order to prevent the model from producing negative power when either the pressure rise or the flow rate is negative. 
An example of this situation is in the model `Buildings.Examples.DualFanDualDuct.ClosedLoop` (as of https://github.com/lbl-srg/modelica-buildings/commit/b65753005f56311322c675892f0fc75fc5b1d9bd). When the cold deck is dominant, the hot deck fan `fanSupHot` is forced to have a pressure drop instead of rise across the fan because it is overpowered by the return fan `fanRet`. This results in negative fan power in the model which is incorrect.

ClosedLoop

hcasperfu:
This closes #1621.


mwetter (approved):

mwetter:
@Mathadon : Can you please review so we have a non-LBL reviewer.

Mathadon (changes requested):
See https://github.com/ibpsa/modelica-ibpsa/issues/1621#issuecomment-1153776458

PR #1623, opened 2022-06-09
Issue1577 equivalent borehole g functions

### LINKED ISSUE #1577: Equivalent borehole method to evaluate g-functions of geothermal borefields
This is a proposal to replace the current evaluation of g-functions in `IBPSA.Fluid.Geothermal.Borefields.BaseClasses.HeatTransfer.ThermalResponseFactors.gFunction` with our recent "equivalent borehole" approach [1].

This new method is much faster than the current one, and it has the added advantage of significantly lowering the memory requirement for the calculation. The current approach is hard-coded to decrease its resolution starting at 125 boreholes to avoid hitting the memory limit allocated by dymola for the initialization step. The new approach would not require that, even for fields of 1000 boreholes or more.

------------------------------------
[1] Prieto, C., & Cimmino, M. (2021). [Thermal interactions in large irregular fields of geothermal boreholes: the method of equivalent boreholes](https://www.tandfonline.com/doi/full/10.1080/19401493.2021.1968953). *Journal of Building Performance Simulation*, 14(4), 446-460.

MassimoCimmino:
This closes #1577.

I have not generated test results for the new tests or updated results for the modified tests. My dymola is not up-to-date to the newest version.

Some tests are expected to fail since the method for the calculation of g-functions has changed. The differences should not be significant in practical simulations.


IBPSA/Fluid/Geothermal/Borefields/BaseClasses/HeatTransfer/ThermalResponseFactors/finiteLineSource_Integrand_EquivalentBoreholes.mo
```diff
@@ -0,0 +1,65 @@
...
+
+  input Real u(unit="1/m") "Integration variable";
+  input Modelica.Units.SI.Distance dis[n_dis] "Radial distance between borehole axes";
+  input Integer wDis[n_dis];
+  input Modelica.Units.SI.Height len1 "Length of emitting borehole";
+  input Modelica.Units.SI.Height burDep1 "Buried depth of emitting borehole";
+  input Modelica.Units.SI.Height len2 "Length of receiving borehole";
+  input Modelica.Units.SI.Height burDep2 "Buried depth of receiving borehole";
+  input Integer nBor2;
+  input Integer n_dis;
```
mwetter, line 14: Add missing comment

IBPSA/Resources/src/fluid/geothermal/visualize_borehole_clusters.py
```diff
@@ -0,0 +1,40 @@
+from buildingspy.io.outputfile import Reader
```
mwetter, line 1: It's not clear why this file is part of the PR. It does not generate a figure (or at least it does not write it to disk for display in the `info` section). Should this file be removed, or if not, can you add a comment in the file that explains its usage?

mwetter (changes requested):
@MassimoCimmino : Please see the inline comments for minor changes. Afterwards it is ready to merge from my point of view.

MassimoCimmino:
It seems I was able to fix my issue with the open modelica tests. This will be ready for re-review.

mwetter (approved):
This is ready to merge. Functions have also been revised so they work with OPTIMICA.

mwetter (approved):
This is ready to merge. Functions have also been revised so they work with OPTIMICA.

Issue #1531, opened 2021-10-13
Reimplement der_spliceFunction


mwetter:
Hubertus reported that the current implementation of `spliceFunction_der` in the MSL (or `der_spliceFunction` as it is called in the IBPSA library), the `cosh` function is used. But this causes an overflow if its argument is around 800 which can be very hard to diagnose if it happens in a solver. He provided another implementation with `tanh` that does not exhibit this potential problem.

This issue is to implement the new formulation.

PR #1641, opened 2022-09-27
der_spliceFunction check limits

mwetter:
this is a follow-up for #1531 and #1532
there are two if-conditions that used a different value: 1 versus 0.9999999999
this PR makes sure the identical limit is used in
spliceFunction and corresponding derivative

This integrates https://github.com/ibpsa/modelica-ibpsa/pull/1640


thorade (approved):
Thanks for adding revision notes

Issue #1624, opened 2022-06-15
Low ACH in IBPSA.Fluid.Examples.SimpleHouse


hcasperfu:
This is to address a problem of `IBPSA.Fluid.Examples.SimpleHouse` which was exposed by an upcoming changes to the movers model.
The new methods will be able to account for variable efficiency values at different flow rate and pressure rise instead of using a constant efficiency. Because `SimpleHouse` is not properly parameterised, it leads to a temperature rise across the fan of 5 degC when the fan has a low flow and high pressure.

Pasted (and modified) from email communications:
@hcasperfu
Because now the default mover efficiency computation methods are .EulerNumber and .GenericCurve, in general we keep the power at a certain level even when the flow work is very low. This has caused a problem with heat dissipated into the medium. 
In Buildings.Fluid.Examples.SimpleHouse, when the flow is low, because P is kept somewhat high, appreciable amount of heat is added to the medium. 

(1 = [lbl-srg/modelica-buildings/issue2668 @bbb23fece0faf13502f5cf178f64a9c69eecd24d](https://github.com/lbl-srg/modelica-buildings/commit/bbb23fece0faf13502f5cf178f64a9c69eecd24d); 2 = [lbl-srg/modelica-buildings/master@a4bf2006b2f8a7620e90e736e8dc5c2e13e9b21d](https://github.com/lbl-srg/modelica-buildings/commit/a4bf2006b2f8a7620e90e736e8dc5c2e13e9b21d))
<img width="468" alt="image" src="https://user-images.githubusercontent.com/91219266/173931380-44219216-d228-4a56-90ba-fadab7edc842.png">

Comment from @mwetter:
The issue is that the damper is allowed to be y=0.1 which it mostly is. This results in an air change per hour of 0.01 1/h, which is way too low. The setting is in the PI controller, parameter yMin. If this limit is raised to 0.25, then we have a minimum ACH = 0.3 1/h. This is more realistic, and leads to a reasonable operating point efficiency of the fan.

PR #1646, opened 2022-10-19
Issue1645 fan efficiency euler

### LINKED ISSUE #1645: Fan efficiency with Euler number
This issue is to integrate the improved fan/pump model from lbl-srg/modelica-buildings#2668 to the IBPSA library for review.

Key features:
- The Euler number method is introduced to compute the hydraulic efficiency to account for efficiency degradation at part load more easily.
- Generic curves are introduced to compute the motor efficiency for part-load considerations as well.
- The above two methods are now the default for any fan/pump instances.
- A subpackage of preconfigured mover models is created. The user only needs to provide the nominal flow rate and the nominal pressure rise (and the nominal speed for the speed-input model only) and the mover model is parameterised automatically. No more need to construct a dummy pressure curve!

Other changes:
- `Utilities.Math.Functions.smoothInterpolation` (and its example model) is also merged here from the `Buildings` library to support the above changes.

hcasperfu:
This closes #1645.


IBPSA/Fluid/Movers/Preconfigured/UsersGuide.mo
```diff
@@ -0,0 +1,48 @@
...
+are used to construct the pressure curve <i>&Delta; p = f(m&#775;,  y)</i> where <i>m&#775;</i> is the mass flow rate and <i>y</i> is the speed.
+The curve support points come from regression of the pump and fan data records
+available in the data package.
+The pump pressure curve is based on all data records in
+<a href=\"Modelica://IBPSA.Fluid.Movers.Data.Pumps\">
+IBPSA.Fluid.Movers.Data.Pumps</a>
+and the fan pressure curve is based on all data records in
+<a href=\"Modelica://IBPSA.Fluid.Movers.Data.Fans\">
+IBPSA.Fluid.Movers.Data.Fans</a>.
+(See also
```
mwetter, line 28: Can you please be more specific what you mean by "all data records". Is it some average, or a typical curve?

mwetter (changes requested):
@hcasperfu : After the minor comment I added, this is fine for external review.

IBPSA/Fluid/Movers/Preconfigured/UsersGuide.mo
hcasperfu: Hopefully this is clarified in improved documentation in ea34e88.

mwetter: Great, now it is clear.

mwetter:
@Mathadon : Can you please review this PR. This is from our point of view ready to be merged.

Mathadon:
@mwetter @hcasperfu thanks for this great functionality! I'm going to do a proper review but it may take some time and some iterations. I have first reviewed the documentation and main components, and did a first test based on the benchmark data that was already in the repository. Main remarks: 
- The model `IBPSA.Fluid.Movers.Validation.Pump_Nrpm_stratos` was originally designed for verifying the pump accuracy. My understanding is that the model should become more detailed especially for the low speed low head regime where it is currently underpredicting the power use.  However, it seems that the results of this model are unaltered, which does not make sense to me. Looking into the code, it seems that the value of `etaHydMet=IBPSA.Fluid.Movers.BaseClasses.Types.HydraulicEfficiencyMethod.Power_VolumeFlowRate` is avoiding the Euler method from being used. I didn't dig deep into the code yet but that does not make sense to me. Either this option should no longer exist, or it should be used to compute the peak efficiency from the flow curve provided by the user?
- The implementation of `IBPSA.Fluid.Movers.BaseClasses.Euler` is quite expensive. I presume that we can find a good approximation for this curve and allow the user to select between both of them? 
image
- `n=12` in `IBPSA.Fluid.Movers.BaseClasses.Euler` is a lot. What is the rationale for the 10% increments that are used?
- `Modelica.Blocks.Tables.CombiTable2Ds` is used, which requires external C code, which is not the best idea for such a basic model. Were other types of implementations considered? Is my interpretation correct that this table interpolates a mapping of `IBPSA.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve`?

hcasperfu:
@Mathadon Thank you for the quick comments!
- For `IBPSA.Fluid.Movers.Validation.Pump_Nrpm_stratos`, it is intended that this model's results remain unchanged (as well as any other model that directly uses available pump data). There are a few reasons for this:
  - I found that the accuracy of the Euler number method is somewhat sensitive to the guessed peak point whose accuracy is hard to guarantee, especially when done automatically. Therefore I believe the power curve method is still valuable to be kept around.
  - The Wilo pump data are based on total electric power draw `PEle` instead of the hydraulic power `WHyd`. So there is a layer of `etaMot` in between. (Although `etaMot` is mostly constant and doesn't cause too much trouble).
  - Many of the new example & validation models are specifically intended to compare the results from different efficiency methods, especially `IBPSA.Fluid.Movers.BaseClasses.Validation.EulerComparison`. In fact, this model is specifically setup in a way that you can replace the performance data `per` with any of the available records in `Movers.Data` to see the difference without changing anything else.
- It is indeed expensive. That's why two lookup tables with `Modelica.Blocks.Tables.CombiTable2Ds` are created (one for efficiency and one for power), so that the correlation function only needs to be evaluated at initialisation. See `curEu` inside `IBPSA.Fluid.Movers.BaseClasses.FlowMachineInterface`. Do you think further simplification is needed?
- There is no serious rationale behind `n=12`. It is simply because 10% is a convenient and intuitive number. Suggestions would be appreciated.
- As for `Modelica.Blocks.Tables.CombiTable2Ds`:
  - I certainly thought about writing a two-way cubic interpolation as a `function` which would also help avoid a few awkward maneuvers with `Connect()` operators inside `FlowMachineInterface` (which is based on written out equations). I looked at the C code behind the block and found it to be quite a hassle and stopped there. But it might still be worth the effort?
  - This block is only used for the hydraulic power `WHyd` and the hydraulic efficiency `etaHyd` when using the Euler number.
    - The motor efficiency `etaMot` when using the `GenericCurve` method is found by doing one-way interpolations twice. (a) The first time, it uses the given `WMot_nominal` to reduce the table in `IBPSA.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve` to one curve. (b) The second time, it finds `etaMot` with `yMot = WHyd / WMot_nominal` inside `FlowMachineInterface`.
    - This approach of double one-way interpolation is possible for $&eta;\_{mot}=f(W_{mot,nominal},y_{mot})$ because `WMot_nominal` is a constant for any given mover, while $&eta;_{hyd}=f(V&#775;,&Delta;p)$ cannot be reduced.

Mathadon:
@hcasperfu thanks for the quick feedback!

- I'm surprised that it's hard to find the peak efficiency point. As you mention `etaMot` should be pretty constant _around the maximum hydraulic efficiency point_.
- I did not realise that `IBPSA.Fluid.Movers.BaseClasses.Euler` is only used at initialisation. Then it's no problem, nice!
- Thinking about it, I can imagine that the value of `n` does not affect computation time a lot as long as there is a good initial guess for the interpolation point, which the `ModelicaStandardTables` should take care of.
- Possibly the `CombiTimeTable2Ds` is fast enough or even faster than a polynomial.. It would be interesting to benchmark that, actually. The problem is that it's not very portable, e.g. for optimisation. 

Notes to self: 
 - [ ] Verify whether the Euler method is used by default
 - [ ] Verify how the Euler method behaves for the Stratos model. Is the maximum efficiency model accurate?
 - [ ] Check `IBPSA.Fluid.Movers.BaseClasses.Validation.EulerComparison`
 - [ ] Check the shape of the data in `CombiTable2Ds` and see whether it could be approximated using polynomials easily.

I'll look again into it next week. Feedback welcome!

Mathadon:
@hcasperfu @mwetter I'd like to do a proper review of this model and need a few hours for that. Please bare with me until I find some time for that in my agenda. Sorry for the delay!

Mathadon:
Do we need two definitions of `per`?
image

`Movers.Data.Generic` contains:
```
  final parameter IBPSA.Fluid.Movers.BaseClasses.Euler.peak peak_internal=
    IBPSA.Fluid.Movers.BaseClasses.Euler.getPeak(pressure=pressure,power=power)
```

which computes the default peak point based on a power curve. It thus seems that the peak efficiency point is not computed when an efficiency curve is provided instead of a power curve?

The code contains two tables:

```
  Modelica.Blocks.Tables.CombiTable2Ds effTab(
    final table=curEu.eta,
    final smoothness=Modelica.Blocks.Types.Smoothness.ContinuousDerivative)
    if per.etaHydMet==
        IBPSA.Fluid.Movers.BaseClasses.Types.HydraulicEfficiencyMethod.EulerNumber
    "Look-up table for mover efficiency";
  Modelica.Blocks.Tables.CombiTable2Ds powTab(
    final table=curEu.P,
    final smoothness=Modelica.Blocks.Types.Smoothness.ContinuousDerivative)
    if per.etaHydMet==
        IBPSA.Fluid.Movers.BaseClasses.Types.HydraulicEfficiencyMethod.EulerNumber
    "Look-up table for mover power";
```

why don't we use just one table `powTab` and compute the efficiency from `eta=∆p*Q/P`?

A comparison for `IBPSA.Fluid.Movers.Validation.Pump_Nrpm_stratos` using similarity laws vs the Euler number computation:

Power:
image

Efficiency:
image


Differences are relatively small, even for zero flow rate.



@mwetter note that we have negative `eta` when a pump is used as a turbine (when flow work is negative):
image

Mathadon:
I have created an alternative implementation for `IBPSA.Fluid.Movers.BaseClasses.Euler.correlation`:

```
within IBPSA.Fluid.Movers.BaseClasses.Euler;
function approximation
  "Correlation of static efficiency ratio vs log of Euler number ratio"
  extends Modelica.Icons.Function;
  input Real x "log10(Eu/Eu_peak)";
  output Real y "eta/eta_peak";

protected
  Real a = if x <-0.5 then 0.056873227 else if x>0.5 then -0.0085494313567 else 0.378246;
  Real b = if x<-0.5 then 0.493231336746 else if x>0.5 then 0.12957001502 else -0.759885;
  Real c = if x<-0.5 then 1.433531254001 else if x>0.5 then -0.65997315029 else -0.0606145;
  Real d = if x<-0.5 then 1.407887300933 else if x>0.5 then 1.1399300301 else 1.01427;
 algorithm
  y:=max(0,a*x^3+b*x^2+c*x+d)/1.01545;
end approximation;
```
The idea behind this curve is that the two sides of the Euler curve can be approximated well using a simple polynomial:
image

Comparison:
image


Benchmarked using

```
algorithm 
  for i in 1:100000 loop
  //etaRat := IBPSA.Fluid.Movers.BaseClasses.Euler.correlation(x+i/1e6);
  etaRat := IBPSA.Fluid.Movers.BaseClasses.Euler.approximation(x+i/1e6);
  end for;
```

using OM and 500 Euler steps.

`approximation`:    0.125734s [ 98.1%] simulation
`correlation`:      60.4951s [ 99.3%] simulation

So the polynomial seems to be 500x faster.

@mwetter  I propose to implement the polynomial directly instead of the `CombiTable2Ds`. Do you agree?

Mathadon:
Xlsx file:

[Euler analysis.xlsx](https://github.com/ibpsa/modelica-ibpsa/files/10047495/Euler.analysis.xlsx)

mwetter:
I think the changing of avoiding `CombiTable2Ds` is good (faster and no need for external C code). I assume @hcasperfu agrees too (he currently is out of office).

Todo before merging:
- [ ] Remove duplicate record
- [ ] Replace `CombiTable2Ds` with polynomial approximation.
- [ ] ensure `eta` is positive if used as a turbine (negative flow work).

hcasperfu:
@Mathadon 
I am (finally) back to office and will work on your comments!
- One quick question: Can you tell me where you found the duplicate records? I can't find it anywhere.
- I will implement the polynomial approximation. Thank you for developing this!

Mathadon:
@hcasperfu you can see the duplicate records here:




One copy is probably defined in one model and the second copy in a model that it extends from.

hcasperfu:
@Mathadon 
Can you tell me which model you were looking at? I don't see the second `per` in `PartialFlowMachine` (screenshot below) nor in any of the four fan/pump models or the four preconfigured models.

Screenshot 2022-12-10 at 12 45 45

hcasperfu:
@Mathadon 
Greetings! I have implemented your polynomial approximation in the latest push (up to 117b028).
- I used your polynomial approximation to replace the old implementation in `BaseClasses.Euler.correlation`. Because I was out of office for a couple of weeks and lost track of emails & github notifications, I didn't see #1665 until after I already integrated your code from the comments. The actual implementation is therefore a tiny bit different.
- Now `BaseClasses.FlowMachineInterface` handles the Euler path in a very similar way to the path with just power data.
  - The model computes a power curve together with its derivatives wrt flow rate at nominal speed with `BaseClasses.Euler.power`. This power curve is handled in a way that it is prevented from doing anything funny near zero flow or zero pressure (see documentation of the function).
  - `FlowMachineInterface` then uses `BaseClasses.Characteristics.{power,pressure}` to scale the power down at reduced speeds.
    - An unintended effect of this is that the Euler method now is also switched on and off by `computePowerUsingSimilarityLaws` for reduced speeds. But this is perhaps not a big deal. For speed-controlled models, this is always `true`. For flow-controlled models, this is set to `final computePowerUsingSimilarityLaws=per.havePressureCurve`. Because the Euler can't even guess the peak without a pressure curve, turning it off has no effect.
    - This is manifested in the significant result change in `Validation.FlowControlled_dp` (compared to 0b2ce50, where Euler still tried to guess even without a pressure curve provided).
    - The preconfigured models are unaffected because they always have a pressure curve, hence always using Euler for reduced speeds.
  - Please recall that the Euler number method does not account for efficiency degradation along any curve of $&Delta;p = k V&#775;^2$ where $k$ is constant. This means using this method is equivalent to using affinity laws with a known power curve as long as the peak is accurate. This is explained in `usersguide` and the documentation of `BaseClasses.Validation.EulerReducedSpeed`. More detailed discussion is in the conference paper ([Fu et al., 2022](https://2022.american.conference.modelica.org/documents/NA_Modelica_2022_Proceedings.pdf)) (cited in the `usersguide`).
  - Then efficiency is then back-calculated from power (as you suggested).
- The original correlation has been moved to `IBPSA.Fluid.Movers.BaseClasses.Validation.EulerCurve` to be compared with the approximation.

Other items:
> Do we need two definitions of `per`?

I don't see the second record in Dymola nor OMEdit on my machine.

> It thus seems that the peak efficiency point is not computed when an efficiency curve is provided instead of a power curve?

That is correct.
The idea is that if the user already knows an efficiency curve then they don't need the Euler number method.
Also, because Euler is equivalent to using affinity curves with a known power curve (discussed above), its value is giving efficiency estimation ALONG the pressure curve at the nominal speed. It cannot give an accurate estimation on a curve at reduced speed. The `getPeak()` function is therefore more intended to be used for our validation against available data rather than a general purpose.

> note that we have negative eta when a pump is used as a turbine (when flow work is negative)

I can no longer reproduce this behaviour in the latest commit (117b028). Efficiency is zero for the forced pumps.
We already had a discussion on the turbine issue in #1621. I concur with Michael's comment in that post. I don't think our average user would expect the mover model to behave like a turbine. (Therefore I rather think it is important to ensure the fan/pump model actually does NOT _accidentally_ generate power. ) To model that accurately and efficiently also requires more work on this model. 
And personally, because this is already a fat 500-commit PR on Buildings, I would _really_ prefer to put any new content on a separate issue.

hcasperfu:
There are three models outside of `Movers` package that have significant change in fluid temperature, which is a result of decreased mover efficiency from the new method (set as default). They are:
- `IBPSA.Fluid.Examples.ResistanceVolumeFlowReversal` (18 K)
Because the nominal pressure rise of the pump is four times its pressure at the normal condition, the hydraulic efficiency is as low as 0.2 from the Euler method. It resulted in a 18 K increase in fluid temperature at the end of the simulation.
image

- `IBPSA.Fluid.Examples.SimpleHouse` (< 1 K)
This one's fan was already looked at in #1624, but this time it's about the pump and the change in temperature is only about 0.3 K most of the time. The pump's nominal pressure is twice its normal operating pressure.
image

- `IBPSA.Fluid.HeatPumps.Examples.ScrollWaterToWater_OneRoomRadiator` (4 K at some time steps)
The temperature had a 4 K increase at some time steps. Both pumps have nominal pressure four times the their usual pressure.
image

- [ ] I will look at these models in more detail next week to see if they can be fixed by simply changing the nominal value.

It is also odd that this wasn't discovered earlier (say, @ 0b2ce50). Upon a brief investigation I believe the old implementation had a bug where nominal values weren't properly passed to the Euler method, and the method would provide an estimated efficiency of either 0.7 or 1E-5 (the small value in place of zero). But this is perhaps not so important now since the old method has been scrapped.

hcasperfu:
Update as of 1326679: I have been debugging with optimica on Buildings side. There are still things that need to be fixed so some other tweaks may be expected. 
Happy holidays!

Mathadon:
Okay, sounds good! Happy holidays :)

hcasperfu:
Update as of 4500fc3: The code now passes all CI tests on the Buildings side. I made some "decorations" in `IBPSA.Fluid.Movers.BaseClasses.Euler.power` to suppress optimica errors. Now the code is ready for review again.

IBPSA/Fluid/Movers/BaseClasses/Euler/getPeak.mo
```diff
@@ -0,0 +1,225 @@
+within IBPSA.Fluid.Movers.BaseClasses.Euler;
+function getPeak
```
Mathadon, line 2: nice work :)

IBPSA/Fluid/Movers/BaseClasses/Euler/package.mo
```diff
@@ -0,0 +1,56 @@
...
+package Euler "Functions and data record templates for Euler number"
+
+  annotation (Documentation(info="<html>
+<p>
+This package implements a power computation path using the Euler number
+and its correlation.
+</p>
+<ul>
+<li>
+Efficiency and power are computed and output as look-up tables by
```
Mathadon, line 11: is this still accurate?

IBPSA/Fluid/Movers/FlowControlled_m_flow.mo
```diff
@@ -20,7 +20,22 @@ model FlowControlled_m_flow
           IBPSA.Fluid.Movers.BaseClasses.Characteristics.flowParameters(
             V_flow = {i/(nOri-1)*2.0*m_flow_nominal/rho_default for i in 0:(nOri-1)},
             dp =     {i/(nOri-1)*2.0*dp_nominal for i in (nOri-1):-1:0}),
-      final use_powerCharacteristic = if per.havePressureCurve then per.use_powerCharacteristic else false),
+        final etaHydMet=
```
Mathadon, line 23: These changes look like a duplicate of the changes in `IBPSA/Fluid/Movers/FlowControlled_dp.mo`. Is it possible to avoid this using the existing partials?

IBPSA/Fluid/Movers/Preconfigured/FlowControlled_dp.mo
```diff
@@ -0,0 +1,37 @@
+within IBPSA.Fluid.Movers.Preconfigured;
```
Mathadon, line 1: handy!

Mathadon (commented):
@hcasperfu @mwetter I have reviewed the code and made a few minor comments.  Thanks for this pull request with many examples and extensive documentation. :) I think we're nearly finished here. I have reviewed all code diffs but still want to have a resulting models and some examples later:

- IBPSA/Fluid/Movers/BaseClasses/Validation/EulerComparison.mo
- IBPSA/Fluid/Movers/BaseClasses/Validation/EulerReducedSpeed.mo
- IBPSA/Fluid/Movers/BaseClasses/Validation/HydraulicEfficiencyMethods.mo

So please bear with me a little longer until I can finish this review :)

IBPSA/Fluid/Movers/BaseClasses/Euler/package.mo
hcasperfu: This is corrected in e885703.

IBPSA/Fluid/Movers/FlowControlled_m_flow.mo
hcasperfu: This is something I have thought about. There is a lot of code shared by the two flow-controlled models (`FlowControlled_dp` and `FlowControlled_m_flow`) and the two speed-controlled models (`SpeedControlled_Nrpm` and `SpeedControlled_y`). All four models extend `BaseClasses.PartialFlowMachine`.
These pairs share top-level declarations. These top-level declarations are awkward to be moved down to `BaseClasses.PartialFlowMachine` because they are not used by the other pair.
I would suggest creating a new layer of partial model in between them, say `BaseClasses.PartialFlowControlled` and `BaseClasses.PartialSpeedControlled`. This should be helpful tidying up all the copy-pastas. The downside is that this makes the `Movers` package even more convoluted and difficult to read and maintain. (But as a former C# user, I have always thought we haven't abused the inheritance nearly enough in Buildings and IBPSA libraries.)
I can work on this if others agree this is helpful.

Mathadon:
@hcasperfu ok for me to introduce an extra layer!

hcasperfu:
@mwetter 
I meant to ask for your thoughts on these proposed changes in the cancelled meeting today.

> There is a lot of code shared by the two flow-controlled models (`FlowControlled_dp` and `FlowControlled_m_flow`) and the two speed-controlled models (`SpeedControlled_Nrpm` and `SpeedControlled_y`). All four models extend `BaseClasses.PartialFlowMachine`.
These pairs share top-level declarations. These top-level declarations are awkward to be moved down to `BaseClasses.PartialFlowMachine` because they are not used by the other pair.
I would suggest creating a new layer of partial model in between them, say `BaseClasses.PartialFlowControlled` and `BaseClasses.PartialSpeedControlled`. This should be helpful tidying up all the copy-pastas. The downside is that this makes the Movers package even more convoluted and difficult to read and maintain. (But as a former C# user, I have always thought we haven't abused the inheritance nearly enough in Buildings and IBPSA libraries.)

@Mathadon and I agree that it would be helpful introducing a new layer in the inheritance structure.

mwetter:
@hcasperfu : I am inclined to not introduce a new layer because there is not that much code duplication in `FlowControlled_dp` and `FlowControlled_m_flow`, other than some of the declarations in the `extends` statement. 

I do however think that `SpeedControlled_Nrpm` could be moved to `IBPSA.Obsolete.Movers` (which we could do after this refactor so we don't overload changes in one PR). This would eliminate some duplication, and I never saw a user using the model that takes Nrpm as an input (and there is not much point if we normalize all other control signals to 0...1).

hcasperfu:
@mwetter 
Thank you for your input!
I agree.
Most of the code duplication is between the two speed-controlled models. Merging them eliminates 33 lines of copied code. However, merging the two flow-controlled models only eliminates 13. Once `_Nrpm` is out of the picture, the value of the new inheritance layer is limited.

Mathadon:
okay, fine by me :)

Mathadon:
I can't get around to reviewing this in further detail within the coming month. I propose to move ahead with merging this and I can still make new pull requests later on if things pop up :)

Mathadon (approved):

IBPSA/Fluid/Movers/BaseClasses/Euler/getPeak.mo
```diff
@@ -0,0 +1,225 @@
...
+        ensureMonotonicity=false);
+      peak.dp:=IBPSA.Utilities.Math.Functions.smoothInterpolation(
+        x=peak.V_flow,
+        xSup=V_flow,
+        ySup=dp,
+        ensureMonotonicity=false);
+    end if;
+  end if;
+
+  annotation(smoothOrder=1,
```
mwetter, line 160: The annotation `smoothOrder=1` must be removed as this function is not C1.

IBPSA/Fluid/Movers/BaseClasses/FlowMachineInterface.mo
```diff
@@ -321,31 +416,40 @@ the simulation stops.");
...
+         the motor's rated power, but per.WMot_nominal is not assigned or
+         cannot be estimated because no power curve is provided.");
+
+  assert(max(per.power.P)<1E-6 or per.WMot_nominal>max(per.power.P)*0.99,
+         "In " + getInstanceName() + ": The rated motor power provided in
+         per.WMot_nominal is smaller than the maximum power provided in per.power.
+         Use a larger value for per.WMot_nominal or leave it blank to allow the
+         model to assume a default value.");
+
+initial algorithm
```
mwetter, line 435: Is it indeed required to put this assertion in an `initial algorithm` section as opposed to simply have it in `initial equation`?

IBPSA/Fluid/Movers/Preconfigured/FlowControlled_dp.mo
```diff
@@ -0,0 +1,39 @@
...
+</html>", revisions="<html>
+<ul>
+<li>
+August 17, 2022, by Hongxiang Fu:<br/>
+First implementation. This is for
+<a href=\"https://github.com/lbl-srg/modelica-buildings/issues/2668\">#2668</a>.
+</li>
+</ul>
+</html>"), Icon(graphics={Line(
+          points={{-100,60},{-80,40},{-40,80}},
```
mwetter, line 36: I suggest to delete this check mark as it makes then schematic diagram busy (and other pre-configured records don't have it either).

mwetter (changes requested):
@hcasperfu : This is almost ready to be merged, I only have some minor changes. Note the comments regarding `smoothOrder`. This is important to state correctly as the solver uses it.

IBPSA/Fluid/Movers/BaseClasses/Euler/correlation.mo
```diff
@@ -0,0 +1,116 @@
...
+
+protected
+  Real a = if x <-0.5 then 0.056873227 else if x>0.5 then -0.0085494313567 else 0.378246;
+  Real b = if x<-0.5 then 0.493231336746 else if x>0.5 then 0.12957001502 else -0.759885;
+  Real c = if x<-0.5 then 1.433531254001 else if x>0.5 then -0.65997315029 else -0.0606145;
+  Real d = if x<-0.5 then 1.407887300933 else if x>0.5 then 1.1399300301 else 1.01427;
+algorithm
+  y:=max(0,a*x^3+b*x^2+c*x+d)/1.01545;
+
+  annotation(smoothOrder = 1,
```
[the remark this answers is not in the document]
hcasperfu: This function was designed by @Mathadon, so he would be the best person to ask whether he put special care into these.
But based on some simple computation from my side, the values of dy/dx at the two connecting points are: 0.98295**48** vs 0.98295**50** @ x=-0.5 and -0.5**37** vs -0.5**43** @ x=+0.5.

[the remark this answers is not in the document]
mwetter: @hcasperfu : Thanks for checking.
@Mathadon : Are these rounding errors -- which are large at x=+0.5 -- or is the implementation not differentiable, which would need to be corrected.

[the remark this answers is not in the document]
hcasperfu: @mwetter 
Apologies! I mistyped the numbers. dy/dx @ x=+0.5 is -0.536815**0** vs -0.536815**2**.

IBPSA/Fluid/Movers/BaseClasses/Euler/efficiency.mo
```diff
@@ -0,0 +1,40 @@
...
+  Real log_r_Eu "Log10 of Eu/Eu_peak";
+  constant Real small = 1E-5 "A small value used in place of zero";
+
+algorithm
+  log_r_Eu:= log10(max(dp * peak.V_flow^2,small)
+                  /max(peak.dp * V_flow^2,small));
+  eta:= peak.eta*
+          IBPSA.Fluid.Movers.BaseClasses.Euler.correlation(x=log_r_Eu);
+
+          annotation(smoothOrder=1,
```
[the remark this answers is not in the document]
hcasperfu: Corrected in 7e4574a.

IBPSA/Fluid/Movers/BaseClasses/Euler/getPeak.mo
hcasperfu: Corrected in 02bdb25.

IBPSA/Fluid/Movers/BaseClasses/FlowMachineInterface.mo
hcasperfu: This particular assert warning no longer causes trouble with Dymola 2023x. This is now moved back to `initial equation` in e889c8e.
However, the other assert warnings in `PartialFlowMachine` still cause the simulation to stop incorrectly if they were moved from `initial algorithm` back to `initial equation`.

IBPSA/Fluid/Movers/Preconfigured/FlowControlled_dp.mo
hcasperfu: Do we somehow differentiate the preconfigured models and the vanilla models at all? Their class names are the same and can be confusing.
Some models on Buildings side (e.g. `Buildings.Examples.ScalableBenchmarks.BuildingVAV.Examples.OneFloor_OneZone`) use both the two versions. (This is because only the fan/pump models using a dummy two-point pressure curve were swapped with the preconfigured version.)

IBPSA/Fluid/Movers/BaseClasses/Euler/correlation.mo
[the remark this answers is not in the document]
mwetter: Thanks for the correction. We should still have higher precision. The current values differ at 1E-6, and solvers sometimes need to solve for higher precision, in some cases, if there are nested loops, this can as well be 1E-10. At that point, this looks like a jump that can throw off algorithms that require continuity of the derivative.

Mathadon:
@mwetter is right, the precision was not high enough.

See:

[Euler analysis2.xlsx](https://github.com/ibpsa/modelica-ibpsa/files/10544547/Euler.analysis2.xlsx)


https://www.wolframalpha.com/input?i=systems+of+equations+calculator&assumption=%7B%22F%22%2C+%22SolveSystemOf4EquationsCalculator%22%2C+%22equation1%22%7D+-%3E%22-0.125+a+%2B+0.25+b+-+0.5c+%2Bd+%3D+8.07320354734741E-01%22&assumption=%22FSelect%22+-%3E+%7B%7B%22SolveSystemOf4EquationsCalculator%22%7D%7D&assumption=%7B%22F%22%2C+%22SolveSystemOf4EquationsCalculator%22%2C+%22equation2%22%7D+-%3E%220.75+a+-b+%2Bc+%3D+9.82954837560552E-01%22&assumption=%7B%22F%22%2C+%22SolveSystemOf4EquationsCalculator%22%2C+%22equation3%22%7D+-%3E%220.125+a+%2B+0.25+b+%2B+0.5c+%2Bd+%3D+8.41267279821246E-01%22&assumption=%7B%22F%22%2C+%22SolveSystemOf4EquationsCalculator%22%2C+%22equation4%22%7D+-%3E%220.75+a+%2Bb+%2Bc+%3D+-5.36815208786659E-01%22

```python
vals=[-0.5+1e-15, -0.5-1e-15, 0.5+1e-15, 0.5 - 1e-15]

for x in vals:
        a = 0.05687322707407 if x <-0.5 else  (-8.5494313567465000E-3  if x>0.5 else 0.37824577860088)
        b = 0.493231336746 if x<-0.5 else (1.2957001502368300E-1  if x>0.5 else -0.75988502317361)
        c = 1.433531254001 if x<-0.5 else (-6.5997315029278200E-1 if x>0.5 else -0.060614519563716)
        d = 1.407887300933 if x<-0.5 else (1.13993003013131 if x>0.5 else 1.01426507307139)

        y=max(0,a*x**3+b*x**2+c*x+d)/1.01545

        print(y)
```

13 digits of precision:

```
0.7950370325813546
0.7950370325813584
0.8284674576013058
0.8284674576013
```

IBPSA/Fluid/Movers/BaseClasses/Euler/correlation.mo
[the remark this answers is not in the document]
hcasperfu: I have used higher-precision coefficients provided by @Mathadon for the approximation polynomials in dcd7485.

IBPSA/Fluid/Movers/BaseClasses/Euler/efficiency.mo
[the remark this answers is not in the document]
hcasperfu: This is further corrected in 933c3cd.

IBPSA/Fluid/Movers/Preconfigured/FlowControlled_dp.mo
hcasperfu: I deleted the check mark in b7fba88.
I tried to think of some other way to do the preconfigured model icon, but feel probably not worth the effort.

hcasperfu:
@mwetter 
I have completed the revisions according to our morning meeting.
In addition to enhancing the regularisation in `BaseClasses.Euler.efficiency()`, I applied the same principle to correct the regularisation in `BaseClasses.FlowMachineInterface`. The power variables in the latter class now also depend on the magnitude of the flow & pressure, instead of being regularised at a constant scale (d777888).

IBPSA/Fluid/Movers/BaseClasses/Euler/efficiency.mo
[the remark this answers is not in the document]
hcasperfu: This is twice further corrected in 8102eb5.

IBPSA/Fluid/Movers/BaseClasses/Euler/correlation.mo
```diff
@@ -0,0 +1,120 @@
...
+  input Real x "log10(Eu/Eu_peak)";
+  output Real y "eta/eta_peak";
+
+protected
+  Real a = if x <-0.5 then 0.05687322707407 else if x>0.5 then -8.5494313567465000E-3 else 0.37824577860088;
+  Real b = if x<-0.5 then 0.493231336746 else if x>0.5 then 1.2957001502368300E-1 else -0.75988502317361;
+  Real c = if x<-0.5 then 1.433531254001 else if x>0.5 then -6.5997315029278200E-1 else -0.060614519563716;
+  Real d = if x<-0.5 then 1.407887300933 else if x>0.5 then 1.13993003013131 else 1.01426507307139;
+algorithm
+  y:=max(0,a*x^3+b*x^2+c*x+d)/1.01545;
```
mwetter, line 14: @mwetter to add regularization with smoothMax.

mwetter (changes requested):

PR #1649, opened 2022-10-22
Updated version number to 4.0.0

### LINKED ISSUE #1648: Version inconsistency in conversion script
The conversion script is used to convert from 3.0 to 4.0.
However, the current version is 3.0.

```annotation (
version="3.0.0",
versionDate="2018-09-28",
dateModified = "2018-09-28",
uses(Modelica(version="4.0.0")),
conversion(from(version="3.0.0",
                script="modelica://IBPSA/Resources/Scripts/Conversion/ConvertIBPSA_from_3.0_to_4.0.mos")),annotation (
version="3.0.0",
versionDate="2018-09-28",
dateModified = "2018-09-28",
uses(Modelica(version="4.0.0")),
conversion(from(version="3.0.0",
                script="modelica://IBPSA/Resources/Scripts/Conversion/ConvertIBPSA_from_3.0_to_4.0.mos")),
```

Either the correct IBPSA version is 4.0.0, or the script should be named `ConvertIBPSA_from_2.0_to_3.0.mos` (o.s.).

With the structure like this, a weird conversion behavior is triggered in Dymola:

image

Conversion from 3.0 to 3.0 makes no sense.

mwetter:
This closes #1648

### DIFF
--- a/IBPSA/package.mo
+++ b/IBPSA/package.mo
@@ -130,6 +130,15 @@ Non-SI units are to be kept at an absolute minimum, and they must be declared as
   package ReleaseNotes "Release notes"
     extends Modelica.Icons.ReleaseNotes;
 
+  class Version_4_0_0 "Version 4.0.0"
+    extends Modelica.Icons.ReleaseNotes;
+      annotation (preferredView="info", Documentation(info="<html>
+<p>
+Fourth release of the IBPSA library, currently in development status.
+</p>
+</html>"));
+  end Version_4_0_0;
+
   class Version_3_0_0 "Version 3.0.0"
     extends Modelica.Icons.ReleaseNotes;
       annotation (preferredView="info", Documentation(info="<html>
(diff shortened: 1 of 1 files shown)


IBPSA/package.mo
```diff
@@ -165,6 +174,9 @@ This section summarizes the releases of the <code>IBPSA</code> library.
 </p>
 <ul>
 <li>
+<a href=\"modelica://IBPSA.UsersGuide.ReleaseNotes.Version_4_0_0\">Version 4.0.0</a>(xxx)
```
FWuellhorst, line 177: Why not set a date?

IBPSA/package.mo
```diff
@@ -130,6 +130,15 @@ Non-SI units are to be kept at an absolute minimum, and they must be declared as
...
 
+  class Version_4_0_0 "Version 4.0.0"
+    extends Modelica.Icons.ReleaseNotes;
+      annotation (preferredView="info", Documentation(info="<html>
+<p>
+Fourth release of the IBPSA library, currently in development status.
+</p>
+</html>"));
+  end Version_4_0_0;
+
```
FWuellhorst, line 141: It's nice to use these release notes. However, the current ones don't hold any information. Is that intended?

FWuellhorst (commented):
@mwetter , please check my comment, especially regarding the date. If this is wanted, then I will approve.

IBPSA/package.mo
mwetter: Because the date will be set when there is an official release that then will be tagged.

IBPSA/package.mo
mwetter: As the IBPSA Library is usually not used directly (all code is in the four derivative libraries), we do not invest the time to enter the release notes. It often is already hard enough to have contributors write good revision notes in the models, which is more important to keep track of why certain changes were done.

FWuellhorst (approved):

PR #1653, opened 2022-11-02
Improved documentation and implementation of regNonZeroPower

### LINKED ISSUE #1652: regNonZeroPower: Undesired behaviour for 2 <= n <= 4
This issue is to correct the regNonZeroPower function based on the report at
https://github.com/lbl-srg/modelica-buildings/issues/3135

mwetter:
This closes #1652

### DIFF
--- a/IBPSA/Utilities/Math/Functions/regNonZeroPower.mo
+++ b/IBPSA/Utilities/Math/Functions/regNonZeroPower.mo
@@ -4,7 +4,9 @@ function regNonZeroPower
   extends Modelica.Icons.Function;
 
  input Real x "Abscissa value";
- input Real n "Exponent";
+ input Real n(
+   min=0.000001,
+   max=1.999999) "Exponent";
  input Real delta = 0.01 "Abscissa value where transition occurs";
  output Real y "Function value";
 protected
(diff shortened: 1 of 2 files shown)


IBPSA/Utilities/Math/Functions/regNonZeroPower.mo
```diff
@@ -4,7 +4,9 @@ function regNonZeroPower
   extends Modelica.Icons.Function;
 
  input Real x "Abscissa value";
- input Real n "Exponent";
+ input Real n(
+   min=0.000001,
```
LauraMaier, line 8: Would Modelica.Constants.eps reach a similar goal like this small number?

LauraMaier (approved):
I am not sure if you should replace the small number by a more standardized Modelica.Constants.eps. Other than that, looks good to me - thanks for the nice work!

IBPSA/Utilities/Math/Functions/regNonZeroPower.mo
mwetter: I first had Modelica.Constants.eps, but this may be too small so I hard coded a small number.

PR #1656, opened 2022-11-12
Issue1654 new radiation model tilted surface

### LINKED ISSUE #1654: Add block to calculate global radiation on tilted surface
For models like photovoltaic models, the global radiation on the tilted surface is needed.
So far, IBPSA library only contains the following model: IBPSA.BoundaryConditions.SolarIrradiation.DirectTiltedSurface 
This model calculates only the direct radiation on the tilted surface.
We want to add a model that calculates the global radiation on the tilted surface based on the global horizontal radiation (HGloHor) which is also a typical output of the TMY3 reader and a common measurement in general.

The model will base on works from:

- Erbs et al.: https://doi.org/10.1016/0038-092X(82)90302-4
- Liu et al.: https://doi.org/10.1016/0038-092X(63)90006-9
- Klucher et al.: https://doi.org/10.1016/0038-092X(79)90110-5

LauraMaier:
This closes #1654 

A block was added to calculate the global irradiation on the tilted surface following the approach of Perez.


mwetter (approved):
This is ready to merge.
I renamed the `add` instance for clarity, and also renamed the block from
`GlobalTiltedSurfacePerez` to `GlobalPerezTiltedSurface` so that first all terms with radiation are listed, and then the tilted surface.

PR #1663, opened 2022-11-16
Assigned parameter riseTime in valve instances

### LINKED ISSUE #1662: Actuators.BaseClasses.PartialThreeWayValve does not assign riseTime
The model `Actuators.BaseClasses.PartialThreeWayValve` does not assign the `riseTime` parameter in the valves. While the value is not used in these valves (because the filter is implemented in the main class and disabled in the two valve instances), it leads to confusing results in the result browser that makes debugging harder. 

This issue is to assign the value so that the results are consistent.

mwetter:
This closes #1662


cvering (approved):
Hey @mwetter ! Thanks for updating this issue.

PR #1664, opened 2022-11-17
Initialized variable Done

mwetter:
This initializes a variable that is accessed before initialization.
See https://github.com/OpenModelica/OpenModelica/issues/9707#issuecomment-1317631281


MassimoCimmino (approved):
Everything looks great. Nice find.

PR #1665, opened 2022-11-19
Issue1645 fan efficiency euler patch

Mathadon:
@mwetter can this be merged? See https://github.com/ibpsa/modelica-ibpsa/pull/1646#issuecomment-1320920539

### DIFF
--- a/IBPSA/Fluid/Movers/BaseClasses/Euler/approximation.mo
+++ b/IBPSA/Fluid/Movers/BaseClasses/Euler/approximation.mo
@@ -1,0 +1,62 @@
+within IBPSA.Fluid.Movers.BaseClasses.Euler;
+function approximation
+  "Correlation of static efficiency ratio vs log of Euler number ratio"
+  extends Modelica.Icons.Function;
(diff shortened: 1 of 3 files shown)


Mathadon:
@mwetter see also [327621a](https://github.com/ibpsa/modelica-ibpsa/pull/1665/commits/327621ab13bf2513e33329bb69a8068bda249d23) 
shouldn't this have been caught by tidylib?

mwetter:
@Mathadon Please go ahead and merge it, but I assume you then will actually used the 500x faster approximation in the model?
Note that `smoothOrder` should be added, and 
```a*x^3+b*x^2+c*x+d``` is likely faster as
```d+x*(c+x*(b+x*a))``` (at least in C).

Regarding tidylib, I would think this should be caught, but I think it is not 100% foolproof.

Mathadon:
@hcasperfu @mwetter Since I'm not very familiar with the new code, are you able to help out with the integration of the new function? Otherwise no problem and I'll have a go at it!

mwetter:
@Mathadon : @hcasperfu should be able to help, but he is currently out of the office.

PR #1683, opened 2023-01-06
Issue1682 heat exchanger model numerics

### LINKED ISSUE #1682: Heat Exchanger Model Numerics
In `IBPSA.Fluid.HeatExchangers.{WetCoilEffectivenessNTU,BaseClasses.PartialEffectivenessNTU}`, there is this equation block:

```
equation
  if (configuration == con.ParallelFlow) then
    flowRegime = if (C1_flow*C2_flow >= 0)
      then 
        flo.ParallelFlow
      else 
        flo.CounterFlow;
  elseif (configuration == con.CounterFlow) then
    flowRegime = if (C1_flow*C2_flow >= 0)
      then 
        flo.CounterFlow
      else 
        flo.ParallelFlow;
  elseif (configuration == con.CrossFlowUnmixed) then
    flowRegime = flo.CrossFlowUnmixed;
  elseif (configuration == con.CrossFlowStream1MixedStream2Unmixed) then
    flowRegime = if (C1_flow < C2_flow)
      then 
        flo.CrossFlowCMinMixedCMaxUnmixed
      else 
        flo.CrossFlowCMinUnmixedCMaxMixed;
  else
    flowRegime = if (C1_flow < C2_flow)
      then 
        flo.CrossFlowCMinUnmixedCMaxMixed
      else 
        flo.CrossFlowCMinMixedCMaxUnmixed;
  end if;
```

This block is numerically problematic. For example, with
```
    flowRegime = if (C1_flow*C2_flow >= 0)
      then 
        flo.CounterFlow
      else 
        flo.ParallelFlow;
```
the numerical noise can cause the model to generate events and switch equations frequently.

After brief discussions with @mwetter, we suggest:
- removing this equation block,
- removing the variable `flowRegime`,
- and replacing the use of `flowRegime` elsewhere in these models with `flowRegime_nominal`.

These changes have the implication that once the flow regime is determined by nominal values, the model now sticks with this set of equations regardless of the actual flow conditions during the simulation. But we believe that reversing the flow inside a heat exchanger during operation is unusual. Having a robust implementation is worth having this limitation.

hcasperfu:
This closes #1682.


Mathadon (changes requested):
@hcasperfu @mwetter I would suggest to add an assert that restricts the operation of the model to regimes where the equations are valid. E.g., for a heat exchanger that has parallel flow in nominal conditions, check whether the flow reverses using an assert. The assert could be disabled using a parameter if needed.

mwetter:
@hcasperfu @Mathadon : Such as assert could trigger a warning *once* when either of the mass flow rate becomes `m_flow < -0.1 * m_flow_nominal` (e.g., don't put it at `0` so it does not trigger if there is noise). I wouldn't allow disabling the assertion as this is a regime where the model is no longer applicable after this change. We could not think about a good use case where one needs to operate the model in a system where the flow reverses.

justnielsen:
Sorry for my interferrence. I have two applications where (MBL) heat exchanger flow reverseal is needed — although not for building simulations but for district heating generation:

1. Seawater/DH water plate heat exchangers used for DH return water cooling are back-flushed to remove suspended particles that accumulates in the (plate) heat exchanger
2. Said heat exchangers which are "heat treated" with warm water to mitigate fouling from biological activity by flushing (both directions).

In my applications I simulate the detailed control sequences + the process. In neither case I need the "correct" heat transfer during flow reversal — I just need the option to _allow_ flow reversal.

I would be very happy if you would consider this, e.g. by allowing the assertion to be disabled with a parameter.

mwetter:
@justnielsen : Thanks for this use case.
@hcasperfu : In view of the above use case, I think best would be to use a `constant` to allow advanced users to disable the assertion. This will allow creating models as needed by @justnielsen and not always have a warning which is irrelevant for that use case. Using a `constant` is preferable over a `parameter` as most users shouldn't be bothered with this option.

justnielsen:
@hcasperfu and @mwetter thanks very much for considering my use case. Your work is highly appreciated!

mwetter:
@hcasperfu : Today I came across a use case where reverse flow of such a heat exchanger is needed: For an aquifer thermal energy storage (ATES), the coupling between geothermal loop and district energy system is via a heat exchanger. Such heat exchangers can be operated by switching the flow direction on the ATES side depending on whether heat is rejected or drawn from the ATES.
Therefore, we should change the model so that the `flowRegime` is changed. This can robustly be done using a hysteresis in the if-then statements. The model should change the `flowRegime` if `m_flow < -0.01*m_flow_nominal` and switch back if `m_flow > +0.01*m_flow_nominal`. This way the event is far enough away from 0 so it won't be triggered if there is small backward flow (such as in an air handler unit) but the model is still accurate. The 0.01 should be a parameter in the `Advanced` tab.

Mathadon:
@mwetter the proposed solution introduces a discontinuity in the heat flow rate. 

I'm thinking again now about the root cause of the problem:

```
flowRegime = if (C1_flow*C2_flow >= 0)
      then 
        flo.CounterFlow
      else 
        flo.ParallelFlow;
```
apparently causes events. Around zero flow, both regimes should be continuous (e.g. each have Q_flow=0) and as such the model is continuous and no event should be generated at all.

Can't we just do:

```
flowRegime = smooth(if (C1_flow*C2_flow >= 0)
      then 
        flo.CounterFlow
      else 
        flo.ParallelFlow,1);
```

or 

```
flowRegime = noEvent(if (C1_flow*C2_flow >= 0)
      then 
        flo.CounterFlow
      else 
        flo.ParallelFlow);
```
or something like that to solve the problem?

mwetter:
@Mathadon `smooth` and `noEvent` cannot be used here as `flowRegime` is an integer, and you are right about the discontinuity, hence https://github.com/ibpsa/modelica-ibpsa/pull/1683#issuecomment-1378750031 is not good.

I therefore suggest
1. put an assertion at `m_flow < -0.1 * m_flow_nominal` as described in (https://github.com/ibpsa/modelica-ibpsa/pull/1683#issuecomment-1374392539). Enable it by default, and write it once. Hence by default, `flowRegime = flowRegime_nominal`. This covers more than 99% of the cases, it avoids the overhead of regularization which would need to be done in `epsilonNTU_Z` (which is a costly function), and it handles small reverse flows due to numerical noise robustly.
2. In addition, add a constant such as `use_dynamicFlowRegime = false`. If it is `false`, the model behaves as in 1), if it is `true`, the model should behave as it does now, e.g., update the `flowRegime` based on the product of the flow rates at `0`. The assertion of 1) should inform the user that for the model to be valid at negative flows, set use_dynamicFlowRegime = true. This allows use of the model for ATES.

hcasperfu:
I am working on the implementation. I have two questions.
1. The code uses the capacity flow `C1_flow` and `C2_flow` to determine the flow regime. Do we use the mass flow rate instead to write the assertions?
2. Also, do we simply write the assertions based on the flow itself crossing zero, instead of their product (e.g. basing on `C1_flow * C2_flow` crossing zero)?

mwetter:
@hcasperfu : I suggest to use the mass flow rate rather than the capacity flow rate to determine the flow regime, and write the assertions based on the flows crossing the thresholds.

mwetter (approved):

mwetter:
This is ready to merge if @Mathadon approves.

Mathadon (approved):
@mwetter @hcasperfu thanks for the revisions!

PR #1689, opened 2023-02-03
Corrected graphical annotation

mwetter:
The graphical primitive Line has no attribute lineThickness. It is instead call thickness


mwetter:
@Mathadon : Can you please review this so we can go ahead and merge it. Thanks.

Mathadon (approved):

PR #1691, opened 2023-02-03
added noEvent()

### LINKED ISSUE #1690: OCT simulation fails to run due to the condition check in assert function
In class `IBPSA.Fluid.HeatExchangers.WetCoilEffectivenessNTU` and `IBPSA.Fluid.HeatExchangers.BaseClasses.PartialEffectivenessNTU`, following `assert()` function causes Optimica failed to run.

```
    assert((m1_flow > -0.1 * m1_flow_nominal)
       and (m2_flow > -0.1 * m2_flow_nominal),
"*** Warning in " + getInstanceName() +
      ": The flow direction reversed.
      However, because the constant use_dynamicFlowRegime is set to false,
      the model does not change equations based on the actual flow regime.
      To switch equations based on the actual flow regime during the simulation,
      set the constant use_dynamicFlowRegime=true.
      Note that this can lead to slow simulation because of events.",
      level = AssertionLevel.warning);
```

The error could be like:

```
Traceback (most recent call last):
  File "/home/jianjunhu/proj/06_runOCT/jmodelica.py", line 90, in <module>
    res = mod.simulate(options=opts)
  File "src/pyfmi/fmi.pyx", line 8125, in pyfmi.fmi.FMUModelME2.simulate
  File "src/pyfmi/fmi.pyx", line 378, in pyfmi.fmi.ModelBase._exec_simulate_algorithm
  File "src/pyfmi/fmi.pyx", line 372, in pyfmi.fmi.ModelBase._exec_simulate_algorithm
  File "install/Python/pyfmi/fmi_algorithm_drivers.py", line 348, in __init__
  File "src/pyfmi/fmi.pyx", line 7862, in pyfmi.fmi.FMUModelME2.event_update
pyfmi.fmi.FMUException: Failed to update the events at time: 1.684800E+07.
Writing final SQL reports
```

By using a larger threshold, like changing `m1_flow > -0.1 * m1_flow_nominal` to `m1_flow > -1 * m1_flow_nominal`, still cannot address the error. It might related to the comment `The condition needs to be implicitly treated with noEvent(..) since otherwise events might be triggered that ...` showing in [here](https://build.openmodelica.org/Documentation/ModelicaReference.Operators.%27assert()%27.html). 

If adding `noEvent()` to the condition expressions, the OCT simulation runs without error.

JayHuLBL:
This closes #1690.

### DIFF
--- a/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
+++ b/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
@@ -177,8 +177,8 @@ equation
     end if;
   else
     flowRegime = flowRegime_nominal;
-    assert((m1_flow > -0.1 * m1_flow_nominal)
-       and (m2_flow > -0.1 * m2_flow_nominal),
+    assert(noEvent(m1_flow > -0.1 * m1_flow_nominal)
+       and noEvent(m2_flow > -0.1 * m2_flow_nominal),
 "*** Warning in " + getInstanceName() +
       ": The flow direction reversed.
       However, because the constant use_dynamicFlowRegime is set to false,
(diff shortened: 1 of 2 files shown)


IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
```diff
@@ -253,6 +253,12 @@ for <code>UA</code>.
 </html>", revisions="<html>
 <ul>
 <li>
+February 3, 2023, by Jianjun Hu:<br/>
+Added <code>noEvent()</code> in the assertion function to avoid optimica error.<br/>
```
mwetter, line 257: Better write "to avoid Optimica to not converge" as an error can be many different things. Same for the message below in the other file.

mwetter (changes requested):

mwetter (approved):

PR #1695, opened 2023-02-07
Removed start value for valves

### LINKED ISSUE #1694: Remove dp(start=dpValve_nominal/2) in PartialThreeWayValve
This issue is to port the changes from 
https://github.com/lbl-srg/modelica-buildings/issues/3231
to the IBPSA library. See the above link for the rationale, and the PR for the effects across the test cases of the Buildings Library:
https://github.com/lbl-srg/modelica-buildings/pull/3232

mwetter:
This closes #1694


Mathadon (approved):
@mwetter in the Buildings issue you state 

> will lead on branch IBPSASync_issue1412_stateSelect_Tp, commit https://github.com/lbl-srg/modelica-buildings/commit/d361f828e4e80ffb8f60294c9bd0edc4ef5c9ea9, the models Buildings.Examples.Tutorial.Boiler.System{3,4} to simulate with Optimica, which indicates that it better not be set (these models start with zero mass flow rate).

I don't really follow the rationale here. Overall it makes sense to set the nominal attribute. I see more reasons to set the variables than to not set them. The fact that one tool can/cannot simulate two models is not really sufficient reason the start making such changes in my opinion. I can see us hopping back and forth for years with these kinds of small tweaks that are 'better' for a few models. Ultimately, they are just a patch for the fundamental problem that the solvers don't properly work for some reason. I think we need more fundamental analysis instead of changes based on trial and error. 

If you want to proceed nonetheless with this pull request, go ahead. :)

mwetter:
@Mathadon The main rationale is consistency:
> The model Buildings/Fluid/Actuators/BaseClasses/PartialThreeWayValve.mo is the only one where the start attribute of dp is modified. It is set to dp(start=dpValve_nominal/2). The two-way valves, and the fixed resistances, do not set this start value, and hence it remains at 0.

Hence, as performance is the same, I see no reason why we should handle the valves differently than other flow resistances, in particular because the default value is not helping.

PR #1698, opened 2023-02-09
Removed uses of non-existing parameter in annotation

### LINKED ISSUE #1697: Humidifier_u.mo uses use_T_in which does not exist
The model `Humidifier_u` uses the parameter `use_T_in` in the graphical annotation, but this parameter does not exist in this model.

mwetter:
This closes #1697

### DIFF
--- a/IBPSA/Fluid/Humidifiers/Humidifier_u.mo
+++ b/IBPSA/Fluid/Humidifiers/Humidifier_u.mo
@@ -57,13 +57,6 @@ equation
           textColor={0,0,127},
           textString="u"),
         Rectangle(
-          visible=use_T_in,
-          extent={{-100,-59},{-70,-62}},
-          lineColor={0,0,0},
-          pattern=LinePattern.None,
-          fillColor={191,0,0},
-          fillPattern=FillPattern.Solid),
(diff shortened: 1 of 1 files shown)


mwetter:
@LauraMaier : can you please review this so we can go ahead and merge it. Thanks.

IBPSA/Fluid/Humidifiers/Humidifier_u.mo
```diff
@@ -107,6 +100,12 @@ equation
               {42,-28},{42,-26}},
           lineColor={255,255,255},
           fillColor={255,255,255},
+          fillPattern=FillPattern.Solid),
```
LauraMaier, line 103: Was this change intended?

LauraMaier (approved):
Just a small question that might be unnecessary for the git history. Other than that: all fine

IBPSA/Fluid/Humidifiers/Humidifier_u.mo
mwetter: Yes, this change is good and consistent with the other drawing elements in this icon.

PR #1709, opened 2023-03-03
Added weather files for BESTEST validation

amaccarini:
@mwetter: I added weather files for BESTEST validation according to ANSI/ASHRAE 140-2020. Both .epw and .mos files have been added for `USA_CO_Denver.Intl.AP.725650`


mwetter (commented):
@amaccarini : As BESTEST uses this weather file rather than DRYCOLD, shouldn't this PR also change the weather file in `Case600` and `Case600FF` for the new reduced order model?

amaccarini:
@mwetter: Yes, I will do all the updates and changes in `issue1707_UpdatesISO13790`. Since it may take a few weeks, I thought about opening a separated quick PR only to add the weather files, in case other users need them. But maybe we can just incorporate this PR into `issue1707_UpdatesISO13790`. It is probably unlikely that someone else will need such files within the next few weeks. Just let me know what you think is best.

mwetter:
@amaccarini : I prefer to add the weather files when they are used in an example or validation.

PR #1712, opened 2023-03-06
Added a constant in order for unit check to pass

### LINKED ISSUE #1711: Unit check fails in Psychometric models
In Dymola 2023xRefresh1-beta1, some models fail the unit tests because the result of `Math.atan()` is assigned to a temperature, but `atan()` returns a `Real` with unit set to `rad`.

This issue is to correct these models, such as
```
Utilities/Psychrometrics/TWetBul_TDryBulPhi.mo
Utilities/Psychrometrics/TWetBul_TDryBulXi.mo
```

mwetter:
This closes #1711


cvering (approved):
Thanks @mwetter for the model improvement.

PR #1734, opened 2023-04-27
Moved assignments to paramater declaration

mwetter:
It turns out that the assignment of the `nominal` attribute leads in Dymola to a warning about non-literal nominal values. This addition to the previous pull requests moves parameter assignment from the `initial equation` section to the parameter declaration which avoids this warning.

### DIFF
--- a/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectiveness.mo
+++ b/IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectiveness.mo
@@ -35,11 +35,20 @@ partial model PartialEffectiveness
 protected
   parameter Real delta=1E-3 "Parameter used for smoothing";
 
-  parameter Modelica.Units.SI.SpecificHeatCapacity cp1_default(fixed=false)
+  parameter Modelica.Units.SI.SpecificHeatCapacity cp1_default =
+    Medium1.specificHeatCapacityCp(Medium1.setState_pTX(
+      Medium1.p_default,
+      Medium1.T_default,
+      Medium1.X_default))
     "Specific heat capacity of medium 1 at default medium state";
-  parameter Modelica.Units.SI.SpecificHeatCapacity cp2_default(fixed=false)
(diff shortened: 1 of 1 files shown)


IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectiveness.mo
```diff
@@ -108,8 +107,10 @@ and <code>QMax_flow &gt; 0</code>.
 </html>", revisions="<html>
 <ul>
 <li>
-April 26, 2023, by Michael Wetter:<br/>
+April 27, 2023, by Michael Wetter:<br/>
 Set nominal and min attributes for capacity flow rates.<br/>
+Moved assignments from <code>initial equation</code> to parameter declaration.
+This avoids a warning in Dymola 2023x about non-literal nominal values.<br/>
```
jelgerjansen, line 113: This also holds for Dymola 2022x. We could also just mention Dymola in general here? Or do you prefer to be as specific as possible in case these warnings would show up in a future release of Dymola?

jelgerjansen (changes requested):
Sorry I missed the warnings when checking the previous implementation. Good that you noticed this. The changes look fine by me. I added a small comment/question for the revisions section.

Out of curiosity: Do you still know why, in the previous implementation, the parameters where set using an initial equation in combination with `fixed=false`? Is there an advantage compared to the implementation we have now?

IBPSA/Fluid/HeatExchangers/BaseClasses/PartialEffectiveness.mo
mwetter: I prefer to add the version with which it was encountered in case someone needs to reproduce it.

mwetter: Making the statements in the `initial equation` section helps readability if the assignments are long. The two versions should be equivalent, but in this case, Dymola seem to treat them slightly different. I was surprised that this change removed the warnings.

PR #1741, opened 2023-05-04
Added type cast for pointers

### LINKED ISSUE #1740: Add type casts to malloc and realloc in C-Sources
This adds type casts to pointers. It integrates the changes from https://github.com/lbl-srg/modelica-buildings/pull/3378 and https://github.com/lbl-srg/modelica-buildings/pull/3383 which seem to be needed for some compiler configurations

mwetter:
This closes #1740

### DIFF
--- a/IBPSA/Resources/C-Sources/fileWriterFree.c
+++ b/IBPSA/Resources/C-Sources/fileWriterFree.c
@@ -22,7 +22,7 @@ void prependString(const char* fileName, const char* string){
   if(fseek(fr, 0, SEEK_SET)!=0)
     ModelicaFormatError("The file %s could not be read.", fileName);
 
-  origString = malloc(fsize + 1);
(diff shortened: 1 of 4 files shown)


jelgerjansen (approved):
I'm not experienced with coding in C(++), but a search on the internet teached me that they rather use `new` instead of `malloc()` in C++, as discussed here: https://stackoverflow.com/questions/184537/in-what-cases-do-i-use-malloc-and-or-new. 

However, I don't see a reason to change the current implementation because:

- There is no alternative for `realloc()` when using `new`
- The amount of C code is very limited.
- The current typecasting implementation solves the C++ compiler errors reported in #1740.

Therefore, I agree with the applied changes.

mwetter:
@jelgerjansen : Thanks for the review. Here use of `malloc` is correct as Modelica uses a C interface and `new` is an operator in C++.

Issue #1744, opened 2023-05-05
Option for non-constant TGround in equivalent air temperature (VDI 6007)


PGorzalka:
Hi,

[Issue 1080 of AixLib](https://github.com/RWTH-EBC/AixLib/issues/1080) intends to add non-constant soil temperatures to the reduced-order model. Regarding the RC model itself, this does not affect IBPSA. However, the equivalent air temperature calculation models currently only allow a constant ground temperature parameter.

I would like to suggest adding a variable `TGroundUsed` in `ThermalZones.ReducedOrder.EquivalentAirTemperature.BaseClasses.PartialVDI6007` that is used for the calculations in `VDI6007` and `VDI6007WithWindow` and gets the value from one of the options
- `TGro` (current constant parameter)
- `TGround` (input connector to be added)

depending on another Boolean parameter to be added which is by default pointing to `TGro`.

Looking forward to hearing your thoughts about this.

mwetter:
That would be a good addition. However for consistency with `ReaderTMY3`, please call the input connector `TGro_in`. 
I assume `TGroundUsed` will be a protected variable.

PGorzalka:
Thank you for the feedback. I implemented this in pull request [1749](https://github.com/ibpsa/modelica-ibpsa/pull/1749). As this is my first contribution here, I hope I did it as intended by the guidelines.

PGorzalka:
Being basically finished with this issue (just missing a Dymola update to run the unit tests), I would like to bring up a related topic.

Currently, the model calculates the equivalent temperature lumping surfaces in contact to air and to the soil (where `TGro` that I just inserted as non-constant is only used as contribution by the surfaces in contact to soil):

`TEqAir = TEqWall*wfWall + TGro*wfGro; `
`TEqAirWin = TEqWin*wfWin;`

where
`TEqWin=TDryBul.+delTEqLWWin*(ones(n)-sunblind);`
`TEqWall=TDryBul.+delTEqLW.+delTEqSW;`

where
`delTEqLW=(TBlaSky - TDryBul)*hRad/(hRad + hConWallOut);`
`delTEqLWWin=(TBlaSky - TDryBul)*hRad/(hRad + hConWinOut);`
`delTEqSW=HSol*aExt/(hRad + hConWallOut);`

[...]

PR #1749, opened 2023-05-10
Option for non-constant TGround in equivalent air temperature (VDI 6007)

PGorzalka:
This should solve #1744


mwetter (changes requested):
@PGorzalka : The changes look good. Can you please also add an example or validation test that uses this new option. This way it will become part of the CI testing.

PGorzalka:
@mwetter I added the example including reference results. After running the unit tests locally (only for ROM), `IBPSA.ThermalZones.ReducedOrder.Validation.VDI6007.TestCase7` is only 80 % successful on my machine. I assume this is due to numeric errors/compiler differences, because my changes should not affect that test case at all.

Please let me know if there is anything I missed or did not do as it's supposed to be done.

mwetter:
@PGorzalka : Thanks, I will move it to a development branch and then do a further review, likely finishing next week.

mwetter:
The unit tests look good, the differences are likely due to Windows vs Linux.

Issue #1755, opened 2023-05-19
correct links to other models


mwetter:
The library has various models with entries of the form
```
href=\"IBPSA...
```
rather than
```
href=\"modelica://IBPSA...
```
This issue is to correct these wrong links, and update the `verifyFiles.py` to catch these errors.

PR #1756, opened 2023-05-19
Corrected hyperlinks

mwetter:
For #1755.
This also updates the script bin/verifyFiles.sh to catch such formatting errors


FWuellhorst (approved):
Looks good, thanks for the fix.

PR #1759, opened 2023-05-25
Issue1754 ATES model

amaccarini:
The PR is about a simplified 1D aquifer thermal energy storage model. There are a couple of things that I cannot fully fix:

- When I add `SingleWell` as component, the default unit for pressure is bar. I think it should be Pa, but I am not sure how to fix this. 
- Even though I selected `true` for the parameter about pressure curves, I still get the warning message.


mwetter:
- [ ] How would one compute flow distribution if an ATES system has 3 wells? It looks to me like a flow resistance to the ground from well to well is missing.

IBPSA/Fluid/Geothermal/Aquifer/SingleWell.mo
mwetter, line 1: The boundary condition does not conserve energy within the ATES system.
Wells should be connected to ensure conservation of energy.

amaccarini:
@mwetter I just updated the ATES model. Now it consists of two wells: one cold and one warm. Here a description of the major changes:

- In order to ensure conservation of energy, the wells are connected via fluid ports. The user is informed of the importance of selecting an appropriate domain radius (see documentation)
- The number of wells can be increased by adjusting the parameters `nHot` and `nCoo`. Increasing the number of wells simply lead to a proportional increment of thermal capacity. 
- Circulation pumps are included in the component. They can be controlled by feeding an input signal between [-1,1]. A positive value circulates water clockwise, while a negative value circulates water anticlockwise.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,419 @@
...
+      Dialog(group="Subsurface"));
+  parameter Modelica.Units.SI.Temperature TCoo_start=283.15
+     "Initial temperature of cold well" annotation (
+      Dialog(group="Subsurface"));
+  parameter Modelica.Units.SI.Temperature THot_start=TCoo_start
+     "Initial temperature of warm well" annotation (
+      Dialog(group="Subsurface"));
+  parameter Modelica.Units.SI.Temperature TGroCoo=285.15
+     "Undisturbed ground temperature (cold well)" annotation (
+      Dialog(group="Properties of ground"));
```
mwetter, line 25: What is the rationale for using an initial temperature for the well that is different from the undisturbed ground? Shouldn't they be the same?

mwetter (commented):
@amaccarini Please see initial comment. I am not quite done yet with the review. Overall it looks good.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,419 @@
...
+  parameter Modelica.Units.SI.Density rhoWat(fixed=false)
+    "Water density";
+  parameter Modelica.Units.SI.ThermalConductivity kWat(fixed=false)
+    "Water thermal conductivity";
+  parameter Modelica.Units.SI.HeatCapacity C[nVol](each fixed=false)
+    "Heat capacity of segment";
+  parameter Modelica.Units.SI.Volume VWat[nVol](each fixed=false)
+    "Volumes of water";
+  parameter Modelica.Units.SI.ThermalResistance R[nVol+1](each fixed=false)
+    "Thermal resistances between nodes";
```
mwetter, line 185: `R` has dimension `nVol+1`, but above it is used in
```
  Modelica.Thermal.HeatTransfer.Components.ThermalResistor theResCoo[nVol](
    R=R/nCoo)
```
The dimension must match (OpenModelica catches this mistake).

mwetter (commented):
The dimension of `R` and its use in the resistor are wrong.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,428 @@
...
+    final dp_nominal=dpAquifer_nominal*2 + dpWell_nominal*2 + dpExt_nominal)
+    "Pump to extract from hot well" annotation (Placement(transformation(
+        extent={{-10,-10},{10,10}},
+        rotation=90,
+        origin={80,20})));
+
+  IBPSA.Fluid.MixingVolumes.MixingVolume volCoo[nVol](
+    redeclare final package Medium = Medium,
+    each final T_start=TCoo_start,
+    each final m_flow_nominal=m_flow_nominal,
```
mwetter, line 82: I suggest to make all these components below protected as a user typically does not need to see these result. If so, we can make a temperature vector `TCoo` and `THot` that has the temperatures of the control volumes so a user can look at them. All other results I doubt a user needs to see (and if so, one could simulate the model with reporting protected variables).

IBPSA/Fluid/Geothermal/Aquifer/Data/Template.mo
```diff
@@ -0,0 +1,31 @@
+within IBPSA.Fluid.Geothermal.Aquifer.Data;
+record Template
+  "Template for soil data records"
+  extends Modelica.Icons.Record;
+  parameter Modelica.Units.SI.ThermalConductivity kSoi
+    "Thermal conductivity of the soil material";
+  parameter Modelica.Units.SI.SpecificHeatCapacity cSoi
+    "Specific heat capacity of the soil material";
+  parameter Modelica.Units.SI.Density dSoi(displayUnit="kg/m3")
+    "Density of the soil material";
```
mwetter, line 10: This data record should also include the porosity `phi` as it is a property of the soil, and not of the well model.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,428 @@
...
+<p>
+By default, the component consists of a cold well and a warm well. The number of wells can be increased by modifing the parameters <code>nCoo</code> and <code>nHot</code>.
+The effect is a proportional increase of thermal capacity.
+</p>
+<p>
+To ensure conservation of energy, the two wells are connected via fluid ports. To avoid thermal interferences, make sure that the aquifer domain radius <code>r_max</code> is large enough for your specific use case.
+</p>
+<p>
+Circulation pumps are included in the model and they can be controlled by acting on the input connector. The input must vary between [1,-1]. A positive value will circulate water
+clockwise (extraction from the cold well and injection into the warm well). A negative value will circulate water anticlockwise (extraction from the warm well and injection into the cold well).
```
mwetter, line 418: Here you should also explain the design pressure drops and how they are used to size the pump.

mwetter (changes requested):
@amaccarini : I think the model is close to being finished. Please see the inline comments. 
If we introduce an array of temperatures, we could display the temperature of the volume that is connected to the pumps (using code such as
```
        Text(
          extent={{-20,120},{-140,70}},
          textColor={0,0,0},
          textString=DynamicSelect("", String(T[0]-273.15, format=".1f")))}),
```
(see for example the temperature sensors in the library.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
amaccarini: I assume that in some cases the user might be interested in performing a seasonal simulation starting from predefined initial temperatures in the aquifer. For example, one may carry out a winter simulation, and assume that the warm aquifer is fully charged at 30°C. However, the undisturbed ground temperature would still be at around 10°C.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
amaccarini: It should be ok now

IBPSA/Fluid/Geothermal/Aquifer/Data/Template.mo
amaccarini: Done

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,419 @@
...
+    final m_flow_nominal=m_flow_nominal,
+    final dp_nominal=dpWell_nominal)
+    "Pressure drop in the warm well" annotation (
+      Placement(transformation(
+        extent={{10,10},{-10,-10}},
+        rotation=-90,
+        origin={80,50})));
+  Movers.Preconfigured.SpeedControlled_y pumCol(
+    redeclare final package Medium = Medium,
+    addPowerToMedium=false,
```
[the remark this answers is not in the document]
amaccarini: Done

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,419 @@
...
+      Dialog(group="Properties of ground"));
+  parameter Modelica.Units.SI.Temperature TGroHot=TGroCoo
+     "Undisturbed ground temperature (warm well)" annotation (
+      Dialog(group="Properties of ground"));
+  parameter IBPSA.Fluid.Geothermal.Aquifer.Data.Template aquDat
+     "Aquifer thermal properties" annotation (choicesAllMatching=true);
+  parameter Modelica.Units.SI.MassFlowRate m_flow_nominal "Nominal mass flow rate" annotation (
+      Dialog(group="Hydraulic circuit"));
+  parameter Modelica.Units.SI.PressureDifference dpAcquifer_nominal(displayUnit="Pa")
+    "Pressure drop at nominal mass flow rate in the aquifer" annotation (
```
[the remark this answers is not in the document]
amaccarini: I changed the pressure drops multipliers, and added a figure that illustrates what these resistances represent.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,420 @@
...
+  parameter Modelica.Units.SI.Temperature TGroHot=TGroCoo
+     "Undisturbed ground temperature (warm well)" annotation (
+      Dialog(group="Properties of ground"));
+  parameter IBPSA.Fluid.Geothermal.Aquifer.Data.Template aquDat
+     "Aquifer thermal properties" annotation (choicesAllMatching=true);
+  parameter Modelica.Units.SI.MassFlowRate m_flow_nominal "Nominal mass flow rate" annotation (
+      Dialog(group="Hydraulic circuit"));
+  parameter Modelica.Units.SI.PressureDifference dpAquifer_nominal(displayUnit=
+        "Pa") "Pressure drop at nominal mass flow rate in the aquifer"
+                                                             annotation (
```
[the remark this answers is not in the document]
amaccarini: Done (see above).

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
amaccarini: Do you mean to make below protected only the `MixingVolume` components or also `HeatCapacitor` and `ThermalResistor` components?

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
amaccarini: I think that the new illustration should be able to explain the rationale behind the selection of the design pressure drops.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
```diff
@@ -0,0 +1,419 @@
...
+  parameter Integer nVol(min=1)=10 "Number of control volumes used in discretization" annotation (
+      Dialog(group="Subsurface"));
+  parameter Modelica.Units.SI.Height h=200 "Aquifer thickness";
+  parameter Real phi(final unit="1")=0.2 "Reservoir porosity";
+  parameter Real nCoo=1 "Number of cold wells";
+  parameter Real nHot=1 "Number of warm wells";
+  parameter Modelica.Units.SI.Radius r_wb=0.1 "Wellbore radius" annotation (
+      Dialog(group="Subsurface"));
+  parameter Modelica.Units.SI.Radius r_max=2400 "Domain radius" annotation (
+      Dialog(group="Subsurface"));
```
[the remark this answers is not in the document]
amaccarini: It is not clear why `RMax` has upper-case R, while `rWB` has lower-case r.

IBPSA/Fluid/Geothermal/Aquifer/Examples/CoolingOffice.mo
```diff
@@ -0,0 +1,75 @@
...
+    nVol=80,
+    h=20,
+    r_max=500,
+    TCoo_start=285.15,
+    THot_start=285.15,
+    aquDat=IBPSA.Fluid.Geothermal.Aquifer.Data.Rock(),
+    m_flow_nominal=mWat_flow_nominal,
+    dpAcquifer_nominal = 10,
+    dpWell_nominal = 10,
+    dpExt_nominal = 100)
```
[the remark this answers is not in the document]
amaccarini: Working on it (I would like to find some references for these values).

IBPSA/Fluid/Geothermal/Aquifer/Examples/CoolingOffice.mo
```diff
@@ -0,0 +1,75 @@
...
+    "Cooling power";
+  parameter Modelica.Units.SI.TemperatureDifference deltaT=4 "Temperature difference at heat exchanger";
+  parameter Modelica.Units.SI.SpecificHeatCapacity cpWat=4186 "Heat capacity of water";
+  parameter Modelica.Units.SI.MassFlowRate mWat_flow_nominal=
+    QCoo_flow_nominal/(deltaT*cpWat) "Nominal water mass flow rate";
+
+  HeatExchangers.HeaterCooler_u hea(
+    redeclare package Medium = IBPSA.Media.Water,
+    m_flow_nominal=mWat_flow_nominal,
+    dp_nominal=100,
```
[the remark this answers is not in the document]
amaccarini: Working on it (I would like to find some references for these values).

IBPSA/Fluid/Geothermal/Aquifer/SingleWell.mo
amaccarini: This has been addressed by connecting the two wells via fluid ports.

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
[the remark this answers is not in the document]
amaccarini: Ok, we talked about this during the meeting yesterday. It is `rMax` and `rWB`

IBPSA/Fluid/Geothermal/Aquifer/MultiWell.mo
amaccarini: Discuss yesterday at the meeting. I made below protected volumes, capacitors and resistances and I created two temperature arrays `TAquHot` and `TAquCol` to output the temperatures in the aquifer. But I removed from protected `rC` as I think that users should be able to see at which distance these temperature are calculated.

amaccarini: I just realized that (obviously), by removing `rC` from protected, this appears in the parameter dialog box, and it shouldn't. Therefore, I created a new variable `rVol`, which is simply equal to `rC`, but it is accessible to users. The alternative would be to have `rC` directly as variable, but then I don't think we can have it under `initial equation`

amaccarini:
@mwetter: I updated the ATES model following our latest discussions. In particular, I added the calculation of the pressure drop from the Darcy's law for the parameter `dpAquifer_nominal`. 

However, I just realized that the calculation of pressure drops when having multiple wells is probably not correct, as these should be scaled according to the number of wells. For example see `Aquifer.Validation.NumberWells`. When simulating the model, one would expect that the power consumption of the pumps in `aquWell2` are double compared to `aquWell1`. But results show that these are about 4 times bigger. I think we need to introduce parameters defined as `nCoo/(nCoo+nHot)` and `nHot/(nCoo+nHot)` and use them in `MultiWell.powCoo` and `MultiWell.powHot`. But I am still a bit unsure on how to do something similar for `MultiWell.resCoo` and `MultiWell.resHot`. I will take a look on Monday. Please let me know if you have any suggestion.

mwetter:
@amaccarini : I think the model is close to being finished.
I changed the model to set the flow resistance on one side to be the same as on the other side, which reduces the amount of equations, and leads to a 1x1 rather than a 2x2 nonlinear system of equations for the initialization.

I have one item I like to discuss, which we can do next week:
- [ ] I think using `rMax` in the computation of `dpAquifer_nominal` is not correct: `rMax` is used for the heat transfer domain, but the wells could be farther away than `2*rMax` in which case the resistance should be larger.

amaccarini:
@mwetter Following our discussion last week, I introduced a new parameter `d` which is the distance between wells. This parameter is used to compute the pressure losses (instead of `rMax`). The value of `rMax` is set to `d/2` by default.

mwetter:
@amaccarini : Should `rMax <= d/2` be enforced with an assertion? I noted for example in `Examples.CoolingOffice`, that `d=200` and `rMax=500`, which I think is not a meaningful configuration.

Also, I assume `rMax` is the same as `r_{Aquifer}` in the graphics of the `Multiwell` documentation. This should be consistent.

amaccarini:
@mwetter: Yes, that sounds like a good idea. I added an assertion to verify that `rMax <= d/2`. I also updated the image by using the names of parameters `rWB` and `rMax`

PR #1782, opened 2023-09-03
Issue1781 port placement

### LINKED ISSUE #1781: Parameterization of port placement for interface classes in Fluid
This is a feature enhancement request to allow specifying the location for outside connectors in interface classes.
Currently, if one want to increase the icon size when extending such classes the only option is using `IconMap`. 
However, this rescales the connector icons as well and connection lines in the diagram layer are also improperly set by some Modelica tools like in the screenshot below (Dymola diagram view).

Capture d’écran 2023-08-28 à 10 09 38

For instance the `Fluid.Interfaces.PartialTwoPort` will be modified with:
```
  constant Integer port_a_x = -100
    "x-coordinate of port_a center";
  constant Integer port_a_y = 0
    "y-coordinate of port_a center";

  Modelica.Fluid.Interfaces.FluidPort_a port_a
    ...
    "Fluid connector a (positive design flow direction is from port_a to port_b)"
    annotation (Placement(transformation(extent={{port_a_x-10,port_a_y-10},{port_a_x+10,port_a_y+10}})));
```
(Using a constant instead of a parameter does not result in any change in the parameter dialogs and the constant is not part of an FMU export.)

The modifications are ***backward compatible***.

AntoineGautier:
This closes #1781.

I also proposed in commit 60c19a778 to modify port placement in `Fluid/Storage/BaseClasses/PartialStratified.mo`. Although it may annoy a few users (whose model diagrams will be modified) it will benefit many future users (with a more intuitive icon layout). But I can also revert that change if the disturbance is too big.

### DIFF
--- a/IBPSA/Fluid/Interfaces/PartialFourPort.mo
+++ b/IBPSA/Fluid/Interfaces/PartialFourPort.mo
@@ -22,23 +22,6 @@ partial model PartialFourPort "Partial model with four ports"
           X_a=0.40)
           "Propylene glycol water, 40% mass fraction")));
 
-  constant Integer port_a1_x = -100
-    "x-coordinate of port_a1 center";
-  constant Integer port_a1_y = 60
-    "y-coordinate of port_a1 center";
-  constant Integer port_b1_x = 100
-    "x-coordinate of port_b1 center";
-  constant Integer port_b1_y = 60
-    "y-coordinate of port_b1 center";
-  constant Integer port_a2_x = 100
-    "x-coordinate of port_a2 center";
-  constant Integer port_a2_y = -60
-    "y-coordinate of port_a2 center";
-  constant Integer port_b2_x = -100
-    "x-coordinate of port_b2 center";
-  constant Integer port_b2_y = -60
-    "y-coordinate of port_b2 center";
-
   parameter Boolean allowFlowReversal1 = true
     "= false to simplify equations, assuming, but not enforcing, no flow reversal for medium 1"
     annotation(Dialog(tab="Assumptions"), Evaluate=true);
@@ -71,6 +54,24 @@ partial model PartialFourPort "Partial model with four ports"
                      h_outflow(start = Medium2.h_default, nominal = Medium2.h_default))
     "Fluid connector b2 (positive design flow direction is from port_a2 to port_b2)"
     annotation (Placement(transformation(extent={{port_b2_x-10,port_b2_y-10},{port_b2_x+10,port_b2_y+10}})));
+
+protected
+  constant Integer port_a1_x = -100
+    "x-coordinate of port_a1 center";
+  constant Integer port_a1_y = 60
+    "y-coordinate of port_a1 center";
+  constant Integer port_b1_x = 100
+    "x-coordinate of port_b1 center";
+  constant Integer port_b1_y = 60
+    "y-coordinate of port_b1 center";
+  constant Integer port_a2_x = 100
+    "x-coordinate of port_a2 center";
+  constant Integer port_a2_y = -60
+    "y-coordinate of port_a2 center";
+  constant Integer port_b2_x = -100
+    "x-coordinate of port_b2 center";
+  constant Integer port_b2_y = -60
+    "y-coordinate of port_b2 center";
 
   annotation (
     preferredView="info",
--- a/IBPSA/Fluid/Interfaces/PartialFourPort.mo
+++ b/IBPSA/Fluid/Interfaces/PartialFourPort.mo
@@ -22,6 +22,23 @@ partial model PartialFourPort "Partial model with four ports"
           X_a=0.40)
           "Propylene glycol water, 40% mass fraction")));
 
+  constant Integer port_a1_x = -100
+    "x-coordinate of port_a1 center";
+  constant Integer port_a1_y = 60
+    "y-coordinate of port_a1 center";
+  constant Integer port_b1_x = 100
+    "x-coordinate of port_b1 center";
+  constant Integer port_b1_y = 60
+    "y-coordinate of port_b1 center";
+  constant Integer port_a2_x = 100
+    "x-coordinate of port_a2 center";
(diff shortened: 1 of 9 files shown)


mwetter (approved):

mwetter:
@FWuellhorst : Can you please do a review of this feature so we have a non-LBNL person reviewing it. Thanks.

IBPSA/Fluid/Interfaces/PartialFourPort.mo
```diff
@@ -22,6 +22,23 @@ partial model PartialFourPort "Partial model with four ports"
...
+    "y-coordinate of port_b1 center";
+  constant Integer port_a2_x = 100
+    "x-coordinate of port_a2 center";
+  constant Integer port_a2_y = -60
+    "y-coordinate of port_a2 center";
+  constant Integer port_b2_x = -100
+    "x-coordinate of port_b2 center";
+  constant Integer port_b2_y = -60
+    "y-coordinate of port_b2 center";
+
```
FWuellhorst, line 41: I guess these could also be in the protected section? When debugging code, I sometimes find myself back in these partial classes. These are the first 20 lines have no importance for understanding the model and could confuse others.

IBPSA/Fluid/Interfaces/PartialTwoPortVector.mo
```diff
@@ -11,6 +11,16 @@ partial model PartialTwoPortVector "Partial component with two ports, one of whi
...
+
+  constant Integer port_a_x = -100
+    "x-coordinate of port_a center";
+  constant Integer port_a_y = 0
+    "y-coordinate of port_a center";
+  constant Integer port_b_x = 100
+    "x-coordinate of port_b center";
+  constant Integer port_b_y = 0
+    "y-coordinate of port_b center";
+
```
FWuellhorst, line 23: Did you consider a partial model (not only for this)? `PartialPort` with `port_a` and `Medium` and `PartialTwoPort` and `PartialTwoPortVector` extending those?

IBPSA/Fluid/Storage/Examples/StratifiedUnloadAtMinimumTemperature.mo
```diff
@@ -123,12 +124,14 @@ equation
...
-  connect(valBot.port_a, tan.fluPorVol[5]) annotation (Line(points={{150,-100},{
-          -112.6,-100},{-112.6,-120}},
+  connect(valTop.port_a, tan.fluPorVol[1]) annotation (Line(points={{112,-20},{
+          -116,-20},{-116,-118},{-112.6,-118},{-112.6,-118.4}},
+                               color={0,127,255}));
+  connect(valMid.port_a, tan.fluPorVol[3]) annotation (Line(points={{132,-60},{
+          -116,-60},{-116,-118},{-112.6,-118}},
+                              color={0,127,255}));
+  connect(valBot.port_a, tan.fluPorVol[5]) annotation (Line(points={{150,-100},
+          {-116,-100},{-116,-118},{-112.6,-118},{-112.6,-117.6}},
```
FWuellhorst, line 134: @mwetter Just a question: Why doesn't git show the real diff? It looks like functional changes at the first glance, but only the annotation changed. Some save-settings in Dymola?

IBPSA/Fluid/Storage/Stratified.mo
```diff
@@ -157,54 +157,6 @@ Icon(graphics={
...
-          extent={{-76,2},{-90,-2}},
-          lineColor={0,0,255},
-          pattern=LinePattern.None,
-          fillColor={0,0,127},
-          fillPattern=FillPattern.Solid),
-        Rectangle(
-          extent={{0,84},{-80,80}},
-          lineColor={0,0,255},
-          pattern=LinePattern.None,
-          fillColor={0,0,127},
```
FWuellhorst, line 170: To be consistent with the other ports, why not put the middle one to the left?
image

FWuellhorst (commented):
Looks good, nice feature. 
I have some comments / questions. Moving to protected would be a viable option, I think.

IBPSA/Fluid/Interfaces/PartialFourPort.mo
AntoineGautier: Good point: I put these declarations in a protected section.

IBPSA/Fluid/Interfaces/PartialTwoPortVector.mo
AntoineGautier: I would prefer not to. Although this would avoid (little) code duplication between `PartialTwoPort` and `PartialTwoPortVector` it would add an inheritance level in interface classes that already derive from each other. For instance, one looking at a simple actuator model such as `Dampers.Exponential` already has to navigate across `PartialDamperExponential`→`PartialResistance`→`PartialTwoPortInterface`→`PartialTwoPort` to travel the whole definition. Adding a new layer that does not abstract different physics may be more cumbersome than helpful.

IBPSA/Fluid/Storage/Stratified.mo
AntoineGautier: I tested it but it looked weird in diagrams, especially when the middle port is left unconnected as in `IBPSA.Fluid.Storage.Examples.Stratified`.
However, I committed some modification to use `FluidPorts_a` instead of `FluidPort_a` to provide correct graphical feedback regarding the dimensionality of this port. 
I also relocated the port icon on the side of the tank.

FWuellhorst (approved):
Thanks for the changes, looks really good!

IBPSA/Fluid/Interfaces/PartialTwoPortVector.mo
FWuellhorst: I agree. I got the impression that you prefer inheritance over code repetition at the LBNL, thats why I asked.

AntoineGautier:
@mwetter This is ready to merge.

PR #1790, opened 2023-09-14
Issue1744 vdi6007 non constant t ground

### LINKED ISSUE #1744: Option for non-constant TGround in equivalent air temperature (VDI 6007)

mwetter:
This closes #1744


mwetter:
@PGorzalka: can you please have a look at my changes, if you agree it is ready to be merged.

PGorzalka:
@mwetter Thank you for the improvements. Looks good to me!

Issue #1794, opened 2023-09-20
Constants for connector coordinates not properly handled by Modelica tools


AntoineGautier:
The bug results from #1782 which introduced some constants to parameterize the placement of outside connectors in interface classes within `Fluid`.
Dymola (2023x Refresh1 on Linux and Windows) 

- does not properly handle these constants when extending such interface classes, as illustrated in the example below,
- (even worse!) modifies existing annotations when opening models that derive from these classes: for instance, open `IBPSA.Fluid.Chillers.BaseClasses.Carnot` and notice how these lines https://github.com/ibpsa/modelica-ibpsa/blob/386a90ab66835be5c38d84566358d85b27f51667/IBPSA/Fluid/Chillers/BaseClasses/Carnot.mo#L237-L238 are modified into 
  ```
  connect(port_a2, eva.port_a)
    annotation (Line(points={{0,0},{10,0},{10,-60}},       color={0,127,255}));
  ```
OMEdit (OpenModelica 1.21.0 on Linux) 
- does not properly handle these constants when extending such interface classes, as illustrated in the example below,
- but does not modify existing models like Dymola does.

I have opened a ticket at Dassault Systemes: SRF01088076 (status: wait for correction delivery). 
However, I see no immediate workaround (changing these constants to parameters and/or making them public has no effect), and I suggest that we readily revert the changes from #1782 as Dymola's behavior is likely to have severe consequences.

@mwetter or @FWuellhorst Can you reproduce what I noticed on your system?

----
Example illustrating the issue: Connecting `Component.port_a` to `Component.sub.port_a` from the diagram view of Dymola yields an incorrect graphical annotation.

```mo
package Test
model Base
  Modelica.Fluid.Interfaces.FluidPort_a port_a
    "Fluid connector a (positive design flow direction is from port_a to port_b)"
    annotation (Placement(transformation(extent={{port_a_x-10,port_a_y-10},{port_a_x+10,port_a_y+10}})));
  Modelica.Fluid.Interfaces.FluidPort_b port_b
    "Fluid connector b (positive design flow direction is from port_a to port_b)"
    annotation (Placement(transformation(extent={{port_b_x-10,port_b_y-10},{port_b_x+10,port_b_y+10}})));
  protected 
  constant Integer port_a_x = -100
    "x-coordinate of port_a center";
  constant Integer port_a_y = 0
    "y-coordinate of port_a center";
  constant Integer port_b_x = 100
[...]

PR #1795, opened 2023-09-20
Revert most of the changes from #1782

AntoineGautier:
This addresses #1794.
From PR #1782, we just keep the modification described at https://github.com/ibpsa/modelica-ibpsa/pull/1782#discussion_r1318335036.

### DIFF
--- a/IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
+++ b/IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
@@ -1,0 +1,113 @@
+within IBPSA.Fluid.Storage.BaseClasses;
+partial model PartialTwoPortInterface
+  "Partial model transporting fluid between two ports without storing mass or energy"
+
+  replaceable package Medium =
+    Modelica.Media.Interfaces.PartialMedium "Medium in the component"
+      annotation (choices(
+        choice(redeclare package Medium = IBPSA.Media.Air "Moist air"),
+        choice(redeclare package Medium = IBPSA.Media.Water "Water"),
+        choice(redeclare package Medium =
+            IBPSA.Media.Antifreeze.PropyleneGlycolWater (
+              property_T=293.15,
+              X_a=0.40)
+              "Propylene glycol water, 40% mass fraction")));
+
+  parameter Boolean allowFlowReversal = true
+    "= false to simplify equations, assuming, but not enforcing, no flow reversal"
+    annotation(Dialog(tab="Assumptions"), Evaluate=true);
+
+  Modelica.Fluid.Interfaces.FluidPort_a port_a(
+    redeclare final package Medium = Medium,
+    m_flow(min=if allowFlowReversal then -Modelica.Constants.inf else 0),
+    p(start=Medium.p_default),
+    h_outflow(start = Medium.h_default, nominal = Medium.h_default))
+    "Fluid connector a (positive design flow direction is from port_a to port_b)"
+    annotation (Placement(transformation(extent={{-10,90},{10,110}})));
+  Modelica.Fluid.Interfaces.FluidPort_b port_b(
+    redeclare final package Medium = Medium,
+    m_flow(max=if allowFlowReversal then +Modelica.Constants.inf else 0),
+    p(start=Medium.p_default),
+    h_outflow(start = Medium.h_default, nominal = Medium.h_default))
+    "Fluid connector b (positive design flow direction is from port_a to port_b)"
+    annotation (Placement(transformation(extent={{10,-110},{-10,-90}})));
+
+  parameter Modelica.Units.SI.MassFlowRate m_flow_nominal
+    "Nominal mass flow rate" annotation (Dialog(group="Nominal condition"));
+  parameter Modelica.Units.SI.MassFlowRate m_flow_small(min=0) = 1E-4*abs(
+    m_flow_nominal) "Small mass flow rate for regularization of zero flow"
+    annotation (Dialog(tab="Advanced"));
+  // Diagnostics
+   parameter Boolean show_T = false
+    "= true, if actual temperature at port is computed"
+    annotation (
+      Dialog(tab="Advanced", group="Diagnostics"),
+      HideResult=true);
+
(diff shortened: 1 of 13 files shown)


mwetter:
@AntoineGautier @FWuellhorst : 
Can you please review my latest changes, in which I introduced a base class so the tank ports remain at the top and bottom, and also checked out the examples and validations from the master branch in which the connection lines have been placed accordingly.
I place the new class in `Fluid.Storage.BaseClasses` as I would think once this tool limitations are corrected, it can be removed again and the original implementation that is cleaner be added back.

IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
```diff
@@ -0,0 +1,113 @@
+within IBPSA.Fluid.Storage.BaseClasses;
+partial model PartialTwoPortInterface
+  "Partial model transporting fluid between two ports without storing mass or energy"
```
AntoineGautier, line 3: Isn't the description "transporting fluid between two ports without storing mass or energy" out of the scope of this model that does not include any transport or balance equation?
I would rather describe this model as "Partial model with two ports and generic variables".

It seems that `IBPSA.Fluid.Interfaces.PartialTwoPortInterface` has the same issue. In addition, the documentation of `IBPSA.Fluid.Interfaces.PartialTwoPortInterface` includes:

> It is similar to Modelica.Fluid.Interfaces.PartialTwoPortTransport, but it does not include the species balance

which I think is not consistent with the model definition, and should be replaced with:

> It is similar to Modelica.Fluid.Interfaces.PartialTwoPort, but contains additional declarations for physical variables of common use.

IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
```diff
@@ -0,0 +1,113 @@
...
+        "Pa") = 0
+    "Start value for dp, used to avoid a warning if not set in dp, and to avoid dp.start in parameter window";
+
+  annotation (
+    preferredView="info",
+    Documentation(info="<html>
+<p>
+This partial class implements the same functionality as 
+<a href=\"modelica://IBPSA.Fluid.Interfaces.StaticTwoPortHeatMassExchanger\">
+IBPSA.Fluid.Interfaces.StaticTwoPortHeatMassExchanger</a>,
```
AntoineGautier, line 90: Isn't it rather `IBPSA.Fluid.Interfaces.PartialTwoPortInterface`?

FWuellhorst, line 90: I don't get why the last version extends `PartialTwoPort` and now you duplicated `StaticTwoPortHeatMassExchanger`

FWuellhorst (approved):
Looks good, only one question.

AntoineGautier (commented):
See my inline comments.
I also added a commit (89217fb) to reroute some connections and increase the size of the diagram layer in `StratifiedEnhancedInternalHex` (now that it is no longer horizontally constrained).

IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
mwetter: I don't understand your comment. The implementation in this PR implements `PartialTwoPortInterface` and does not duplicate `StaticTwoPortHeatMassExchanger`. 

My new version of `PartialTwoPortInterface` implements however also `PartialTwoPort` directly inside it, as I did not want to have multiple layers of inheritance and never use all layers by themselves only.

IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
mwetter: The description "without storing mass or energy" is to distinguish the model from
```modelica
model TwoPortHeatMassExchanger
  "Partial model transporting one fluid stream with storing mass or energy"
...
```
which instantiates a control volume. So, this base class does not store mass or energy, but a model that extends it may add these equations, as the tank does.

In `PartialTwoPortInterface` I would not add "...does not include the species balance" as `PartialTwoPortTransport` also adds other balances (pressure, mass, trace substances). Also "...variables of common use" is in my opinion quite vague as it is not clear what common use means.

IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
mwetter: You are right, I will correct this (78acfb24f)

FWuellhorst: It was basically the same comment as Jelger down below, I just confused `PartiaTwoPortInferface` with `PartiaTwoPort`.

IBPSA/Fluid/Storage/BaseClasses/PartialTwoPortInterface.mo
AntoineGautier: My comment is more about "***transporting fluid*** between two ports". I believe this is not a fair description of this model, since we don't have any of the following equations, as in `PartialTwoPortTransport` for example.
```
  // Mass balance (no storage)
  port_a.m_flow + port_b.m_flow = 0;
  // Transport of substances
  port_b.Xi_outflow = inStream(port_a.Xi_outflow);
```
This means that models extending this base class may not transport fluid at all. 
The base class essentially provides two fluid ports and a set of variables. In this respect, it is closer to `PartialTwoPort` than to `TwoPortHeatMassExchanger`, and this "set of variables" makes the difference with `PartialTwoPort` (there is no additional equation).
However, this is not a blocking point and it should not prevent from merging this PR.

mwetter: I see. Let me do a separate PR for clarifying the documentation so we can go ahead and merge this PR.

mwetter: To be addressed through https://github.com/ibpsa/modelica-ibpsa/issues/1796

PR #1808, opened 2023-10-24
Updated documentation of Examples package

mwetter:
This adds more information to the top-level documentation of IBPSA.Examples.
The changes have been made so we can automatically merge IBPSA to Buildings without overriding the documentation that Buildings had, and which was not added when we integrated the `Tutorial` into the IBPSA library.


amaccarini (approved):
It looks good.

PR #1815, opened 2023-11-22
Corrected wrong use of getInstanceName()

### LINKED ISSUE #1814: Correct wrong use of getInstanceName()
In the package `IBPSA.Fluid.Geothermal.Borefields.BaseClasses.Boreholes.BaseClasses.Functions`, the functions `internalResistancesOneUTube` and `internalResistancesTwoUTube` call `getInstanceName()`.
However, per the MLS, this function can only be called from models/blocks, but not from functions.

OpenModelica returns for these functions
```
InternalResistancesOneUTube_functions.c:718:43: warning: implicit declaration of function 'getInstanceName' is invalid in C99
```
The work-around is to pass the instance name as an argument to the function, and add a default argument name so that adding the argument to the function is backwards compatible.

mwetter:
This closes #1814.
This function cannot be called inside a function.

### DIFF
--- a/IBPSA/Fluid/Geothermal/Borefields/BaseClasses/Boreholes/BaseClasses/Functions/partialInternalResistances.mo
+++ b/IBPSA/Fluid/Geothermal/Borefields/BaseClasses/Boreholes/BaseClasses/Functions/partialInternalResistances.mo
@@ -29,6 +29,8 @@ partial function partialInternalResistances
   input Modelica.Units.SI.SpecificHeatCapacity cpMed
     "Specific heat capacity of the fluid";
   input Modelica.Units.SI.MassFlowRate m_flow_nominal "Nominal mass flow rate";
+  input String instanceName "undeclared caller"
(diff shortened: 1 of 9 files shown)


IBPSA/Fluid/Geothermal/Borefields/BaseClasses/Boreholes/BaseClasses/Functions/partialInternalResistances.mo
```diff
@@ -29,6 +29,8 @@ partial function partialInternalResistances
   input Modelica.Units.SI.SpecificHeatCapacity cpMed
     "Specific heat capacity of the fluid";
   input Modelica.Units.SI.MassFlowRate m_flow_nominal "Nominal mass flow rate";
+  input String instanceName="undeclared caller"
+    "Instance name of the model or block that calls this function";
```
FWuellhorst, line 33: I know this exact name is used in other models as well, but why the exception to the 3 letter scheme?

PR #1820, opened 2023-12-18
Issue1576 heat pump model integration remove cp and rho parameters

mwetter:
This sets `cp` and `rho` as `final` based on the selected media.
This is in response to https://github.com/ibpsa/modelica-ibpsa/pull/1628#discussion_r1429645650

The issue about the nonliteral default value is not really an issue of the heat pump model (after setting https://github.com/ibpsa/modelica-ibpsa/compare/issue1576_heatPumpModelIntegration...issue1576_heatPumpModelIntegration_removeCp_and_rho_parameters?expand=1#diff-f521b14773b678fda6b78bb210ce273d2f54b10b629871e256079568a61812e7R241) but rather caused by using the `heaPum.mCon_flow_nominal` from the heat pump (this is the non-literal value) and reusing it to set values in other components that then report about the non-literal value.
To avoid this, one could introduce a parameter for the design flow rate, and then use it.

The change in the reference results is very small (due to numeric).

There are however some warning such as
```
Warning: The following was detected at time: 0
  In ModularReversible_OneRoomRadiator.heaPum.refCyc.refCycHeaPumCoo: The deviation of the given mEva_flow_nominal to the table data is 80.2387 %. Carefully check results,
      you are extrapolating the table data.
  Failed condition: heaPum.refCyc.refCycHeaPumCoo.perDevMasFloEva < 10
```
in `HeatPumps.ModularReversible.Examples.ModularReversible_OneRoomRadiator`. Can you please see if these, and the ones from the pumps, can be removed, as an 80% difference is huge which questions the validity of the example.

Please also check in `ReversibleAirToWater2D_OneRoomRadiator` the pressure drop of the evaporator. Shouldn't this be 200 Pa rather than 2000 Pa?


FWuellhorst:
Checking `IBPSA.Fluid.HeatPumps.ModularReversible.Examples.ModularReversible_OneRoomRadiator` I still get the warning due to `heaPum.mEva_flow_nominal`  being used in the pump. 

Your solution assumes a cp of 1000 and a temperature spread of 10 K, which are not necessarily good assumptions, if the nominal attribute has an impact on the pressure drop model and the simulation. If it does not have an influence, could we avoid setting the nominal attribute all together or simply use the default you proposed in here? I know a lot of users who calculate their m_flow and Q_flow_nominal values given other parameters, and the warnings always raise the questions if something is wrong.

FWuellhorst:
Regarding the other comments:
I used the dp of 2000 Pa from the model `IBPSA.Fluid.HeatPumps.Examples.ScrollWaterToWater_OneRoomRadiator`.
I can change it to 200 Pa, which sounds like a better default. Just to note that the other example might need a change, too.

Regarding the deviation warning: I added the warning to indicate a possible mismatch in nominal flow rates. It does not necessarily mean that the model in invalid, as the mass flow rates in each point of the table may vary due to different heat capacities and temperature spreads. I can update the warning to only warn if the nominal mass flow rate is outside the whole range of possible mass flow rates, indicates temperature spreads which are to large or to small.

mwetter:
@FWuellhorst : It is fine for the nominal *attribute* to assume a cp of 1000, as the nominal attribute is only used in the solver to scale the residual, so using 1000 or 4200 makes little if any difference. As the nominal attribute is set in the base class of the pressure drop, and the heat pump it depends on the nonliteral value for cp, we have to set it with an expression that can be evaluated by Dymola at compile time.

Regarding the pressure drop of the evaporator, please set it in your model to 200 Pa. Note that `ScrollWaterToWater_OneRoomRadiator` is fine as this is water-to-water, and for water, the pressure drop is more likely around a few kilo Pascals.

FWuellhorst:
@mwetter Ok, I see the point. I still think it would benefit users to not have it final. But I can use a workaround if you are against it.

I fixed the assertion warning by only warning users if the mass flow rate is to high or to low for reasonable temperature spreads (3-10 K) according to EN 14511.

mwetter:
I would make the binding of `cp` final, as there is no other meaningful value, and the warnings can be avoided by setting the design flow rate. (And we should ask Dassault to evaluate such bindings, which bugs us in a variety of other models too.) If you remove the `final` then please put it into the `Advanced` tab, and I suggest to add an assertion to make sure the value is consistent with the medium properties (within a few percentage).

On that note, I also noticed that not all parameters on tabs other than the General tab have default values. I think this is the first model that requires users to set values on other tabs than on the General tab. Our instructions are that all parameters that need inspection are on the `General` tab, but more detailed settings can be found on the other tabs. I therefore suggest to change those parameters that require a user input from the `tab="Evaporator"` to `group="Evaporator"` and the same for the `Condenser`.

Then please go ahead and merge this PR so we have one code base again.

FWuellhorst:
Ok, I agree. I changed the tab as required and will merge this now

Issue #1831, opened 2024-02-01
Add HeatMeter as sensor


JAG-1337:
When validating systems of components or heating networks, I would like to have a simple heat meter for measurement of thermal energy.

The implementation inherits from `PartialDynamicFlowSensor` and adds a temperature input `T_other` to finally calculate

    Q_flow = port_a.m_flow * Medium.cp_const * (T - T_other);

with `T` being the temperature of the fluid passing through that sensor.

mwetter:
That would be a good addition. Can you please make an implementation and pull request to the branch `issue1831_heatMeter`.

PR #1832, opened 2024-02-01
Add HeatMeter as sensor

JAG-1337:
Adds a HeatMeter as sensor corresponding to issue #1831 .


mwetter:
@JAG-1337  : Thanks for the PR. I will further review.

PR #1843, opened 2024-02-27
Corrected wrong annotation

### LINKED ISSUE #1842: MediumColumnDynamic uses non-existent parameter in annotation
The model `Airflow.Multizone.MediumColumnDynamic` uses the annotation
```
visible=densitySelection == IBPSA.Airflow.Multizone.Types.densitySelection.actual
```
but this model has no parameter called `densitySelection`.

This issue is to remove this and other wrong annotations in this model.

mwetter:
This closes #1842


jelgerjansen (changes requested):
Hi Michael

Three minor remarks regarding the icon graphics:
1. I noticed you now use a dark blue rectangle in the icon, while `MediumColumn` has a light blue background (that partly changes to red, depending on the choice of `densitySelection`. Maybe we can change line 120 into `fillColor={85,170,255}`? Or did you pick another color on purpose?
2. The texstring "Top" has a larger font than "Bottom" because the font size is not specifically set. This is also the case for `Airflow.Multizone.MediumColumn`. Explicitly setting the fontsize to 36 gives both strings a font size that corresponds to the current font size of Bottom.
3. The textstring "name" is not centered in the middle. Again, this is also the case for `Airflow.Multizone.MediumColumn`.

These are very small (and fussy) comments, so feel free to ignore them ;)

mwetter:
@jelgerjansen : Thanks for the feedback. I addressed them in https://github.com/ibpsa/modelica-ibpsa/pull/1843/commits/995bedbb456dda866ae694ea8d756b115790032d

Can you please review and approve if all is fine so it will merged.

jelgerjansen (approved):
@mwetter while looking over the changes I noticed another minor icon inaccuracy in `MediumColumn`, which I fixed in 5ffc3085. All other changes look fine by me

PR #1877, opened 2024-05-13
Short cycling in air-to-water HP example

### LINKED ISSUE #1876: Short-cycling in heat pump example
@FWuellhorst 
In the model `Fluid.HeatPumps.ModularReversible.Examples.AirToWater2D_OneRoomRadiator` the heat pump short-cycles during part of the simulation. It switches on and off in a 1 minute interval during cooling mode.
It looks like the controller needs to be changed to avoid such operation.
This is with the model from https://github.com/ibpsa/modelica-ibpsa/pull/1628 but I suggest to handle this in a new PR so that we can merge https://github.com/ibpsa/modelica-ibpsa/pull/1628 once it passes the CI tests.

FWuellhorst:
Closes #1876.

I also changed the grouping of the safety control to match the enable clause and added my name to the contributors list, is that ok?


IBPSA/Fluid/HeatPumps/ModularReversible/Examples/AirToWater2D_OneRoomRadiator.mo
```diff
@@ -87,6 +87,16 @@ equation
...
+<p>
+  Furthermore, this example is useful to understand the different sizing options.
+  The selected value of <code>QCoo_flow_nominal</code> leads to
+  conflicting scaling factors for heating and cooling operation. This triggers a warning.
+  Using the default <code>QCoo_flow_nominal</code>, the warning disappears. 
+  Increasing (or decreasing) it, the number of on-off switches will increase (or decrease)
+  in cooling operation due to the low inertia of the system. 
+  In real air-to-water systems, thermal energy storage would be used to 
+  decrease cycling if the device is oversized.
+</p>
```
mwetter, line 99: I think it would be better to set up the model to avoid the warning
```
Warning: The following was detected at time: 0
  In AirToWater2D_OneRoomRadiator.heaPum: Scaling factors for heating and cooling
    operation differ by 41.2778 %. The simulated nominal heating and cooling
    capacities may not be realistic for a single device
  Failed condition: abs(heaPum.scaFacHea-heaPum.scaFacCoo)/heaPum.scaFacHea < heaPum.limWarSca
```
as from a user point of view, I would be concerned about seeing examples that don't seem to be configured in a realistic way.
Can you configure the model to avoid this warning.

If there is indeed a large discrepancy between loads, adding a small buffer tank in the return would help.

mwetter (changes requested):
Please see inline comment.
And thanks for adding your name to the contributor list, it certainly should be there.

IBPSA/Fluid/HeatPumps/ModularReversible/Examples/AirToWater2D_OneRoomRadiator.mo
FWuellhorst: Ok, I removed the warning by using the default and enabled the minimal runtime which was previously disabled. 
image
Now, each operation lasts at least 300 s.

FWuellhorst:
@mwetter I also added mover data to all examples to avoid the warnings raised by the movers. 
@hcasperfu : I also tried using the preconfigured models as suggested in the warning, but get the warning
image
which will lead to CI failure. Dynamics are set to SteadyState, FixedInitial does not change this. Do you have a guess why this happens?

mwetter (commented):
See inline comment about `perHeaLos` being what I think is unreasonably large.

IBPSA/Fluid/HeatPumps/ModularReversible/Examples/CarnotWithLosses_OneRoomRadiator.mo
```diff
@@ -5,7 +5,11 @@ model CarnotWithLosses_OneRoomRadiator
...
     sin(nPorts=1),
-    booToReaPumEva(realTrue=heaPum.mEva_flow_nominal));
+    booToReaPumEva(realTrue=heaPum.mEva_flow_nominal),
+    pumHeaPum(redeclare IBPSA.Fluid.Movers.Data.Pumps.Wilo.Stratos25slash1to6
+        per),
+    pumHeaPumSou(redeclare
+        IBPSA.Fluid.Movers.Data.Pumps.Wilo.Stratos25slash1to6 per));
   extends Modelica.Icons.Example;
 
   parameter Real perHeaLos=0.1
```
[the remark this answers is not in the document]
FWuellhorst: Agree, I think 1-2 % would be more suitable.

hcasperfu:
@FWuellhorst - The warning should be suppressed by 86f73be. 
To clarify, this warning did not cause CI to fail. But since we have been working on managing the warning messages produced by mover models (#1878), this report is welcome.

FWuellhorst:
@hcasperfu : Ok, I opted to use the preconfigured models in the examples in this case and changed the dp_nominal to the fitting values of the heat pump, which was previously not the case.
For the LargeScaleWaterToWater example, I get the following error in OpenModelica:
```
[1] 08:33:44 Übersetzung Fehler
[IBPSA.Fluid.Movers.BaseClasses.Characteristics.efficiencyParameters_yMot: 5:3-6:45]: [Failed to evaluate function: Modelica.Blocks.Types.ExternalCombiTable2D.constructor.](omeditmessagesbrowser:///IBPSA.Fluid.Movers.BaseClasses.Characteristics.efficiencyParameters_yMot?lineNumber=5)

[2] 08:33:44 Übersetzung Fehler
[C:/OM122/OM64bit/OMCompiler/Compiler/NFFrontEnd/NFCeval.mo: 1494:9-1494:67]: Internal error NFCeval.evalLogicBinaryAnd failed to evaluate ‘true and sum(pumHeaPum.per.pressure.dp) > Modelica.Constants.eps‘
```
When I change the `dp_nominal` in the movers to a fixed value, e.g. 150000 (roughly the output of `heaPum.dpEva_nominal`), the model simulates. My guess is that the combination of the preconfigured models and the non-literal scaling factors lead to this error. 

I will keep the 150000 for now and check after your mover updates if the error persists. If so, I will raise it to OM.

FWuellhorst:
@mwetter : I changed the movers and also the LargeScaleWaterToWater example as the cooling operation was previsously not triggered. To avoid cycling in this example as well, I adjusted the sizing and updated the documentation.

mwetter (approved):

Issue #1885, opened 2024-05-17
Borefield models allow inconsistent declaration for steady-state calculation of grout


mwetter:
The borefield models use data records for the materials and these data records have a parameter `steadyState`:
```modelica
  parameter Boolean steadyState = (cFil < Modelica.Constants.eps or dFil < Modelica.Constants.eps)
    "Flag, if true, then material is computed using steady-state heat conduction"
    annotation(Evaluate=true);
```
The models also have a parameter
```modelica
  // Dynamics of filling material
  parameter Boolean dynFil=true
    "Set to false to remove the dynamics of the filling material."
    annotation (Dialog(tab="Dynamics"));
```
Thus, one can set `steadyState` to any value and this won't affect `dynFil`, leading to potential inconsistent declarations.

This issue is to assign
```modelica
  final parameter Boolean dynFil=not borFieDat.filDat.steadyState
      "Set to false to remove the dynamics of the filling material.";
```
and remove the parameter `dynFil`.

PR #1886, opened 2024-05-17
Removed parameter dynFil in borefield models

mwetter:
This closes 1885


mwetter:
@jelgerjansen @FWuellhorst @nytschgeusen  : This is a backward incompatible change that avoids confusion due to a parameter having no effect on the simulation. See notes at #1885

Please raise any objections in the next 2 weeks as this is backward incompatible due to the removal of a parameter.

FWuellhorst:
Fine for AixLib!

jelgerjansen:
Fine for IDEAS as well!

FWuellhorst:
@mwetter : Why didn't you create the conversion script and increased the version number of IBPSA?

mwetter:
@FWuellhorst : good point. I will make a PR for the update to the conversion script. We increase the number only after an official release and version 4 has not yet been released (https://github.com/ibpsa/modelica-ibpsa/releases) so I left the version number.

Issue #1893, opened 2024-06-05
FlatPlateWithTank use code that is not in the IBPSA library


mwetter:
The instantiation of
https://github.com/ibpsa/modelica-ibpsa/blob/d78f71f340c81124590c561f58412c9624b5ad63/IBPSA/Fluid/SolarCollectors/Examples/FlatPlateWithTank.mo#L115

is incorrect.

Also, this model has an invalid annotation that is caught by Dymola 2024x Refresh 1.

mwetter:
Closed through https://github.com/ibpsa/modelica-ibpsa/pull/1894

PR #1894, opened 2024-06-05
Corrected instantiation of non-existent model

mwetter:
For #1893


jelgerjansen (changes requested):
@mwetter good catch, I missed that one when reviewing the solar collector merge into IBPSA.
I noticed that you removed the annotation of the block, so it's not visible in the GUI. I presume this was done by accident?
I added the annotation again. If you agree, this can be merged.

mwetter:
Thanks, the annotation was indeed removed by mistake.

PR #1909, opened 2024-07-15
Fix hysteresis values 

### LINKED ISSUE #1908: Fix tolerance in modular HP examples
This issue is linked to https://github.com/lbl-srg/modelica-buildings/issues/3930 and https://github.com/lbl-srg/modelica-buildings/issues/3943, see comments there for further explanations. Setting the tolerance to `0.001` should fix these issues

FWuellhorst:
Closes #1908 

@mwetter Can you or someone at lbl compare and generate the regression results for me? The compare option does not work for me in Docker.


mwetter (approved):

FWuellhorst:
@mwetter thanks for the heat pump results, can you re-create the chillers ones as well?

PR #1920, opened 2024-08-19
Changed markup commands in heat pump base class

### LINKED ISSUE #1919: For PartialReversibleRefrigerantMachine, change how excluded code is marked up
The code mark-up to exclude certain Modelica code for AixLib is not working.
This issue is to change the markup identifiers to allow merging it with the refactored version of BuildingsPy.
See the BuildingsPy issue https://github.com/lbl-srg/BuildingsPy/issues/566

mwetter:
This is to allow use of the refactored BuildingsPy, see https://github.com/lbl-srg/BuildingsPy/issues/566
that properly adds/removes code for AixLib. Other libraries worked already, and won't change the behavior due to this PR.

This closes #1919


FWuellhorst (approved):
Thanks for the quick fix!

PR #1936, opened 2024-10-04
Issue1935_unprotect dh and change prefixes to input

### LINKED ISSUE #1935: Unprotect dh in IBPSA.Airflow.Multizone.BaseClasses.DoorDiscretized
For non-vertical discretised large openings, the baseclass `IBPSA.Airflow.Multizone.BaseClasses.DoorDiscretized` can not be used because it requires overriding the protected parameter dh which is not good Modelica. However, with a small change the baseclass can be used for a broader purpose which matches the needs in IDEAS.

It can be solved by the prefix of `dh`, `hAg` and `hBg` as `input` in stead of `parameter`

kldjonge:
unprotect dh and change prefixes to input to hide it from the dialog box

This closes #1935


mwetter (changes requested):
Please add revision notes, otherwise it looks good provided the tests for https://github.com/lbl-srg/modelica-buildings/pull/4021 run. If they fail we will need to see why.

mwetter:
@kldjonge : can you please add the missing revision notes so we can complete this PR.

mwetter (approved):

Issue #1937, opened 2024-10-11
Refactor CheckValve using quintic Hermite spline interpolation


AntoineGautier:
We've experienced some NL solver convergence failures that point to the check valve model:
https://github.com/lbl-srg/modelica-buildings/issues/4002 
This ticket is to revise the implementation using a similar approach as the pressure-independent actuator models, based on a quintic Hermite spline interpolation (_C2_) between the closed valve flow function and the fully open valve flow function.
Initial tests with this new formulation show that all convergence issues are eliminated on the models from the above thread (with minor but favorable impact on simulation time).

PR #1938, opened 2024-10-14
Refactor check valve with spline interpolation

AntoineGautier:
This addresses #1937.


AntoineGautier:
@mwetter This is ready. Can you suggest a name for an external review?

AntoineGautier:
@mwetter This is ready. Can you suggest a name for an external review?

IBPSA/Fluid/FixedResistances/CheckValve.mo
```diff
@@ -27,48 +27,75 @@ model CheckValve "Check valve that avoids flow reversal"
...
+      x=dp,
+      x1=0,
+      x2=dpValve_closing,
+      y1=m_flow_min,
+      y2=m_flow_max,
+      y1d=IBPSA.Fluid.BaseClasses.FlowModels.basicFlowFunction_dp_der(
+        dp=0,
+        k=k_min,
+        m_flow_turbulent=m_flow_turbulent,
+        dp_der=1),
```
mwetter, line 75: Can you explain why `dp_der=1` and `dp_der2=0`. I don't understand what you implemented here.
(A side note is that these are parameter expressions so for efficiency that could be computed as a parameter.)

mwetter (commented):
@AntoineGautier : I am not sure we approached this the right way, in part because I don't understand your implementation (see inline comment).

Wouldn't it be simpler if we were to compute $dp$ from $\dot m$? This would make it simpler because both components have the same $\dot m$, so we can simply add their resistances. For the fixed resistance the evaluation is easy (with the flow function). For the actual valve, we can make a linear behavior for $\dot m \le 0$ and quadratic for $\dot m \ge \dot m_0/2$. In between we use a polynominal that fits the first and second derivatives at these points. The coefficients of the polynomial can be computed as parameter expressions.

IBPSA/Fluid/BaseClasses/FlowModels/basicFlowFunction_dp_der.mo
```diff
@@ -31,7 +31,7 @@ Documentation(info="<html>
 Function that implements the first order derivative of
 <a href=\"modelica://IBPSA.Fluid.BaseClasses.FlowModels.basicFlowFunction_dp\">
 IBPSA.Fluid.BaseClasses.FlowModels.basicFlowFunction_dp</a>
-with respect to the mass flow rate.
+with respect to time.
```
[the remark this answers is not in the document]
AntoineGautier: Considering that
- the algorithm of the function `basicFlowFunction_dp_der` can be written as: 
$`m\_flow\_der = \frac{\partial basicFlowFunction\_dp}{\partial dp} * \frac{\partial dp}{\partial t}`$ (where $`\frac{\partial dp}{\partial t} = dp\_der`$)
- its output is declared with kg/s2 unit: `m_flow_der(unit="kg/s2")`
- the function is validated in `Fluid.BaseClasses.FlowModels.Validation.BasicFlowFunction_dp_DerivativeCheck` by comparing the time integral of the derivative `der(m_flow_comp)` with the original function `basicFlowFunction_dp`

I conclude that `basicFlowFunction_dp_der` is the derivative of `basicFlowFunction_dp` with respect to *time* (not `dp`).

Then, considering 
$`\frac{\partial m\_flow}{\partial t} = \frac{\partial m\_flow}{\partial dp} * \frac{\partial dp}{\partial t}`$
I compute $`\frac{\partial m\_flow}{\partial dp}`$ by calling `basicFlowFunction_dp_der(..., dp_der=1)`.

IBPSA/Fluid/FixedResistances/CheckValve.mo
AntoineGautier: See above.
For the second derivative, considering
$`\frac{\partial^2 m\_flow}{\partial t^2} = \frac{\partial^2 m\_flow}{\partial dp^2} * \left(\frac{\partial dp}{\partial t}\right)^2 + \frac{\partial m\_flow}{\partial dp} * \frac{\partial^2 dp}{\partial t^2}`$
I compute $`\frac{\partial^2 m\_flow}{\partial dp^2}`$ by calling `basicFlowFunction_dp_der2(..., dp_der=1, dp_der2=0)`.

AntoineGautier:
@mwetter With [11097a8](https://github.com/ibpsa/modelica-ibpsa/pull/1938/commits/11097a81bf2b34fb2c77ec733fd70d8ae9dd70be) the function calls are now included in the `if else` clauses, `min` and `max` are removed and parameters are introduced when possible.
I've also updated the documentation for the derivatives of the basic flow functions as discussed. Note the important information that these derivatives are formulated under the assumption of a constant flow coefficient. This is consistent with the annotation `derivative(zeroDerivative=k)` in `basicFlowFunction_dp`. This also means that if a model features a varying flow coefficient (e.g., any actuator model), these derivative functions will not be used by the tools, which will rely on automatic differentiation instead.

mwetter (approved):
@AntoineGautier : Thanks for the changes. This addresses my concerns that we discussed on Friday.

PR #1941, opened 2024-11-02
Changed guard against division by zero in electrical resistance

mwetter:
This is to port the improvement for Buildings, #4032


mwetter:
@LauraMaier : Can you please review this change as it causes issues in a downstream application.

LauraMaier (approved):

PR #1943, opened 2024-11-14
Changed default initialization of electrical load

### LINKED ISSUE #1942: Model Electrical.DC.Loads.Examples.LinearizedLoad fails in OM due to a wrong initialization
This issue is to correct the model to avoid the division by zero reported in https://github.com/lbl-srg/modelica-buildings/issues/4048

mwetter:
This closes #1942


mwetter:
@LauraMaier : Can you please review this PR which is pending since quite a while already. If these pile up, it just causes more work for downstream developments. Thanks.

FWuellhorst (approved):

Issue #1958, opened 2025-01-29
Include SDF library


AntoineGautier:
This ticket is to include https://github.com/ScientificDataFormat/SDF-Modelica into the IBPSA library, e.g., within `IBPSA.Utilities.IO.SDF`.

## Use Case

The multi-dimensional interpolation is useful for modeling chiller and heat pump performance, which usually depends on more than two variables: evaporator and condenser temperature, part load ratio or compressor speed. 
Especially, the models in `IBPSA.Fluid.HeatPumps.ModularReversible` and `IBPSA.Fluid.Chillers.ModularReversible` can be readily updated to use SDF classes for calculating efficiency and capacity.

## Licensing

The SDF library is released under the [2-Clause BSD license](https://github.com/ScientificDataFormat/SDF-Modelica/blob/main/LICENSE.txt).
It requires a copyright notice to Dassault Systèmes.

AntoineGautier:
This issue is closed without being addressed based on the discussions from the corresponding PR page https://github.com/ibpsa/modelica-ibpsa/pull/1959

PR #1959, opened 2025-01-29
Include SDF

AntoineGautier:
This is for #1958.

TODO:
- [ ] Add copyright notice
- [ ] Add some auto-generated unit tests for models in `IBPSA.Utilities.IO.SDF.Examples`
- [x] Test example models with OMC: ✅ tests pass
- [x] Investigate why example models fail to simulate with OCT
  - [Modelon - 3172] Support Ticket: Support for arrays that can be initialized at runtime
     Response from Modelon:
    > It is currently not allowed to have the size of a parameter depend on a "impure" function. The size needs to be computed by pure functions for the model to work with OCT.
The model might work if "impure" is removed from the functions involved.
  - Testing various refactoring of the Modelica functions to avoid invoking impure functions for assigning array sizes, the initial translation error mutates into either a C-compilation error or a Java null pointer exception. I conclude from these first tests that OCT support is not granted. This is somewhat surprising as the SDF library was initially developed by Modelon AB (see https://github.com/modelica/ModelicaStandardLibrary/issues/1516#issue-200837333).


AntoineGautier:
@mwetter @FWuellhorst This draft PR shows how the SDF library could be integrated into IBPSA. The integration is done with the script https://github.com/ibpsa/modelica-ibpsa/blob/issue1958_sdf/bin/importSDF.py

Please let me know what you think in terms of
- package location
- licensing & copyright aspects
- unit tests (currently the models in `IBPSA.Utilities.IO.SDF.Examples` are not included in CI testing)
- opportunity to update the modular HP models with 3D-table data

FWuellhorst:
@AntoineGautier @mwetter : Nice work, glad to see SDF as an option. 

- If it is allowed license-wise (I am not a legal expert), I think the location is a good fit. 
- If we have examples using SDF in the heat pump models, testing would not necessarily be required. If SDF itself does not have testing included, and we start to use it in other models as well, some basic tests could be in order. 
- Regarding HP models, I already included options in AixLib, and we could just merge those into IBPSA. I just extended the SDF based approach a little to avoid code redundancy when using 4 or 5 dimensions and am really close to finish an experimental validation which uses 5D performance data (including compressor speed, condenser mass flow rate and degree superheat). So once SDF is supported in IBPSA, I could open that PR. Based on your needs, we could adjust the format.

mwetter:
@AntoineGautier @FWuellhorst 
- The package location looks good to me. I would however put the importSDF.py to Resources/src/Utilities/IO/SDF (this is what we also use in Buildings for other code that imports 3rd party code). The script is also picky regarding the location from which it is invoked.
- The licenses are BSD, so if we reproduce them I don't see any issue
- Adding test would be good. For example, `IBPSA.Utilities.IO.SDF.Examples.Playback` fails (the other two examples work)
- We should also put a sentence in IBPSA.Utilities.IO.SDF explaining where it comes from.
- Including this feature in the HP models would be good (and is in my view the main motivation to add SDF).

AntoineGautier:
@mwetter @FWuellhorst After doing some tests, I would rather close this PR without merging for the following reasons.

### Tool Support

The first issue is the current lack of support by OCT, see https://github.com/ibpsa/modelica-ibpsa/pull/1959#issue-2818303282. 

### Compliance Issues

There also seems to be some compliance issues with the current implementation of SDF.

The external function `evaluate` in `IBPSA.Utilities.IO.SDF.NDTable` is defined without `pure` or `impure` which violates the MLS (https://specification.modelica.org/maint/3.6/functions.html#pure-modelica-functions).

Fundamentally this is an impure function, and its use in the model `IBPSA.Utilities.IO.SDF.NDTable` does not respect the requirements for impure function calls specified at https://specification.modelica.org/maint/3.6/functions.html#pure-modelica-functions.

### Good Alternative

The main motivation to use SDF was to add an additional interpolation variable for the chiller & HP capacity and power calculation.
But this can be simply achieved by reading multiple 2D-tables (corresponding to multiple PLR values) from the same external file with `Modelica.Blocks.Types.ExternalCombiTable2D[:]` and locally interpolating along the PLR values with `Modelica.Math.Vectors.interpolate`.

The external ASCII file can have the following syntax, where `q@(PLR_VALUE)` and `p@(PLR_VALUE)` provide the heat flow rate and input power at a given PLR for the condenser and evaporator entering or leaving temperatures.

```
#1
double q@0.10(5, 5)
0 297.04 306.3 315.56 324.82
277.59 -9.67E+03 -9.94E+03 -8.54E+03 -5.47E+03
279.44 -1.06E+04 -1.06E+04 -9.03E+03 -5.77E+03
281.3 -1.13E+04 -1.12E+04 -9.36E+03 -5.91E+03
283.15 -1.18E+04 -1.15E+04 -9.53E+03 -5.88E+03
...
double q@1.00(5, 5)
0 297.04 306.3 315.56 324.82
277.59 -9.67E+04 -9.94E+04 -8.54E+04 -5.47E+04
279.44 -1.06E+05 -1.06E+05 -9.03E+04 -5.77E+04
281.3 -1.13E+05 -1.12E+05 -9.36E+04 -5.91E+04
283.15 -1.18E+05 -1.15E+05 -9.53E+04 -5.88E+04
double p@0.10(5, 5)
0 297.04 306.3 315.56 324.82
277.59 2.36E+03 3.25E+03 3.63E+03 2.95E+03
279.44 2.52E+03 3.35E+03 3.67E+03 2.96E+03
281.3 2.62E+03 3.37E+03 3.61E+03 2.86E+03
283.15 2.67E+03 3.31E+03 3.47E+03 2.68E+03
...
double p@1.00(5, 5)
0 297.04 306.3 315.56 324.82
277.59 2.36E+04 3.25E+04 3.63E+04 2.95E+04
279.44 2.52E+04 3.35E+04 3.67E+04 2.96E+04
281.3 2.62E+04 3.37E+04 3.61E+04 2.86E+04
283.15 2.67E+04 3.31E+04 3.47E+04 2.68E+04
```

When developing models for energy analysis, interpolating along the PLR values locally has an additional benefit as it gives immediate access to the inverse model of the performance calculation algorithm. 
So an ideal controller can simply be implemented by 
- computing the heating or cooling load, 
- interpolating along the heat flow rate values to calculate the PLR value that enables meeting the load,
- (optionally modifying this PLR value to account for internal safeties),
- using this PLR value to calculate the input power.

In contrast, when relying on external interpolation (as with SDF), accessing the inverse of the performance calculation algorithm becomes an iterative solving problem (e.g., using `Modelica.Blocks.Math.InverseBlockConstraints`). I've tested this approach and formulating the problem so that it always converges turns out tricky.

### Benefit for Users

Using external ASCII files as suggested above eliminates the need for users to install and manipulate heavy dependencies such as https://pypi.org/project/SDF/.

### Library Size

Including SDF increases the IBPSA Library's on-disk footprint by 23 %.

mwetter:
@AntoineGautier : I agree with your suggested path of using `Modelica.Blocks.Types.ExternalCombiTable2D[:]` and locally interpolating along the PLR values with `Modelica.Math.Vectors.interpolate` in view of better numerical performance, SDF not complying to the standard, and lack of OCT support. (Even if OCT were to support it, if it violates the standard, there is a real risk that it stops working if a Modelica tool is updated.)

FWuellhorst:
@AntoineGautier Thanks for this thorough check! While I agree with the points above, I have the following concerns:
- More dimensions: Is your approach also feasible when going to 4D or even more dimensions? 
- Size of data: While IBPSA increases in size when including sdf, storing the data as ASCII will increase the size of the data-tables. I have an 4D table wich is 800 kB in sdf and 2.3 MB in .csv format.
- Python-Support: While sdf is a dependency, one would have to write a converter to enable converting e.g. pandas DataFrame tables to the required file format.
- Extrapolation: Is this as easy with the `Modelica.Math.Vectors.interpolate`? It seems to linearly extrapolate, which is unwanted for most efficiency maps.
- IBPSA or MSL? If both easy ND inter- and extrapolation are possible with some sort of easy way to automate the creating of the ASCII files, wouldn't this be a feature for MSL instead of IBPSA?

AntoineGautier:
@FWuellhorst Thanks for elaborating on this. Here's my take on the different aspects you pointed out.

### More dimensions

Yes, you can extend the approach to higher dimensionalities. 
Let's take 4D, with 2D handled externally with `Modelica.Blocks.Types.ExternalCombiTable2D` and 2D handled locally with `Modelica.Math.Vectors.interpolate`.
The data file now contains tables tagged with `q@v1@v2`.
You have a 2D instance of `Modelica.Blocks.Types.ExternalCombiTable2D tabQ[n1, n2]`.
And you sequentially interpolate along v1 and v2:
```mo
  QInt = Modelica.Blocks.Tables.Internal.getTable2DValueNoDer2(
    tabQ, fill(TLoa, n1, n2), fill(TSou, n1, n2));
  for i in 1:n2 loop
    QV2_flow[i] = Modelica.Math.Vectors.interpolate(v1Sup, QInt[:, i], v1);
  end for;
  Q_flow = Modelica.Math.Vectors.interpolate(v2Sup, QV2_flow, v2);
```

### Size of data

I agree that SDF is better suited for handling large data files, as it is based on HDF5. But I was going to use it as an additional argument to *not* use SDF. In my experience, it is not straightforward to obtain manufacturer data. The best I could ever get is 5 PLR values at 3 temperature levels, so 45 capacity and power values. I don't think that catalog data give you much more than that either. So SDF seemed rather like an overkill to me.
I would consider table data with several thousands of capacity and power values to be more outliers than usual. And as such, should they be included in the library?

### Python support

Exporting a data frame into the specific format is relatively straightforward, and only requires a few for loops.

### Extrapolation

Yes, `Modelica.Math.Vectors.interpolate` does extrapolate through the first or last two x[:] values. But this is easily reverted by limiting either the x-value or the resulting y-value.

### IBPSA or MSL? 

I would not claim for generalization, as there is still this arbitrary separation of interpolation domains between  `Modelica.Blocks.Types.ExternalCombiTable2D` and `Modelica.Math.Vectors.interpolate`.

Issue #1956, opened 2025-01-29
Wrong calculation of nominal mass flow rate in PartialSolarCollector


louisher:
In `IBPSA.Fluid.SolarCollectors.BaseClasses.PartialSolarCollector`, there is an issue with the calculation of the nominal mass flow rate when the number of panels is set by using the `totalArea` parameter, instead of the `nPanels` parameter. The parameter `nPanelsPar_internal` used to calculate `m_flow_nominal_final` is defined as: 

```
 parameter Real nPanelsPar_internal=
    if sysConfig == IBPSA.Fluid.SolarCollectors.Types.SystemConfiguration.Parallel then 
      nPanels
    else if sysConfig == IBPSA.Fluid.SolarCollectors.Types.SystemConfiguration.Series then 
      1
    else 
      nPanelsPar "Number of panels in parallel";
```

When `nPanels` is not set by the user and `sysConfig == IBPSA.Fluid.SolarCollectors.Types.SystemConfiguration.Parallel`, the nominal mass flow rate remains zero. To fix this, `nPanelsPar_internal` should be used instead of `nPanels`. The same should be done for `nPanelsSer_internal`.

jelgerjansen:
@mwetter I already implemented the proposed change and created a pull request (PR #1957). Could you review this (or assign another (non-IDEAS) reviewer)?

PR #1960, opened 2025-01-29
Wrong calculation of nominal mass flow rate in PartialSolarCollector

mwetter:
This closes issue https://github.com/ibpsa/modelica-ibpsa/issues/1956

### DIFF
--- a/IBPSA/Fluid/SolarCollectors/BaseClasses/PartialSolarCollector.mo
+++ b/IBPSA/Fluid/SolarCollectors/BaseClasses/PartialSolarCollector.mo
@@ -158,14 +158,14 @@ protected
       totalArea/per.A "Number of panels used in the simulation";
   parameter Real nPanelsSer_internal=
     if sysConfig == IBPSA.Fluid.SolarCollectors.Types.SystemConfiguration.Series then
-      nPanels
+      nPanels_internal
     else if sysConfig == IBPSA.Fluid.SolarCollectors.Types.SystemConfiguration.Parallel then
       1
     else
       nPanelsSer "Number of panels in series";
   parameter Real nPanelsPar_internal=
     if sysConfig == IBPSA.Fluid.SolarCollectors.Types.SystemConfiguration.Parallel then
-      nPanels
+      nPanels_internal
     else if sysConfig == IBPSA.Fluid.SolarCollectors.Types.SystemConfiguration.Series then
       1
     else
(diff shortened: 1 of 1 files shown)


mwetter:
@jelgerjansen : There is still something wrong with the parameter calculation. Can you please run
```
simulateModel(
  IBPSA.Fluid.SolarCollectors.Examples.FlatPlate(solCol(nColType=IBPSA.Fluid.SolarCollectors.Types.NumberSelection.Area, totalArea=10, nPanelsPar=4))",
  stopTime=86400,
  tolerance=1e-06,
  resultFile="FlatPlate");
```
which triggers the assertion for which I added more output. This was already wrong before your change, but should be fixed too.

jelgerjansen:
@mwetter the current implementation requires the user to input both the total number of collector panels (either via `nPanels` or `totalArea`) and the number of collector panels in series and parallel (`nPanelsSer` and `nPanelsPar`). This is an "overspecified problem". 

As a solution, I kept `nPanelsSer` as a user-defined parameter in case of an array and calculated the other parameter `nPanelsPar_internal = nPanels_internal/nPanelsSer_internal`. This also avoids the need for an assert to check whether `nPanels=nPanelsSer*nPanelsPar`. 

However, this implementation is not yet ideal, as `totalArea` is not necessarily an exact multiple of the area of a single collector. The user might have an array of collectors which is 5 times the area of a single collector, and model this as 2 parallel branches of 2.5 collectors in series. This is not possible in the current implementation as `nPanelsSer` is defined as an integer. Therefore, I suggest defining `nPanelsSer` as a `Real` parameter.

mwetter:
@jelgerjansen : I agree that the problem is over-specified, that we should remove one parameter.

I however suggest to remove `nPanelsSer`, and keep instead `nPanelsPar`, and make `nPanelsPar` an `Integer`. This way, we have always an integer value for the number of parallel circuits, which seems more logical from a construction point of view. We can then keep `nPanelsSer_internal` as a `Real`, and if it ends up to not be equal to an integer within say 1%, issue a assertion with a level `warning`. 

With this approach, if a user specifies to use `totalArea`, we have an integer number for `nPanelsPar`, and perhaps a fraction for `nPanelSer`, in which case we warn the user and proceed with the simulation.

Does this sound good?

jelgerjansen:
@mwetter I agree and implemented the proposed changes. Feel free to update the models if something is still missing or if you want to rephrase the warning message.

jelgerjansen:
@mwetter I assigned @FWuellhorst to do a review (as he also faced this issue recently). Can we merge this PR after his review? Or do you still want to do a final review yourself?

FWuellhorst (approved):
Looks good!

mwetter:
This is ready to merge if @FWuellhorst and @jelgerjansen agree.

FWuellhorst (approved):

jelgerjansen (approved):
Thanks for the final check @mwetter! The additional changes look good to me!

PR #1967, opened 2025-02-10
Ported changes from Buildings for CDL.Logica.TrueFalseHold

### LINKED ISSUE #1966: Integrate changes to TrueFalseHold
This issue is to implement the changes from 
https://github.com/lbl-srg/modelica-buildings/pull/4094
to the `OBC.CDL.Logical.TrueFalseHold` block.

mwetter:
This closes #1966


FWuellhorst (approved):
Changes look good, but I am wondering if a documentation warning or similar should be raised if users use cvode with dymola < 2025x or other tools. Or is this irrelevant as the bug error is not present in cvode in OM or FMU-export?

mwetter:
This issue only happens if the block is used in certain models, and with Dymola 2024x and the CVode solver, so a blank warning is in my opinion difficult.

PR #1969, opened 2025-02-11
Removed assignment of parameter that no longer exits in SolarCollector example

### LINKED ISSUE #1968: Bug in solar collector example after refactoring
It turns out that the `SolarCollectors` package is not run in the CI tests, and it has a model whose parameter also need to be removed due to https://github.com/ibpsa/modelica-ibpsa/pull/1960

This issue is to correct it.

mwetter:
This closes #1968
It also adds the package to the CI tests.


jelgerjansen (approved):
@mwetter thank you for noticing and fixing this! I noticed an obsolete whitespace in the revision history which I removed myself. The changes now look good to me!

PR #1988, opened 2025-03-10
Add missing semi-colon in BESTEST validation models

### LINKED ISSUE #1987: Correct missing semi-colon in BESTEST validation models
A couple of models in `IBPSA.ThermalZones.ISO13790.Validation.BESTEST` miss a semi-colon after the `extends` statement, and this causes the following `annotation` to be interpreted as an annotation of the `extends` statement rather than of the model.

mwetter:
This closes #1987


amaccarini (approved):
Thanks for catching that, Michael!

Issue #1976, opened 2025-02-24
IBPSA.Fluid.Geothermal.Borefields.Data.Configuration error with 5 or 6 boreholes


xenkGit:
Dear IBPSA team,

it seems IBPSA.Fluid.Geothermal.Borefields.Data.Configuration does not work with 5 or 6 boreholes.
Works:
cooBor={{22.24,50.03},{17.21,46.06},{11.12,43.68},{4.76,43.68},{0,43.68},{0,37.59},{0,31.5}}
cooBor={{22.24,50.03},{17.21,46.06},{11.12,43.68},{4.76,43.68}}
Does not work:
cooBor={{22.24,50.03},{17.21,46.06},{11.12,43.68},{4.76,43.68},{0,43.68}}
cooBor={{22.24,50.03},{17.21,46.06},{11.12,43.68},{4.76,43.68},{0,43.68},{0,37.59}}

The issue is reproducible by adding one borehole to IBPSA.Fluid.Geothermal.Borefields.Data.Configuration.Example and then simulating IBPSA.Fluid.Geothermal.Borefields.Examples.Borefields.

Sorry if I'm overlooking something obvious...

<img width="944" alt="Image" src="https://github.com/user-attachments/assets/f375d5a0-c4fa-4a54-bb6f-c5f78d6c47ca" />

mwetter:
I can reproduce this error by overriding `cooBor` with
```
cooBor={{22.24,50.03},{17.21,46.06},{11.12,43.68},{4.76,43.68},{0,43.68}}
```
in `IBPSA.Fluid.Geothermal.Borefields.Data.Configuration.Example` and then simulate `IBPSA.Fluid.Geothermal.Borefields.Examples.Borefields`

I made the change on the branch `issue1976_borefield_coord`.

@MassimoCimmino : Can you please look into what causes this division by zero. The same error also happens with OpenModelica.

mwetter:
I can reproduce this error by overriding `cooBor` with
```
cooBor={{22.24,50.03},{17.21,46.06},{11.12,43.68},{4.76,43.68},{0,43.68}}
```
in `IBPSA.Fluid.Geothermal.Borefields.Data.Configuration.Example` and then simulate `IBPSA.Fluid.Geothermal.Borefields.Examples.Borefields`

I made the change on the branch `issue1976_borefield_coord`.

@MassimoCimmino : Can you please look into what causes this division by zero. The same error also happens with OpenModelica.

[...]

PR #1993, opened 2025-03-18
Add absolute tolerance to KMeans algorithm

### LINKED ISSUE #1985: Division by zero in Clustering KMeans function
When trying to simulate `IBPSA.Utilities.Clustering.Validation.KMeans_2d` using Wolfram System Modeler, we get a division by zero in this line in the first iteration of the loops:

https://github.com/ibpsa/modelica-ibpsa/blob/b43e77de032d3cc8d7f8a1b501ef8a91c72282f8/IBPSA/Utilities/Clustering/KMeans.mo#L87

One of the zeroes from `data` seems to make it into the `old_centroids`.

Is this expected? Are other tools just swallowing the division by zero silently, and is there support for that in the specification?

MassimoCimmino:
This closes #1985 and #1976.

I did not add any test since the issue occured in the already-implemented cases. Issue #1976 was caused by an error in the initialization of centroid positions.


mwetter:
@MassimoCimmino : Thanks for the changes. The verification fails due to a change in result of 0.4 K in one model, and in `IBPSA.Utilities.Clustering.Validation.KMeans_1d`, the returned clustering is different from the expected `labelsExp`. Can you please review and update as needed.

MassimoCimmino:
I updated `labelsExp` in `IBPSA.Utilities.Clustering.Validation.KMeans_1d`, `IBPSA.Utilities.Clustering.Validation.KMeans_2d` and `IBPSA.Fluid.Geothermal.Borefields.BaseClasses.HeatTransfer.ThermalResponseFactors.Validation.ClusterBoreholes_100boreholes`. The changes to `IBPSA.Utilities.Clustering.KMeans` enables the algorithm to find clusters with lower inertia. The difference in clusters in `IBPSA.Fluid.Geothermal.Borefields.BaseClasses.HeatTransfer.ThermalResponseFactors.Validation.ClusterBoreholes_100boreholes` would also explain the new results from the other models which use the same borefield geometry. I updated the results there also.

mwetter (approved):

PR #2000, opened 2025-04-17
Fix unit in frost calculation

### LINKED ISSUE #1975: Bug in heat pump frosting model
In `IBPSA.Fluid.HeatPumps.ModularReversible.RefrigerantCycle.Frosting.Functions.wetterAfjei1997`, the evaporator inlet is currently in K, but the function expects degC. This was missed when switching from degC to K during the move from `AixLib` to `IBPSA`.

This was partly also missed as the example containing the iceFac model `IBPSA.Fluid.HeatPumps.ModularReversible.Examples.CarnotWithLosses_OneRoomRadiator` has a constant source temperature of 8 °C but is an air-to-water heat pump.

Suggestions:
- [x] Convert to degC in the function itself
- [x] Update Example to use `TDryBul` from weather bus as in a real air-to-water heat pump. @mwetter , do you think its also better to update all air-to-water examples to use `TDryBul` directly?

FWuellhorst:
Closes #1975.

I kept the changes to only address #1975. 
With future features and new frosting approaches, a separate examples package would make sense. 
Also, the `CarnotWithLosses` always assumes frosting, even though the air side medium is not fixed. @mwetter should I change this in a new PR or in this one? 

Note: I changed the results in [IBPSA/Fluid/HeatPumps/ModularReversible/Examples/CarnotWithLosses_OneRoomRadiator.mo](https://github.com/ibpsa/modelica-ibpsa/pull/2000/files#diff-f4f7b23774549857562131169984f6d9555dd63ff7b5a7bfe072c8006b52bfd1) as the example now has frosting effect. Can anyone create the reference result? I currently have no docker env to do this.


mwetter:
@FWuellhorst : The example `IBPSA/Fluid/HeatPumps/ModularReversible/Examples/CarnotWithLosses_OneRoomRadiator.mo` became quite a Frankenstein model: It uses water at the evaporator, that starts at 0 degC, and then uses this to test that the heat pump models frost from the air water vapor.
I think we should leave the model as is. If we want to test frost formation, then let's make another example or validation case that we can explain physically.

I think it would be good to make a new PR for disabling frost formation if the medium is not air, and then we can proceed with merging this after reverting the example.

FWuellhorst:
@mwetter The model now uses air, as it was originally intended. See the change
```
redeclare package MediumEva = MediumAir,
```
I can revert this and keep the water medium if you like, but with this change I don't see the issue with the model. It is air to water carnot with losses, including frosting.

FWuellhorst:
@mwetter That was my initial suggestion in #1975, but you opted to keep TEva_in and don't use TDryBul. I can revert this change for know, but in the long run it makes sense to use the outdoor air temperature in case of air-source heat pump examples in my opinion.

mwetter:
@FWuellhorst : I think I now see the discrepancy: The HP in this example is computing the loss due to icing and defrost, **but** the model uses water as a the fluid in the evaporator as it was set up to mimic for example a geothermal coupled heat pump. So my question is why this model computes icing and defrost, which is not physical.  Could the instance be disabled for example if the medium density indicates that it is a liquid rather than air?

The frosting model is from 
```modelica
model CarnotWithLosses
  "Heat pump using the Carnot approach, but with added reversibility and losses (heat, frost, inertia)"
  extends IBPSA.Fluid.HeatPumps.ModularReversible.Modular(
    QCoo_flow_nominal=-PEle_nominal*refCyc.refCycHeaPumCoo.EER_nominal,
    redeclare model RefrigerantCycleHeatPumpCooling =
        IBPSA.Fluid.Chillers.ModularReversible.RefrigerantCycle.ConstantCarnotEffectiveness
        (
        PEle_nominal=PEle_nominal,
        redeclare 
          IBPSA.Fluid.HeatPumps.ModularReversible.RefrigerantCycle.Frosting.NoFrosting
          iceFacCal,
        etaCarnot_nominal=etaCarnot_nominal,
        use_constAppTem=true,
        TAppCon_nominal=TAppCon_nominal,
        TAppEva_nominal=TAppEva_nominal),
    redeclare model RefrigerantCycleHeatPumpHeating =
        IBPSA.Fluid.HeatPumps.ModularReversible.RefrigerantCycle.ConstantCarnotEffectiveness
        (
        redeclare 
          IBPSA.Fluid.HeatPumps.ModularReversible.RefrigerantCycle.Frosting.FunctionalIcingFactor
          iceFacCal(redeclare function icingFactor =
              IBPSA.Fluid.HeatPumps.ModularReversible.RefrigerantCycle.Frosting.Functions.wetterAfjei1997),
```
It is in a redeclared model that is in a redeclared model and which has a redeclared function. I don't think many users know what to do here (and as they are not replaceable, probably nothing can be done by the user). Should therefore in the icing base class the icing calculation be disabled if the medium density shows it is a fluid? Otherwise I think many users will get wrong results for water-source heat pumps.

FWuellhorst:
@mwetter I agree. I added an option to switch `iceFac=1` if `cpEva > 1500`. Trying to disable the whole block did not work as `cpEva` is not know during model check. One could add an explicit boolean, but I think the current approach is more user-friendly.

I reverted the example but added a Validation package to test if the function from your paper is correct (it is now) and if the disabling works as intended. 
If you agree with the changes, could you or someone else please generate the regression results for the new model? I currently don't have access to docker.

mwetter (approved):

PR #2002, opened 2025-04-23
add option to advanced tab. Change one example to indicate if the cha…

### LINKED ISSUE #2001: Enable computeFlowResistance=false
To simplify complex hydraulic models, it should be possible to disable the pressure loss equation in a model but still use the `dp_nominal` and aggregate the value with other `dp_nominals` in series to one `dpFixed_nominal` of an actuator. 
As discussed and proposed by @Mathadon , the parameter 
```
protected 
  final parameter Boolean computeFlowResistance=(dp_nominal_pos > Modelica.Constants.eps)
    "Flag to enable/disable computation of flow resistance"
   annotation(Evaluate=true);
```
in `IBPSA.Fluid.FixedResistances.PressureDrop` could be public and non-final in a tab advanced.

FWuellhorst:
Closes #2001 .

@mwetter @thorade Beside the change, I used the advanced option in one example to see if everything works as intended.
If it works as you suggested, we could think about changing other examples as well to show users how this parameter can be used.

### DIFF
--- a/IBPSA/Fluid/Examples/FlowSystem/Simplified2.mo
+++ b/IBPSA/Fluid/Examples/FlowSystem/Simplified2.mo
@@ -1,10 +1,11 @@
 within IBPSA.Fluid.Examples.FlowSystem;
 model Simplified2 "Using from_dp"
   extends Simplified1(
-    valSouth1(each from_dp=true),
+    valSouth1(each from_dp=true, dpFixed_nominal=tabsSouth1.dp_nominal),
     valSouth2(each from_dp=true),
     valNorth1(each from_dp=true),
-    valNorth2(each from_dp=true));
+    valNorth2(each from_dp=true),
+    tabsSouth1(each dp_nominal=50000, each computeFlowResistance=false));
   annotation (Documentation(info="<html>
 <p>
 The model is simplified: using <code>from_dp</code> to find more efficient tearing variables.
--- a/IBPSA/Fluid/FixedResistances/PressureDrop.mo
+++ b/IBPSA/Fluid/FixedResistances/PressureDrop.mo
@@ -9,14 +9,15 @@ model PressureDrop
        annotation(Evaluate=true,
                   Dialog(group = "Transition to laminar",
                          enable = not linearized));
+  parameter Boolean computeFlowResistance=(dp_nominal_pos > Modelica.Constants.eps)
+    "=false to disable computation of flow resistance and aggregate pressure drop in an actuator model"
+   annotation(Dialog(tab="Advanced"), Evaluate=true);
 
   final parameter Real k = if computeFlowResistance then
         m_flow_nominal_pos / sqrt(dp_nominal_pos) else 0
     "Flow coefficient, k=m_flow/sqrt(dp), with unit=(kg.m)^(1/2)";
+
 protected
-  final parameter Boolean computeFlowResistance=(dp_nominal_pos > Modelica.Constants.eps)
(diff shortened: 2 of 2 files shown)


mwetter:
These changes look OK, but I reverted the change to `IBPSA/Fluid/Examples/FlowSystem/Simplified2.mo` as this is used in a paper, and to build the various simplified models. We shouldn't change this model.

I am still not quite sure if the higher complexity is really useful as now one has to check if `dp_nominal` is not 0 **and** whether `computeFlowResistance` is unchanged when reviewing models. Why do you want to compute `dp_nominal` (which is not really computed in `PressureDrop`) and then set it to zero for the simulation to aggregate it somewhere else? Maybe that advanced setting should not be in this simple workhorse model?

FWuellhorst:
Thanks for the better example.

I see your point in the pressure drop model, but we often use this model in other components (e.g. heat pump, boiler etc.). Having the dp_nominal associated to the model and then using the nominal pressure drop inside `dpFixed_nominal` is much less likely to cause user errors instead of setting them to 0 and then knowing which values inside the `dpFixed_nominal` are for which component.
Consider for example a pipe, boiler and heat pump in series.
Either have `dpFixed_nominal=1000 + 2000 + 2400` or `dpFixed_nominal=pip.dp_nominal + boi.dp_nominal + heaPum.dp_nominal`. The latter better explains without further doc where these values come from in my opinion.

mwetter:
But `dpFixed_nominal=pip.dp_nominal + boi.dp_nominal + heaPum.dp_nominal` is in my view bad style. We typically avoid that components access other components' parameters because it defeats the modular modeling principle. Rather why not using
```modelica
parameter Modelica.Units.SI.PressureDrop dpPip_nominal = 1000;
parameter Modelica.Units.SI.PressureDrop dpBoi_nominal = 1000;
parameter Modelica.Units.SI.PressureDrop dpHeaPum_nominal = 1000;
...
val(
  dpFixed_nominal = dpPip_nominal + dpBoi_nominal + dpHeaPum_nominal,
...
)
```
This way, a user sees all parameters in the parameter dialog. I found it always painful if parameters are access from other models rather than assigned top-down. (I am debugging now a huge model, and understanding what parameter value is from where is a real challenge without a uniform structure.)

FWuellhorst:
I see your point, at the same time it could lead to errors if the component `boi` is removed, but the parameter is still there. Also, we sometimes declare dp_nominal based on this equation using constant values for `a` and `n` (the latter being close to 2 according to datasheets). 
```
  parameter Modelica.Units.SI.PressureDifference dp_nominal=a*(m_flow_nominal/rho_default)^n
      "Pressure drop at nominal mass flow rate"
    annotation (Dialog(group="Nominal condition"));
```
In this case, propagating all parameters is possible, but adds a lot of redundant code and parameters. For these two reasons I would like to have the option under Advanced also for the simple pressure drop. 
But, if you disagree, I will set it to final and we will just introduce an extra parameter in AixLib in those cases.

mwetter:
@FWuellhorst : As there is an option to add an extra parameter in AixLib I would rather set it to final as this is a simple model that is used very often.

FWuellhorst:
@mwetter : Ok. As `computeFlowResistance` is defined in this simple model, I had to define an extra `_internal` protected parameter in `PressureDrop` to disable the calculation only in `HydraulicDiameter`. 
I think the new parameter `disableComputeFlowResistance` is much clearer for users, and having the default to `false` instead of the equation is also better.

mwetter (approved):

PR #2015, opened 2025-05-26
Fix error counter in heat pump safety control

### LINKED ISSUE #2011: Wrong test if heat pump safety control for minimum flow rate
@FWuellhorst 
The model `IBPSA.Fluid.HeatPumps.ModularReversible.Controls.Safety.MinimalFlowRate`  increments the error counter `disErr` each time the mass flow rates `mConMea_flow` and `mEvaMea_flow` are zero (or below the minimum allowed flow rate). However, if a heat pump is off, these are zero, and `ySet` is also zero.

The error counter should take into account whether the heat pump is actually operating before incrementing the error counter.

Image

FWuellhorst:
Closes #2011 .

@mwetter: The issue is valid for the other blocks, as well. Thus, I added the fix to the partial safety controller with error counter class.

I checked the examples and the counts look fine to me. 
Only in the aggregated example, there sometimes is an increase in err for two components, e.g. `opeEnvErr` and `minFlowErr`, see picture below.
image
The event log shows two time events close to each other, see analysis:
image

I do think the new hys blocks could lead to more state events in a closed loop simulation, or do you think it makes no difference?
If it leads to more events, should I add a general `countErrors` Boolean to the safety blocks?

### DIFF
--- a/IBPSA/Fluid/HeatPumps/ModularReversible/Controls/Safety/AntiFreeze.mo
+++ b/IBPSA/Fluid/HeatPumps/ModularReversible/Controls/Safety/AntiFreeze.mo
@@ -44,16 +44,21 @@ equation
       points={{1,20},{20,20},{20,0},{38,0}},
       color={255,0,255}));
   annotation (Documentation(revisions="<html><ul>
-<li>
-  <i>October 2, 2022</i> by Fabian Wuellhorst:<br/>
-  Adapted based on IBPSA implementation (see issue <a href=
-  \"https://github.com/ibpsa/modelica-ibpsa/issues/1576\">#1576</a>)
-</li>
-<li>
-  <i>November 26, 2018</i> by Fabian Wuellhorst:<br/>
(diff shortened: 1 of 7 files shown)


mwetter:
@FWuellhorst : I don't think the two events close to each other will pose an issue. I suggest we keep the counter always present as it is now.
I also cleaned up the model to remove two instance of `ySetOn` which were equal and removed the size of the boundary check blocks as they were overlapping with a `not` block underneath.
From my point of view this is ready to be merged.

A future improvement could be to put `yOn` after the on off safety control on the bus, and then use this in the downstream blocks to avoid the multiple instances of `ySetOn`.

FWuellhorst:
Thanks for the adjustments. I further added revisions.

Regarding `yOn`, this will not work with the current serial control logic, where one output affects the others. While series connection is required for the onOff checks, the other three blocks could be put in parallel, if I am not mistaken. Still, one global `yOn` would not work.

PR #2032, opened 2025-07-10
Use max pressure for state if flow reversal allowed for airflow models

kldjonge:
Updated the pressure input to Medium.setState_phX to use the maximum of port_a.p and port_b.p when allowFlowReversal is true. This ensures more correct state calculation during flow reversal scenarios. In the door models, the flow is allready calculated based on actual rho (direction-dependent).

### DIFF
--- a/IBPSA/Airflow/Multizone/BaseClasses/PartialOneWayFlowElement.mo
+++ b/IBPSA/Airflow/Multizone/BaseClasses/PartialOneWayFlowElement.mo
@@ -57,7 +57,7 @@ equation
   else
     sta = if homotopyInitialization then
         Medium.setState_phX(
-          port_a.p,
+          if allowFlowReversal then max(port_a.p,port_b.p) else port_a.p,
           homotopy(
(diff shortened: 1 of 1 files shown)


mwetter:
Can you please test whether this change affects results more than around $10^{-3}$. If the change is that small, which I expect, any other uncertainty will dominate it, and it is similar to the solver tolerance. In that case I think a boolean constant just adds complexity for no reason.

So, in summary I suggest to first change if this option would indeed make a practical difference, and only then, add a boolean constant. Keeping models simple is important as we often get comments regarding complexity and learning curve.

kldjonge:
I'll check it! Thanks for the feedback.

10^-3 for units of SI units kg/s? Because the relevant unit for comparison would be m^3/h in case of airflow.

mwetter:
The $10^{-3}$ I mentioned is relative error, so unitless. (Solvers in Modelica tools compare relative and absolute errors.)

kldjonge:
I will close this for now and re-open if I found thh time to look more into the relevance of it.

PR #2042, opened 2025-09-03
Guarded against division by zero in borefield data

mwetter:
This closes #2014

### DIFF
--- a/IBPSA/Fluid/Geothermal/Borefields/Data/Filling/Template.mo
+++ b/IBPSA/Fluid/Geothermal/Borefields/Data/Filling/Template.mo
@@ -11,7 +11,8 @@ record Template
   parameter Boolean steadyState = (cFil < Modelica.Constants.eps or dFil < Modelica.Constants.eps)
     "Flag, if true, then material is computed using steady-state heat conduction"
     annotation(Evaluate=true);
-  final parameter Modelica.Units.SI.ThermalDiffusivity aFil=kFil/(dFil*cFil)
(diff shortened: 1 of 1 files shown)


IBPSA/Fluid/Geothermal/Borefields/Data/Filling/Template.mo
```diff
@@ -11,7 +11,8 @@ record Template
   parameter Boolean steadyState = (cFil < Modelica.Constants.eps or dFil < Modelica.Constants.eps)
     "Flag, if true, then material is computed using steady-state heat conduction"
     annotation(Evaluate=true);
-  final parameter Modelica.Units.SI.ThermalDiffusivity aFil=kFil/(dFil*cFil)
+  final parameter Modelica.Units.SI.ThermalDiffusivity aFil=
+    if (dFil*dFil > Modelica.Constants.eps) then kFil/(dFil*cFil) else 0
```
thorade, line 15: should that be if(dFil*cFil) instead?

MassimoCimmino: There is also a potential conflict with the previous statement on line 11. For example, if `cFil < sqrt(eps)` and `dFil < sqrt(eps)`, then `steadyState = false` and `aFil = 0`. This could cause simulations to fail.

I suggest the (corrected) condition `dFil * cFil > Modelica.Constants.eps` should also be applied to line 11.

While it should not affect results when `steadyState = false`, `aFil` should be overloaded by a large number (or `inf`) instead of 0 to preserve the expected behavior.

Issue #2043, opened 2025-09-11
Performance improvement for IDEAS.Airflow.Multizone.BaseClasses.TwoWayFlowElement


kldjonge:
I noticed some discrepencies in the simulation results between mBA_flow and VBA_flow*rho_a1_inflow at low flows when using the operable door model (IDEAS.Airflow.Multizone.DoorDiscretizedOperable). If VAB_flow was **exactly** zero, mAB_flow wasn't while it should be. So somehow mBA fluctuated around zero while it should have been zero and, consequently, the 'error control' on that flow limited the stepsize a lot.  

Upon further investigation, this was fixed by  explicitly linking Vflow*rho to the port:

  **port_a1.m_flow = VAB_flow*rho_a1_inflow;**
  **port_a2.m_flow = VBA_flow*rho_a2_inflow;**

For me, this decreased simulation time significantly. Since the door models all actually calculate volume flows at top level, it will most-likely improve the performance in general for all models using this class.

A larger snippet:

```  
  mAB_flow=VAB_flow*rho_a1_inflow;
  mBA_flow=VBA_flow*rho_a2_inflow;
  // Average velocity (using the whole orifice area)
  vAB = VAB_flow/A;
  vBA = VBA_flow/A;

  port_a1.m_flow = VAB_flow*rho_a1_inflow;
  port_a2.m_flow = VBA_flow*rho_a2_inflow;

  // Energy balance (no storage, no heat loss/gain)
  port_a1.h_outflow = inStream(port_b1.h_outflow);
  port_b1.h_outflow = inStream(port_a1.h_outflow);``
```

mwetter:
This model still has peaks in a trajectory that should be smooth, as reported in https://github.com/lbl-srg/modelica-buildings/issues/4360

Setting `unbounded=true` for the states `mExcAB` and `mExcBA` solves the problem. These states are indeed unbounded.

PR #2053, opened 2025-09-18
Issue2043 two way flow element

mwetter:
Changes to IBPSA.Airflow.Multizone.BaseClasses.TwoWayFlowElement

For #2043

To do:
- [x]  Add revision notes


IBPSA/Resources/Scripts/Dymola/Airflow/Multizone/Examples/OneOpenDoor.mos
```diff
@@ -1,4 +1,4 @@
-simulateModel("IBPSA.Airflow.Multizone.Examples.OneOpenDoor", stopTime=7200, method="CVode", tolerance=1e-06, resultFile="OneOpenDoor");
+simulateModel("IBPSA.Airflow.Multizone.Examples.OneOpenDoor", stopTime=7200, numberOfIntervals=120, method="CVode", tolerance=1e-09, resultFile="OneOpenDoor");
```
mwetter, line 1: I don't think any of these changes are needed. Also, I think this is the first model where we use such high a tolerance, which is surprising for such simple a model. Can you revert this?

IBPSA/Airflow/Multizone/BaseClasses/TwoWayFlowElement.mo
```diff
@@ -41,9 +41,9 @@ protected
   Modelica.Units.SI.VolumeFlowRate VZer_flow(fixed=false)
     "Minimum net volume flow rate to prevent zero flow";
 
-  Modelica.Units.SI.Mass mExcAB(start=0, fixed=true)
+  output Modelica.Units.SI.Mass mExcAB(start=0, fixed=true)
     "Air mass exchanged (for purpose of error control only)";
-  Modelica.Units.SI.Mass mExcBA(start=0, fixed=true)
+  output Modelica.Units.SI.Mass mExcBA(start=0, fixed=true)
```
mwetter, line 46: Why is an `output` qualifier needed for a state variable?

IBPSA/Airflow/Multizone/BaseClasses/TwoWayFlowElement.mo
```diff
@@ -88,8 +88,8 @@ equation
   vAB = VAB_flow/A;
   vBA = VBA_flow/A;
 
-  port_a1.m_flow = mAB_flow;
-  port_a2.m_flow = mBA_flow;
+  port_a1.m_flow = rho_a1_inflow*VAB_flow;
```
mwetter, line 91: What is the rational for this change? Note that the original model has
```modelica
  mAB_flow = rho_a1_inflow*VAB_flow;
  mBA_flow = rho_a2_inflow*VBA_flow;
...
  port_a1.m_flow = mAB_flow;
  port_a2.m_flow = mBA_flow;
```
and this is now
```modelica
  mAB_flow = rho_a1_inflow*VAB_flow;
  mBA_flow = rho_a2_inflow*VBA_flow;
...
  port_a1.m_flow = rho_a1_inflow*VAB_flow;
  port_a2.m_flow = rho_a2_inflow*VBA_flow;
```
which is exactly the same math, but one more operation to compute. This should not have any impact on the result.

IBPSA/Airflow/Multizone/BaseClasses/TwoWayFlowElement.mo
```diff
@@ -19,13 +19,13 @@ partial model TwoWayFlowElement "Flow resistance that uses the power law"
   parameter Modelica.Units.SI.Velocity vZer=0.001
     "Minimum velocity to prevent zero flow. Recommended: 0.001";
 
-  Modelica.Units.SI.VolumeFlowRate VAB_flow(nominal=0.001)
+  Modelica.Units.SI.VolumeFlowRate VAB_flow(nominal=1/3600)
```
mwetter, line 22: None of these changes to the nominal attribute has an effect, as tested on the original model. They are not bad per se, but not necessary or helpful either. Unless there is a reason, we should not just change setting for no reason as any change indicates to others that this was needed.

mwetter (commented):
@kldjonge : I looked at your code changes, but see no reason for any of the changes to the `.mo` code.
The only change that causes smoother trajectories in the reference results is the change in solver tolerance. However,
1. a change from 1E-6 to 1E-7 suffices, there is no need to use 1E-9
2. there is no need to specify the `numberOfIntervals` which we typically don't do
3. the non-smooth change in the trajectory near t=5615 is actually removed with Dymola 2026x beta2

I don't see a reason for implementing any of the changes. We could change the tolerance to 1E-7, but in view of the upcoming Dymola update, I suggest we leave it as-is.

IBPSA/Airflow/Multizone/BaseClasses/TwoWayFlowElement.mo
kldjonge: I noticed that in the results, the mAB_flow was near zero but not exactly at times that VAB_flow was. This caused the error control to unnecessarily limit the time step enormous amounts. I figured out that this probably had to do with how causality of the variables was interpreted during symbolic processing.  Since the doordiscretized model is based on the powerlaw equation, which calculated the volume flow the only actual reason to calculate the variable mAB_flow and mBA_flow is for the error control states.  The change I propose directly links the output of the powerlaw with the port and thus the rest of the model.

This improvement turns out to be more robust and solved the issue at hand.

If avoiding the additional operation is needed,  mAB_flow and mBA_flow should be deleted altogether since they are unnecessary.

IBPSA/Airflow/Multizone/BaseClasses/TwoWayFlowElement.mo
kldjonge: Togheter with the next change, the ouput qualifier helps the IDE interpret the causality of the equations. Here specicially it might not be needed but it is better to do it if the causality is known.

kldjonge:
> @kldjonge : I looked at your code changes, but see no reason for any of the changes to the `.mo` code. The only change that causes smoother trajectories in the reference results is the change in solver tolerance. However,
> 
> 1. a change from 1E-6 to 1E-7 suffices, there is no need to use 1E-9
> 2. there is no need to specify the `numberOfIntervals` which we typically don't do
> 3. the non-smooth change in the trajectory near t=5615 is actually removed with Dymola 2026x beta2
> 
> I don't see a reason for implementing any of the changes. We could change the tolerance to 1E-7, but in view of the upcoming Dymola update, I suggest we leave it as-is.

@mwetter: I commented on the review. If the smaller changes will either be solved in the new dymola or are of too little impact I suggest to only implement this change:

```
  port_a1.m_flow = rho_a1_inflow*VAB_flow;
  port_a2.m_flow = rho_a2_inflow*VBA_flow;
```

This solves the main issue.

IBPSA/Airflow/Multizone/BaseClasses/TwoWayFlowElement.mo
mwetter: I will revert them. They are in fact not used in this example model. (The `nominal` attribute is only used for states and for iteration variables of nonlinear equations, for which we have none here.)

mwetter:
@kldjonge : can you please do a review and then we can merge this to the `master`.

PR #2054, opened 2025-09-18
Refactored powerLawFixedM.mo for m=0.5

mwetter:
@kldjonge : This refactoring gives about 10% faster simulation time for OneOpenDoor.
Also, the profiling in Dymola shows that the now optimized function, together with `powerLaw`, dominate the computing time. I tried to optimize `powerLaw` too, but as `m` varies with time, the formulation I tried introduced a discontinuity and the CVode solver issued a lot of warning.

I think the optimization in the PR would be worthwhile. Results change slightly due to numerical noise, presumably because `x^0.5` and `sqrt(x)` are implemented differently, and some of these air flow models are very sensitive to noise.

I have not yet tested this change on the `Buildings` library.


IBPSA/Airflow/Multizone/BaseClasses/powerLawFixedM.mo
```diff
@@ -22,13 +22,14 @@ protected
   Real pi2 "Square of normalized pressure";
 algorithm
  if (dp >= dp_turbulent) then
-   V_flow :=C *dp^m;
+   V_flow :=C * (if (m < 0.500000000001 and m > 0.499999999999) then sqrt(dp) else dp^m);
  elseif (dp <= -dp_turbulent) then
```
kldjonge, line 26: I propose to use the isEqual() function:

V_flow :=C * (if Modelica.Math.isEqual(m,0.5,1e-12) then sqrt(dp) else dp^m);

kldjonge:
Great improvement the 10%! I suggest to use the Modelica.Math.isEqual() function for clarity and compactness. But no other suggestions.

kldjonge:
@mwetter nice work! 35% is really great! especially since these functions/models often end up in a non-linear system. The powerlaw function is **by far** the most-called function when I use the embedded airflow implementation in IDEAS.

mwetter:
Great, hope that has an impact on the IDEAS models as well.
I move ahead and merge this to the dev branch.

PR #2089, opened 2026-02-18
improved ill-posed integer rounding

### LINKED ISSUE #2088: Ill-posed integer rounding
This is for the [issue 4480](https://github.com/lbl-srg/modelica-buildings/issues/4480) raised in Modelica Buildings Library repo.

JayHuLBL:
This closes #2088.


mwetter:
@tsnouidui : I still don't understand why there are two different equations to assign `tNext`. 
If there is indeed a rational, it must be documented in the `.mo` file so users and developers understand.
Also, is the `start` attribute needed for `tNext` if it is assigned in an initial() clause?

PR #2109, opened 2026-04-13
Add nominal value for integrator state of PID controller in CDL package

mwetter:
This closes #2106

### DIFF
--- a/IBPSA/Controls/OBC/CDL/Reals/PID.mo
+++ b/IBPSA/Controls/OBC/CDL/Reals/PID.mo
@@ -481,7 +481,7 @@ American Society of Heating Refrigerating and Air-Conditioning Engineers Inc. At
 <ul>
 <li>
 April 13, 2026, by Michael Wetter:<br/>
-Added value for <coded>nominal</code> attribute for integrator state <code>I.y</code><br/>
+Added value for <code>nominal</code> attribute for integrator state <code>I.y</code><br/>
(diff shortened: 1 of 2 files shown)


IBPSA/Controls/OBC/CDL/Reals/PID.mo
```diff
@@ -60,6 +60,7 @@ block PID
     "Gain for proportional control action"
     annotation (Placement(transformation(extent={{-50,130},{-30,150}})));
   IBPSA.Controls.OBC.CDL.Reals.IntegratorWithReset I(
+    y(nominal=max(abs(yMax), abs(yMin))),
```
FWuellhorst, line 63: Why use abs? If both yMin and yMax are negative (if that is possible), shouldn't the nominal be negative, as well?

mwetter: As nominal values are used for scaling, I never saw a negative value. Hence I think it is better to use `abs` here rather than hope that tools compute the absolute value.
Another reason is that without `abs`, the assignment 
```
nominal = max(0.1, -10)
```
would return 0.1, but we want it to be 10.

FWuellhorst: Makes sense, thanks for the explanation!

FWuellhorst (approved):

PR #2116, opened 2026-04-28
Refactor Antifreeze package to enable function calls

### LINKED ISSUE #2115: Update Antifreeze package to allow function calls from record
The current implementation of the `Media.Antifreeze` package does not allow calling medium property functions from a Modelica record, because a `record` is not allow to instantiate a `package`.
Moreover, the functions are `protected` and hence cannot be called, such as here:
https://github.com/ibpsa/modelica-ibpsa/blob/a473c121155ba185407023ed705ff7a83925f6e8/IBPSA/Media/Antifreeze/EthyleneGlycolWater.mo#L145
Also, they are for some reason `replaceable` which is not needed.

This issue is to refactor the implementation to move the media functions into a package where they are actual functions, and then call these functions from the media `packages`.

mwetter:
This closes #2115


mwetter:
@amaccarini : Can you please do a review of this PR. Thanks.

amaccarini (approved):

PR #2118, opened 2026-04-29
Issue2111 evaluate table on file

### LINKED ISSUE #2111: Evaluate parameter tableOnFile in Controls.SetPoints.Table
Controls.SetPoints.Table uses MSL CombiTimeTable with modifier tableOnFile=false
https://github.com/ibpsa/modelica-ibpsa/blob/master/IBPSA/Controls/SetPoints/Table.mo#L19
It is still considered a parameter, and it is not evaluated.

There seem to be three ways to have it evaluated:
- Make it final,
- Use a parameter with annotation Evaluate=true,
- Use a constant Boolean in protected section 
 
All three result in slightly simpler flat Modelica code, potentially speeding up compilation by a few seconds, but I did not benchmark it. To me it looks like the intention is to never use a file, and I suggest to mark it as final. There are a few more similar places elsewhere, if this change is useful I can edit those as well.

The alternative (and maybe cleaner) solution would be to add annotation Evaluate in MSL:
https://github.com/modelica/ModelicaStandardLibrary/pull/4776

mwetter:
This closes #2111

@thorade : Thanks for pointing out this improvement. Can you please review this PR.
I changed the configuration in the models, and also in some of the examples, focusing the latter on those that are computationally demanding. For the smaller examples, I didn't make the change.


thorade:
I think this looks good. The interesting part is mostly the base classes, these three:

-  Set Point table
- PartialWindTurbineBase
- WeeklySchedule

Where paramter tableOnFile is propagated, adding annotation Evaluate is the best solution.

thorade (approved):

PR #2145, opened 2026-06-23
correct text label annotation in connect statements

### LINKED ISSUE #2137: Wrong argument in connection label annotations
Per the Modelica Specification, the argument for the connection `Text` label is `string`, not `textString`:
https://specification.modelica.org/maint/3.7/annotations.html#connections1

But there are almost 100 cases of getting this wrong in IBPSA (and downstream libraries like Buildings). Here is the first example:

https://github.com/ibpsa/modelica-ibpsa/blob/18cb5d00dd8ff168d7aa56c19cb0b9b30dd513ed/IBPSA/BoundaryConditions/SkyTemperature/Examples/BlackBody.mo#L21-L27

mwetter:
This closes #2137

### DIFF
--- a/bin/verifyFiles.py
+++ b/bin/verifyFiles.py
@@ -61,6 +61,7 @@ INVALID_REGEXP_IN_MO=[r"StopTime\s*=\s*\d\s*[*]\s*\d+",
                       r"file\s*=\s*\"Resources", # This should be file="modelica://IBPSA/Resources
                       r"parameter.*Boolean.*homotopyInitialization",
                       r"(Text\s*\([^\)]*)lineColor",
+                      r'(Text\(\s*)textString\s*=\s*"[^"]*"\s*,\s*',
                       r"(Line\s*\([^\)]*)lineThickness"]
 # List of strings that are required in .mo files, except in Examples
 REQUIRED_IN_MO=["documentation"]
--- a/IBPSA/BoundaryConditions/SkyTemperature/Examples/BlackBody.mo
+++ b/IBPSA/BoundaryConditions/SkyTemperature/Examples/BlackBody.mo
@@ -22,7 +22,7 @@ equation
       points={{10,10},{-20,10}},
       color={255,204,51},
       thickness=0.5), Text(
-      textString="%first",
+      string="%first",
       index=-1,
       extent={{-6,3},{-6,3}}));
   connect(weaBus.TDewPoi, TBlaSky.TDewPoi) annotation (Line(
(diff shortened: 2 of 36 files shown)


mwetter:
@jelgerjansen : Can you please review this PR. We also had many of them in the Buildings library, so you may want to check IDEAS for these wrong annotations as well.

jelgerjansen:
@mwetter please check with @LoneMeertens or @annadellisola if they can do a KU Leuven review (as I left the university 2 months ago)

bin/verifyFiles.py
```diff
@@ -61,6 +61,7 @@
                       r"file\s*=\s*\"Resources", # This should be file="modelica://IBPSA/Resources
                       r"parameter.*Boolean.*homotopyInitialization",
                       r"(Text\s*\([^\)]*)lineColor",
+                      r'(Text\(\s*)textString\s*=\s*"[^"]*"\s*,\s*',
```
annadellisola, line 64: In the current implementation, `textString` is flagged as an error when it appears as the first argument in `Text(...)`. However, this might be problematic for graphical annotations, where `Text` is used within `Icon(graphics={...})` and `textString` is the correct attribute according to the Modelica specification.

How does this check distinguish between graphical `Text` annotations (where `textString` is valid) and other contexts, to avoid incorrect error reporting?

annadellisola (commented):
@mwetter Thank you for addressing the issue. We will check possible wrong annotations also in IDEAS. 
I only had a comment on the python verification code.

bin/verifyFiles.py
mwetter: @annadellisola : I looked into reformulating the regular expression but working around this situation where `textString` is valid in `Icon(graphics={...})` but not in `Text(...)` is rather complex. To do it robust, one should probably use a tokenizer and then do a syntax check.

The proposed check does however capture the wrong annotations that some previous version of Dymola introduced, in which case `textString` is the *first* argument. In contrast, if one draws a text in an `Icon` using Dymola, it is *not* the first argument. So the current version, although not perfect, captures the annotations that Dymola wrongly introduced (and which we may still have on some active development branches) and it does not trigger the correct use of `textString` in an `Icon`.

Hence, my suggestion is to move forward with the current test, and then correct it if it indeed causes problems later on.

annadellisola: @mwetter Thank you for looking into this. I agree, moving forward with the current test seems like the most practical approach, and we can revisit it if it causes issues in the future.

annadellisola (approved):
@mwetter Thank you for looking into this and for the clarification. I agree with the proposed approach and approve the PR.

PR #2147, opened 2026-06-25
Added nominal attribute to state variables

### LINKED ISSUE #2132: der operator creates dynamic states, states should have nominals
A bit similar to #2106 the temperature two port sensor 
Fluid.Sensors.TemperatureTwoPort
has a state if dynamics are enabled, because of usage of der-operator in line 68/70:

https://github.com/ibpsa/modelica-ibpsa/blob/18cb5d00dd8ff168d7aa56c19cb0b9b30dd513ed/IBPSA/Fluid/Sensors/TemperatureTwoPort.mo#L65-L74

but the state does not have a nominal value set:

Image

Would it be reasonable to use the start value also as nominal value?
The same probably applies to other sensors that have dynamics based on der() operator

mwetter:
This closes #2132

This sets nominal attributes in a variety of models.


mwetter:
@thorade : This PR adds the nominal attributes to the state variables. Could you please do a review. (Perhaps some more reference results need to be updated, which I could do the week after next.)

thorade (approved):

thorade:
These all look reasonable to me! 
I would only expect to see an effect on performance or results for the variables that are several orders of magnitude different from 1.0, for example enthalpy sensor

PR #2157, opened 2026-07-30
Corrected argument in splineDerivative

### LINKED ISSUE #2130: Fluid Dampers PressureIndependent question
https://github.com/ibpsa/modelica-ibpsa/blob/c52a650877e5246cc8b19b1e7e7ba609df73aa68/IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo#L135-L143

On line 143, should this use kSupSpl_raw as input, or the sorted kSupSpl?
splineDerivatives function expects strictly increasing x values

mwetter:
This closes #2130 

This corrects the argument. Models that do not have this fix would stop the simulation due to an assertion that actually checks for the right ordering of the element vectors.

### DIFF
--- a/IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo
+++ b/IBPSA/Fluid/Actuators/Dampers/PressureIndependent.mo
@@ -140,7 +140,7 @@ initial equation
   for i in 1:sizeSupSpl loop
     ySupSpl[i] = sum((if k == idx_sorted[i] then ySupSpl_raw[k] else 0) for k in 1:sizeSupSpl);
   end for;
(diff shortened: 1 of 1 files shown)


mwetter:
@AntoineGautier : Can you please do a review of this PR.

AntoineGautier (approved):
The changes are correct and match the original formulation before https://github.com/ibpsa/modelica-ibpsa/pull/1514/changes

On a side note, I've tested the natural formulation `ySupSpl = ySupSpl_raw[idx_sorted]` instead of the trick using a sum with masking. Both models `IBPSA.Fluid.Actuators.Dampers.Examples.Damper` and `IBPSA.Fluid.Actuators.Dampers.Validation.PressureIndependent` can be simulated without warning with OCT 1.66.
So, I would suggest to adopt `ySupSpl = ySupSpl_raw[idx_sorted]` which is more readable.

PR #2178, opened 2026-09-15
Fix distance computation in KMeans inertia evaluation

### LINKED ISSUE #2175: centroids in KMeans
This might be related to or a duplicate of #1985

The output centroids is instantiated in line 13: 

https://github.com/ibpsa/modelica-ibpsa/blob/93d17fe2deb655d650c0fb10f13d92cf230c89c2/IBPSA/Utilities/Clustering/KMeans.mo#L13

and used for the first time in line 99
https://github.com/ibpsa/modelica-ibpsa/blob/93d17fe2deb655d650c0fb10f13d92cf230c89c2/IBPSA/Utilities/Clustering/KMeans.mo#L96-L101

but it is used on the right-hand side, so at least during the first run it is used uninitialized.
It is only assigned a value later on, in line 105
https://github.com/ibpsa/modelica-ibpsa/blob/93d17fe2deb655d650c0fb10f13d92cf230c89c2/IBPSA/Utilities/Clustering/KMeans.mo#L103-L108 

I suggest to change line 99 to use new_centroids instead of centroids

mwetter:
Corrected the computation of distance to use new_centroids instead of centroids, which was not initialized.

This closes #2175


mwetter:
@MassimoCimmino : Do you agree with this bug fix?

mwetter:
@MassimoCimmino : Note that `ClusterBoreholes_100boreholes` compares the expected clustering with the actual clustering, and this comparison now fails. Can you please check if the new result is valid.

